PR #458, opened 2026-06-08
fix(analytics): resolve GA client id from async gtag callback

VincentShipsIt:
## What

`getGAClientId()` read `gtag('get', ...)` synchronously, which always returned `undefined` — GA resolves the value through an async callback.

## Changes

- `getGAClientId()` now returns `Promise<string | undefined>`, resolving from the gtag callback
- Use `EnvironmentService.GA_ID` for the measurement id instead of a hardcoded/missing value
- Moved `GoogleTag` / `WindowWithGoogleTag` inline types into `@genfeedai/interfaces/analytics/link-tracking.interface`
- Updated test to await the async client id and assert the gtag `get` call shape

## Why

Resolves CodeRabbit review finding on #443. Reading the client id synchronously was a real bug — callers never got a value.

## Verification

- `bun run test --filter=@genfeedai/services` — 7 pass
- `bun type-check` — green
- `bunx turbo lint` (services + interfaces) — green
- `npx biome check` — clean

### DIFF
--- a/packages/interfaces/src/analytics/link-tracking.interface.ts
+++ b/packages/interfaces/src/analytics/link-tracking.interface.ts
@@ -107,6 +107,24 @@ export interface IGoogleAnalyticsEvent {
   };
 }
 
+export type GoogleTag = {
+  (
+    command: 'event',
+    eventName: string,
+    eventParams: IGoogleAnalyticsEvent['eventParams'],
+  ): void;
+  (
+    command: 'get',
+    measurementId: string,
+    fieldName: 'client_id',
+    callback: (id: string) => void,
+  ): void;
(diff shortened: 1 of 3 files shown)


VincentShipsIt (commented):
Blocking review findings:

- [P1] getGAClientId can now return a Promise that never settles if GA is blocked, misconfigured, or does not invoke the gtag get callback. The method is public and future awaited call sites would hang indefinitely. Guard an empty EnvironmentService.GA_ID and add a short timeout that resolves undefined while clearing the timeout on callback.

CI note: Format is red, but the failing files are existing Unipile/integrations files outside this PR diff.

PR #487, opened 2026-06-09
fix(devops): remove ghost fanvue/llm services, add postgres to selfhosted, add staging migrations

VincentShipsIt:
## Summary

- **Remove ghost fanvue service** from production + staging compose files, volumes, WAVE arrays in both deploy scripts, `fanvue_changed` output from `detect-changes.yml`, and all consumers in `_deploy.yml`. `apps/server/fanvue` does not exist and never had a Dockerfile target.
- **Remove orphan llm path filter** (`llm_changed`) from `detect-changes.yml` — `apps/server/llm` does not exist.
- **Delete `docker/docker-compose.fanvue.yml`** — was only referenced by the now-removed ghost service.
- **Delete `docker/crontab`** — references 6 non-existent scripts; no Dockerfile installs it; zero references elsewhere in the repo.
- **Add `postgres:16-alpine` to `docker-compose.selfhosted.yml`** — with `POSTGRES_DB/USER/PASSWORD` matching the default `DATABASE_URL` in `selfhosted-entrypoint.sh`, a `postgres_data` named volume, a `pg_isready` healthcheck, and `genfeed` service `depends_on: postgres (service_healthy)`.
- **`selfhosted-entrypoint.sh`**: add `bunx prisma migrate deploy` with a 5-attempt retry loop before starting services; add `mcp` (port 3014) to match what the compose file advertises.
- **`deploy-staging.sh`**: add `run_db_migrations()` (verbatim from production) and call it after image pull, before wave deployment (matching production flow).

## Skipped / Noted

- **`discord` `build:prod`** (task 7): `apps/server/discord/package.json` only has a `build` script (no `build:prod`). Dockerfile.server line 160 left unchanged. Fixing this requires a separate discord package change to add the `build:prod` script.

## Test plan

- [ ] `bash -n` passes on all three edited `.sh` files (verified locally)
- [ ] All edited YAML files parse cleanly via `python3 -c "import yaml; yaml.safe_load(...)"` (verified locally)
- [ ] Verify staging deploy succeeds end-to-end with a test push to `develop`
- [ ] Verify self-hosted `docker compose up` brings postgres healthy before genfeed starts


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**PR #487 — remove ghost fanvue/llm services, add postgres to selfhosted, add staging migrations**

Reviewed across five maintainability dimensions — script duplication, atomicity/divergence, removal completeness, yaml/shell simplification, structural correctness — with each finding adversarially verified. The cleanup intent is sound and the service removal is largely thorough, but the diff leaves confirmed structural and completeness gaps that should land before merge.

### Blockers

**1. The migration step — the highest-risk action in the pipeline — is now duplicated verbatim across two deploy scripts with no shared source. (major)**
`docker/deploy-staging.sh` (~L220-251) adds `run_db_migrations()` byte-for-byte identical to the function already living in `docker/deploy-production.sh`. There is no `docker/deploy-common.sh`; the two scripts already share large near-identical blocks (`WAVE_1..WAVE_4`, `service_env_override_file`, the logging helpers), and this PR deepens that copy-paste with the one function you least want to drift. Any future change to migration behavior — working directory, network, timeout, retry — must be hand-applied to both files in lockstep, with nothing enforcing it.
→ Extract the shared body into `docker/deploy-common.sh` and `source` it from both scripts. Each top-level script then only sets its env-specific vars (`COMPOSE_FILE`, `ENV_FILE`, `DEPLOY_ENV`) and runs its sequencing. At an absolute minimum, `run_db_migrations` must have one authoritative definition.

**2. The fanvue removal is incomplete — dead references survive in the canonical templates. (completeness)**
The PR's stated objective is to remove the ghost fanvue service, but it does not finish the job:
- `docker/.env.example:29` still defines `FANVUE_PORT=3017` (that file is not in the diff at all).
- `.github/workflows/_deploy.yml:179` — the wave-order comment still reads "Wave 4: fanvue, clips, workers" while the actual `WAVE_4` array was changed to `(clips workers)` in this same PR.
Individually these are small, but a removal PR that leaves the variable and the documented wave order behind is precisely the "half-applied change" this bar rejects.
→ Strip both so the removal is atomic.

### Should-fix (non-blocking)

**3. `docker/selfhosted-entrypoint.sh:62` — `MCP_PORT` breaks the sibling pattern. (minor)**
Every other service derives its port from a `GENFEEDAI_*_URL` default declared in the top-of-file block (`PORT=$(echo "$URL" | sed ...)`), but MCP uses a standalone `MCP_PORT=${MCP_PORT:-3014}` with no `GENFEEDAI_MCP_URL`. An operator who reroutes via the established URL knob gets no effect for MCP alone.
→ Add `GENFEEDAI_MCP_URL=${GENFEEDAI_MCP_URL:-http://localhost:3014}` to the defaults block and derive `MCP_PORT` from it like the others.

**4. `docker/selfhosted-entrypoint.sh:38-55` — the migration retry loop has no DB-readiness gate of its own. (minor)**
The 5×5s (25s ceiling) loop is safe under `docker-compose.selfhosted.yml` because `depends_on: postgres: condition: service_healthy` gates it. But this entrypoint is also the image's entry for a plain `docker run` against an external postgres, where cold-start cluster init regularly exceeds 25s and exhausts the budget before migrations can run.
→ Prefix the loop with a `pg_isready`-based wait (e.g. 60s ceiling, 2s poll) against the host/port from `DATABASE_URL`, so readiness does not silently depend on Compose semantics.

**5. `docker/deploy-staging.sh` migration omits the retry the selfhosted path has. (nit)**
This is acceptable on its own — the no-retry shape is the pre-existing deploy-script norm (RDS is persistent and already up), not a regression. Add a one-line comment saying so, to stop anyone cargo-culting the selfhosted retry into a context that does not need it.

### Verdict

**BLOCK.** The direction is right and the new postgres/migration wiring is structurally reasonable, but two things must land first: (1) the migration step is the worst possible thing to leave duplicated across two scripts with no shared source — extract it into a sourced common file; and (2) the fanvue removal is not actually complete — finish stripping the dangling `FANVUE_PORT` and the stale wave comment so the change is atomic. Resolve those two and this is an easy approve; the three "should-fix" items can follow.

Issue #474, opened 2026-06-09
Security: add signature verification secrets for HeyGen/KlingAI/OpusPro webhooks


VincentShipsIt:
## Problem

Three production webhook endpoints accept arbitrary POST requests with zero authentication or signature verification:

- `apps/server/api/src/endpoints/webhooks/heygen/webhooks.heygen.controller.ts` — `@Public()`, no HMAC
- `apps/server/api/src/endpoints/webhooks/klingai/webhooks.kling.controller.ts` — `@Public()`, no HMAC
- `apps/server/api/src/endpoints/webhooks/opuspro/webhooks.opuspro.controller.ts` — `@Public()`, no HMAC

Any unauthenticated caller who knows or can guess a document id can POST a crafted payload to:
- Mark a generation as completed or failed, triggering S3 uploads and WebSocket notifications
- Overwrite `videoUrl` or metadata on arbitrary user records
- Force-fail any generation job

Stripe, Replicate, and Vercel webhooks in the same codebase all validate cryptographic signatures. These three do not.

Audit evidence: production-readiness audit, area "Main API Service", verified bugs at severity `high`.

## Operator task — do not route to a coding agent

**No agent or code work remains in this issue.** It is vendor coordination plus deploy configuration: obtain each signing secret from the vendor dashboard and set it in every deploy environment. Code-side HMAC verification ships separately in `feat/fix-webhook-security`, following the Replicate pattern.

## Operator steps

1. Obtain `HEYGEN_WEBHOOK_SECRET` from the HeyGen dashboard and record it in the infra secrets store.
2. Obtain `KLINGAI_WEBHOOK_SECRET` from the KlingAI dashboard and record it in the infra secrets store.
3. Obtain `OPUSPRO_WEBHOOK_SECRET` from the OpusPro dashboard and record it in the infra secrets store.
4. Set all three secrets in the production and staging deploy environments.
5. Confirm `feat/fix-webhook-security` is merged and deployed before or alongside closing this issue.

## Acceptance Criteria

1. THE SYSTEM SHALL hold `HEYGEN_WEBHOOK_SECRET`, `KLINGAI_WEBHOOK_SECRET`, and `OPUSPRO_WEBHOOK_SECRET` in the infra secrets store, each value obtained from its vendor dashboard.
2. THE SYSTEM SHALL define all three secrets in both the production and staging deploy environments.
[...]

PR #488, opened 2026-06-09
fix(security): close unauthenticated webhook + service surfaces

VincentShipsIt:
## Problem (P1 — audit-confirmed, adversarially verified)

- **HeyGen / KlingAI / OpusPro** webhooks: `@Public()`, zero caller verification. Forged callbacks trigger S3 uploads, DB writes, WebSocket events.
- **Replicate**: signature validated only when `NODE_ENV=production` — staging/preview accepted forgeries even with the secret configured.
- **Clerk**: Svix HMAC computed over `JSON.stringify(parsed body)` — re-serialization breaks the byte contract.
- **MCP**: `DELETE /mcp` was the only route without `mcpAuthMiddleware` — anyone could destroy sessions by id.
- **Clips**: `ClipperController` had no auth + no validation while proxying into the main API with a privileged key; routes were also double-prefixed (`/v1/v1/clips`).

## Fix

- New `webhook-token.util` (`assertWebhookToken` constant-time compare + `appendWebhookToken`). Secret rides on the callback URL we register with each vendor — works without vendor HMAC support. Env keys `HEYGEN_WEBHOOK_SECRET` / `KLINGAI_WEBHOOK_SECRET` / `OPUSPRO_WEBHOOK_SECRET` (optional; warn+accept when unset → zero-downtime rollout; provisioning tracked in #474).
- Replicate gate: `if (secret && isProduction)` → `if (secret)`.
- Clerk verifies over raw body (`express.raw` registered for `/v1/webhooks/clerk`, same as Stripe).
- MCP DELETE gets `mcpAuthMiddleware`.
- Clips: `InternalApiKeyGuard` (bearer = `GENFEEDAI_API_KEY`, timing-safe), class-validator DTOs, global `ValidationPipe`, prefix fix.
- Bonus: `heygen.service` stops reading `process.env` directly.

## Tests

234 webhook specs green, incl. new: token util (9), HeyGen 401/token/warn paths, Clerk raw-body verification. Type-check + lint green across api/mcp/clips/config.

Remediation P1-A. Closes the attack surface for #474's vendor-secret rollout.

### DIFF
--- a/apps/server/api/src/config/config.service.ts
+++ b/apps/server/api/src/config/config.service.ts
@@ -23,6 +23,7 @@ import {
   microservicesSchema,
   // Infrastructure
   newsApiSchema,
+  opusProSchema,
   postgresSchema,
   redisSchema,
   replicateSchema,
@@ -101,6 +102,7 @@ const apiSchema = Joi.object({
   ...elevenlabsSchema,
   ...leonardoSchema,
   ...heygenSchema,
+  ...opusProSchema,
   ...hedraSchema,
   ...newsApiSchema,
   ...darkroomSchema,
--- a/apps/server/api/src/endpoints/webhooks/clerk/webhooks.clerk.controller.spec.ts
+++ b/apps/server/api/src/endpoints/webhooks/clerk/webhooks.clerk.controller.spec.ts
@@ -67,10 +67,14 @@ describe('ClerkWebhookController', () => {
   });
 
   describe('handleClerk', () => {
-    const mockRequest = (headers: Record<string, string>) =>
+    const mockRequest = (headers: Record<string, string>, body?: unknown) =>
       ({
+        body:
+          body === undefined
+            ? undefined
+            : Buffer.from(JSON.stringify(body), 'utf8'),
         headers,
-      }) as Request;
+      }) as unknown as Request;
 
     const validHeaders = {
       'svix-id': 'test-id',
@@ -84,7 +88,7 @@ describe('ClerkWebhookController', () => {
         object: 'user',
         type: 'user.created',
       };
-      const request = mockRequest(validHeaders);
+      const request = mockRequest(validHeaders, payload);
       const mockEvent = { data: { id: 'user_123' }, type: 'user.created' };
 
       const svix = await import('svix');
@@ -96,15 +100,17 @@ describe('ClerkWebhookController', () => {
 
       clerkWebhookService.handleWebhookEvent.mockResolvedValue(undefined);
 
-      const result = await controller.handleClerk(request, payload);
+      const result = await controller.handleClerk(request);
 
-      expect(loggerService.log).toHaveBeenCalledWith(
-        'ClerkWebhookController createClerk received',
-        payload,
-      );
+      // Svix HMAC must be computed over the exact raw bytes received,
+      // never a re-serialized copy of the parsed body
       expect(mockVerify).toHaveBeenCalledWith(
         JSON.stringify(payload),
         validHeaders,
+      );
+      expect(loggerService.log).toHaveBeenCalledWith(
+        'ClerkWebhookController createClerk received',
+        { id: 'test-id', type: 'user.created' },
       );
       expect(clerkWebhookService.handleWebhookEvent).toHaveBeenCalledWith(
         mockEvent,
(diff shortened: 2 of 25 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**Scope:** `fix(security): close unauthenticated webhook + service surfaces` — 25 files, +705 lines, reviewed against `develop`. Multi-agent finder sweep across 7 dimensions, adversarially verified.

**Verdict: APPROVE.** The security fix is sound and should ship. Two structural items — a layer inversion and a config boundary leak — should be corrected in a fast follow; the rest are casts and dead branches the new code should not carry forward.

### 🟠 Structural — correct in a fast follow

**1. `webhook-token.util.ts` lives in `endpoints/` but is imported by `services/` — layer inversion**
`apps/server/api/src/endpoints/webhooks/webhook-token.util.ts`
`appendWebhookToken` is imported by three integration services (`klingai.service.ts:2`, `heygen.service.ts:4`, `opuspro.service.ts:2`). The services layer must not depend on the endpoints layer. The same change imports `encodeJwtToken` from `@api/helpers/utils/jwt/` — the correct location — so the pattern was already in hand. **Fix:** move the util (and its spec) to `apps/server/api/src/helpers/utils/webhook/`. Pure rename, no logic change.

**2. API-only webhook secrets added to shared `packages/config`**
`packages/config/src/schemas/ai.schema.ts:84-91` (and `schemas/index.ts`)
`OPUSPRO_WEBHOOK_SECRET` (new `opusProSchema`) plus `KLINGAI_/HEYGEN_WEBHOOK_SECRET` are consumed only by `apps/server/api`, yet now sit in the shared config surface that `workers`, `mcp`, `clips`, etc. all import. The api config already has `apiSpecificSchema` for exactly this boundary (`SOLANA_KEY`, `MARKETPLACE_API_URL` live there). **Fix:** move the three webhook secrets into `apiSpecificSchema` in `apps/server/api/src/config/config.service.ts` and out of the shared schemas.

### 🟡 Simplifications

**3. Identity casts `as string | undefined` on `configService.get()` — six sites**
`webhooks.heygen.controller.ts:33-35`, `…kling…:36-38`, `…opuspro…:36-38`, `heygen.service.ts:30`, `klingai.service.ts:49`, `opuspro.service.ts:26`
`BaseConfigService.get<K>` already returns `T[K]` = `string | undefined` for these optional keys. The cast suppresses nothing and misleads the reader into thinking `get()` returns `unknown`. The sibling `webhooks.stripe.controller.ts:49` casts nothing. **Fix:** delete all six.

**4. Clerk raw-body handling: dead fallback that silently breaks HMAC + a no-op cast chain**
`webhooks.clerk.controller.ts:63-66`
`Buffer.isBuffer(rawBody) ? … : JSON.stringify(rawBody as ClerkWebhookPayload)` — this PR registers `express.raw()` for the route, so the Buffer path is always true and the else branch is dead. If it ever ran, re-serialized JSON yields a wrong HMAC that *verifies as a 400* rather than failing loudly (the comment directly above even acknowledges this). The `as unknown` / `as ClerkWebhookPayload` chain is a no-op on Express's `any` body, and the sibling Stripe controller reads `request.body` with no fallback. **Fix:** assert `Buffer.isBuffer(request.body)` and throw if absent (matching the missing-header handling), then `const signedPayload: string = request.body.toString('utf8')`.

**5. Clerk bad-signature requests log twice**
`webhooks.clerk.controller.ts:71-93`
The inner catch around `webhook.verify` logs `"invalid signature"` then throws `BadRequestException`, which the outer catch re-logs as `"failed"`. Siblings (heygen/kling/opuspro) place their auth check *outside* the outer try so a rejection never enters it. **Fix:** move the svix verification ahead of the outer try.

**6. `InternalApiKeyGuard` duplicates `AdminApiKeyGuard`'s Bearer + `timingSafeEqual` logic**
`apps/server/clips/src/guards/internal-api-key.guard.ts:39-57`
Same extract-Bearer → length-check → `timingSafeEqual` structure; only the dev-mode bypass differs. **Fix:** extract a `compareBearerToken(provided, expected)` into a shared auth util and have both guards call it (each keeps its own error/bypass policy). Not required to merge, but do not let a third copy appear.

**7. `webhook-token.util.ts`: dead header-fallback branch + unsafe `query.token` cast**
`webhook-token.util.ts:31-33`
The `?? request.headers['x-webhook-token']` fallback has no production call site — every `appendWebhookToken` emits `?token=` URLs — yet the JSDoc advertises it as supported API. The `request.query?.token as string | undefined` cast also drops Express's `string[] | ParsedQs` cases silently. **Fix:** read `request.query.token` only, guard with `typeof rawToken === 'string'` instead of casting, and drop the header mention from the JSDoc.

**8. Duplicate OpusPro env section with inconsistent naming**
`packages/config/src/interfaces/env-config.interface.ts:110-112`
A new `// === OpusPro ===` block (`OPUSPRO_WEBHOOK_SECRET`) is inserted while a pre-existing `// === Opus Pro ===` block (`OPUS_PRO_KEY`) already exists — two sections, `OPUSPRO_*` vs `OPUS_PRO_*`, for one provider. **Fix:** consolidate into the existing block and normalize to `OPUS_PRO_WEBHOOK_SECRET`.

**9. Duplicate object key in a test mock**
`webhooks.kling.controller.spec.ts:46-47` — `warn: vi.fn()` is declared twice; JavaScript keeps the second and drops the first. Delete line 46.

---
*Automated thermo-nuclear review: 7-dimension finder sweep + adversarial verification (only confirmed findings shown). The security change itself is correct; items 1–2 are boundary corrections to land soon, 3–9 are cleanups. No tests were run.*

Issue #463, opened 2026-06-08
Finish Prisma schema-drift migration: write data-preserving migration.sql + gated prod deploy


VincentShipsIt:
## Context

Prisma schema-drift reconciliation in `apps/server/api` drove the api type-check from 38 errors → **0 (green)**. That work is **committed and pushed** on `hotfix/prisma-prod-deploy` (commit `e7c0cbcea`). The **data-preserving migration was never written**, so `schema.prisma` and the migrations directory are now out of sync — and a separate deploy commit (`b4071b644`) wired `prisma migrate deploy` into the deploy path. This issue tracks finishing the migration safely.

Session log: `.agents/SESSIONS/2026-06-08.md`.

## Done ✅

- api type-check is **green (0 errors)**, re-verified on the committed bundled state.
- Root cause fixed: **enum drift** (code enums lowercase `'pending'` vs Prisma enums UPPERCASE `PENDING`; `@map` can't bridge it because the generated TS enum VALUE = Prisma member NAME). Fix = collapse drifted columns to `String`, matching existing `status String?` precedent.
- 8 enum columns collapsed to `String`: `Post.platform`, `Post.status`, `Schedule.platform`, `Workflow.status`, `Task.status`, `ContentDraft.status`, `OutreachCampaign.status`, `CampaignTarget.status`.
- Added missing `ContentSchedule` cron columns (`cronExpression`, `timezone`, `isEnabled`, `nextRunAt`, `lastRunAt`, `skillSlugs`, `skillParams`) + indexes — cron path was broken at runtime.
- Added missing `ContentDraft` content columns (`content`, `mediaUrls`, `metadata`, `platforms`, …) + index — create path was throwing at runtime.
- Targeted code fixes: `content-performance` lowercase platform normalization, `templates` DTO-only field exclusion, `schedules` required `userId` + `Prisma.InputJsonValue` casts, `content-draft` interface cleanup, `agent-strategy-autopilot` JSON-spread cast.
- `CredentialPlatform` enum intentionally **kept** (still used by `Credential.platform` + `PostAnalytics.platform` — not drifting).

### 1. Write the data-preserving `migration.sql` — HIGHEST PRIORITY
`schema.prisma` has ~386 lines of changes but the last migration is still `20260603083913_reconcile_filter_fields`. **No migration exists for the drift.** Since `b4071b644` runs `prisma migrate deploy` on deploy, the prod DB will not get the drift changes and will mismatch the generated client → runtime errors.

[...]

PR #489, opened 2026-06-09
promote: develop → staging (Sentry root-cause fixes + docs)

VincentShipsIt:
## Summary

Promotes 8 commits resolving **all 13 unresolved production Sentry issues** (api-genfeed-ai ×7, files-genfeed-ai ×1, genfeed-ai website ×5) plus docs/memory updates.

- `2b5eac194` **fix(prisma): reconcile newsletters table with the application domain** — data-preserving migration `20260610010000_reconcile_newsletters` (title→label, scheduledAt→scheduledFor, +8 columns, status enum→lowercase TEXT with legacy remap); service writes scalar FK keys; new regression spec. Fixes both `PrismaClientValidationError` issues.
- `72f38b628` **fix(api): stop reporting expected 4xx traffic to Sentry** — global filter chain matched in reverse registration order, so the catch-all captured everything; 404 bot probes and Prisma P2025 paged Sentry (P2025 even returned 500). Reordered filters, ≥500 capture gate, skip expected P2002/P2003/P2025, deleted dead `DatabaseValidationExceptionFilter`, Clerk webhook signature failures now answer 400.
- `96f3e8086` **fix(config): make Google Ads env vars optional** — recurring boot-time Joi validation error in cloud prod.
- `e790e1a93` **fix(docker): give the files service writable runtime dirs and HOME** — fixes `EACCES: permission denied, mkdir`.
- `8ce04b4f3` **fix(web): resolve the website's production Sentry errors** — RSC fn-prop serialization on 5 landing pages, `suppressHydrationWarning` for next-themes, extension-noise `ignoreErrors`, interceptor throws real `Error`.
- `3fffaba1c` fix: dev:backend turbo concurrency · `679cd69c3` + `b87683c45` docs/memory.

## Verification

- develop CI green on `679cd69c3` (UI Guards re-run after runner infra flake — all steps null, log blob missing; passes locally and on rerun).
- `bun type-check` 42/42 · `turbo lint` 39/39 · scoped vitest green (api 283, config 133, services 891, ui 16) · `prisma validate`/`generate` · api build compiles.

## Deploy follow-ups (after master)

1. `prisma migrate deploy` against prod `genfeed-data` (includes `20260610010000_reconcile_newsletters`; see #463).
2. Rebuild server image (Dockerfile change) and one-time `chown -R 1001:1001` on the existing `files-logs` volume (or drop it) — pre-existing volumes stay root-owned.
3. Resolve the 13 Sentry issues after deploy.


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**Scope:** `promote: develop → staging` — 39 files, ~1000 lines (mostly docs plus accumulated commits). This is a promote PR, so the code below already landed on `develop` via its constituent PRs; the findings are structural debt to fix forward, not reasons to hold the promote. Multi-agent finder sweep across 7 dimensions, adversarially verified.

**Verdict: APPROVE.** No structural regressions or correctness blockers in the promoted code. Three maintainability items worth follow-up tickets.

### 🟡 Findings

**1. Prisma error-code set is enumerated twice in the same filter**
`apps/server/api/src/helpers/filters/database-exception/database-exception.filter.ts:29-38`
`isExpectedClientError` lists `P2002/P2003/P2025` for Sentry suppression; `mapPrismaError` case-matches the same three codes below. The two must be kept in sync by hand — add a new mapped code to `mapPrismaError` and it silently reaches Sentry. The sibling filters (`HttpExceptionFilter`, `AllExceptionFilter`) derive Sentry routing from the resolved status (`status >= 500`) instead. **Fix:** have `mapPrismaError` return an `isExpected` flag (or share a single `EXPECTED_CLIENT_PRISMA_CODES` set) so there is one source of truth.

**2. `ServiceLandingPage` slug prop is `string` — typos compile, crash at runtime**
`apps/website/packages/components/landing/ServiceLandingPage.tsx:29,35`
The PR swapped `config: ServiceLandingConfig` for `slug: string` to dodge the Next.js server→client serialization error on icon functions — a valid reason — but a bare `string` means a typo in any of the five page files compiles and fails at runtime. `serviceLandingConfigBySlug[slug] as ServiceLandingConfig` also adds a redundant cast: the `satisfies Record<…>` index already returns `ServiceLandingConfig`, and the cast would mask `undefined` if `noUncheckedIndexedAccess` is ever enabled. Each page now looks the config up twice (once for metadata, once in JSX). **Fix:** type the prop as `slug: keyof typeof serviceLandingConfigBySlug` and remove the cast — restores compile-time slug checking with the boundary fix intact.

**3. Production interceptor throws two different error shapes**
`packages/services/core/interceptor.service.ts:275-309`
`sanitizedError` is built as a real `Error` (`:278-282`) but thrown only on the fall-through path (`:308`); the JSON-API branch (`:303`) throws `response.data` — a plain object — so the constructed `Error` is dead in the common NestJS case. Callers branching on `instanceof Error` (e.g. `logger.service.ts`) behave differently depending on response format. The matching `IHttpSanitizedError` is still declared as a plain object while the runtime value is now an `Error`, which is precisely what the `as Error & IHttpSanitizedError` cast papers over. **Fix:** throw the constructed `Error` in both branches (attach the JSON-API `errors` array as a property), and make `IHttpSanitizedError extends Error` to match its sibling `IHttpInterceptorError`.

---
*Automated thermo-nuclear review: 7-dimension finder sweep + adversarial verification (only confirmed findings shown). Promote PR — these are forward-fix tickets, not promote blockers. No tests were run.*

PR #490, opened 2026-06-10
feat(workflow-engine): honor abortSignal cancellation + bound maxConcurrency (#479)

### LINKED ISSUE #479: Workflow engine: wire abortSignal cancellation + honor maxConcurrency
## Problem

Two capabilities are declared in the workflow engine but never wired:

### 1. `abortSignal` cancellation not checked

`ExecutionContext.abortSignal` is declared in the engine context but the `execute` loop in the engine never reads or checks it. Cancellation signals from callers are silently ignored — a cancelled execution continues to run all remaining nodes to completion.

Relevant file: `packages/workflow-engine/src/engine.ts` execute loop.

### 2. `maxConcurrency` unused

`EngineConfig.maxConcurrency` is set to `3` in the engine config but is never enforced in the execution scheduler. Concurrent node execution is unbounded, which can cause resource exhaustion on complex parallel workflow branches.

Relevant file: `packages/workflow-engine/src/engine.ts` scheduler / node dispatch.

Audit evidence: production-readiness audit, workflow engine area.

## Acceptance criteria

- [ ] `execute` loop checks `abortSignal.aborted` before dispatching each node; aborted executions return a `cancelled` terminal status
- [ ] `maxConcurrency` limits the number of concurrently executing nodes; additional nodes queue until a slot is free
- [ ] Unit test: trigger abort mid-execution, assert subsequent nodes do not run
- [ ] Unit test: workflow with 6 parallel branches and `maxConcurrency=3` executes no more than 3 nodes simultaneously
- [ ] No regression in existing workflow engine tests

VincentShipsIt:
## What

Wires up cooperative cancellation and concurrency bounding in the workflow execution engine (issue #479). The engine previously declared `ExecutionContext.abortSignal` and an `ExecutionStatus` of `cancelled`, but never populated the signal and ran every node strictly sequentially regardless of `maxConcurrency`.

## Changes

- **Wire `abortSignal` end-to-end.** `options.abortSignal` now flows into `ExecutionContext`, and is added to the `ExecutionOptions` contract in both `@genfeedai/workflows` (`contracts/execution.ts`) and the workflow-engine shim (the two are kept byte-identical).
- **Bounded ready-set scheduler.** Replaces the sequential `for…of executionOrder` loop. Ready nodes are dispatched in `executionOrder` priority up to `maxConcurrency`; free slots are reclaimed via `Promise.race` over in-flight work; a node whose dependency is still in flight is **deferred, not failed**. At `maxConcurrency: 1` this degrades to the previous strictly-sequential behavior (the telegram-bot caller relies on that).
- **Cancellation semantics.** When the signal fires, dispatch halts and the run returns `cancelled`. Abort takes **priority** over a concurrent in-flight failure during the drain phase — a cancelled run can no longer masquerade as `failed`.
- **No unobserved rejections.** Already-dispatched nodes are always drained on failure/abort; their progress/`onNodeStatusChange` callbacks are suppressed once the run is no longer healthy, so integrators don't see a `completed`/`progress: 100` event on a `failed` run.

## Review

This change was self-reviewed by an adversarial multi-agent pass (find → refute → fix). Three confirmed low-severity findings were folded in before this PR: abort-vs-concurrent-failure status masking, drain-phase callback leakage, and a redundant local alias.

## Verification

- `bun run test` (workflow-engine): **613 passed**. The 4 failing *files* are pre-existing `@genfeedai/enums` entry-resolution errors in unrelated saas-executor specs — 0 test assertions fail and they pre-date this branch.
- New specs: cancellation (sequential + concurrent-failing-sibling) and `maxConcurrency` bounding (never exceeds, actually reaches the limit).
- `tsc` typecheck (workflow-engine + workflows builds): clean.
- Biome check on all changed files: clean.

Closes #479

### DIFF
--- a/packages/workflow-engine/src/execution/__tests__/engine.spec.ts
+++ b/packages/workflow-engine/src/execution/__tests__/engine.spec.ts
@@ -597,6 +597,142 @@ describe('WorkflowEngine', () => {
     });
   });
 
+  describe('execute — cancellation', () => {
+    it('should stop dispatching and return cancelled when aborted mid-execution', async () => {
+      const controller = new AbortController();
+      const executed: string[] = [];
+      const abortingExecutor: NodeExecutor = vi.fn(async (node) => {
+        executed.push(node.id);
+        if (node.id === 'n1') {
+          controller.abort();
+        }
+        return { result: node.id };
+      });
+
+      engine.registerExecutor('generate', abortingExecutor);
+      engine.registerExecutor('upscale', abortingExecutor);
+      engine.registerExecutor('publish', abortingExecutor);
+
+      // Sequential chain so the abort lands before n2/n3 are dispatched.
+      const workflow = makeWorkflow(
+        [
+          makeNode('n1', 'generate'),
+          makeNode('n2', 'upscale'),
+          makeNode('n3', 'publish'),
+        ],
+        [makeEdge('n1', 'n2'), makeEdge('n2', 'n3')],
+      );
+
+      const result = await engine.execute(workflow, {
+        abortSignal: controller.signal,
+      });
+
+      expect(result.status).toBe('cancelled');
+      // Only the first node ran; later nodes were never dispatched.
+      expect(executed).toEqual(['n1']);
+      expect(result.nodeResults.has('n2')).toBe(false);
+      expect(result.nodeResults.has('n3')).toBe(false);
+    });
+
+    it('should report cancelled when an in-flight sibling fails after the signal fires', async () => {
+      // maxConcurrency 2 so n1 and n2 fill both slots while n3 is deferred.
+      const cancelEngine = new WorkflowEngine({ maxConcurrency: 2 });
+      const controller = new AbortController();
+
+      // n1 aborts the run the moment it runs, then succeeds immediately.
+      const aborter: NodeExecutor = vi.fn(async (node) => {
+        controller.abort();
+        return { result: node.id };
+      });
+      // n2 is already in flight when the abort fires; it fails during the drain.
+      const failer: NodeExecutor = vi.fn(async () => {
+        await new Promise((resolve) => setTimeout(resolve, 5));
+        throw new Error('sibling failed during drain');
+      });
+
+      cancelEngine.registerExecutor('generate', aborter);
+      cancelEngine.registerExecutor('upscale', failer);
+      cancelEngine.registerExecutor('publish', aborter);
+
+      // n1 + n2 occupy both slots; n3 is deferred so the dispatch phase re-runs
+      // after n1 settles and observes the abort, then n2 fails during drain.
+      const workflow = makeWorkflow([
+        makeNode('n1', 'generate'),
+        makeNode('n2', 'upscale'),
+        makeNode('n3', 'publish'),
+      ]);
+
+      const result = await cancelEngine.execute(workflow, {
+        abortSignal: controller.signal,
+        maxRetries: 0,
+      });
+
+      // Abort wins over the concurrent in-flight failure.
+      expect(result.status).toBe('cancelled');
+      // n3 was deferred and never dispatched once the abort was observed.
+      expect(result.nodeResults.has('n3')).toBe(false);
+    });
+  });
+
+  describe('execute — concurrency limit', () => {
+    it('should not run more than maxConcurrency nodes simultaneously', async () => {
+      let active = 0;
+      let maxActive = 0;
+      const releases: Array<() => void> = [];
+
+      const gatedExecutor: NodeExecutor = vi.fn(async () => {
+        active++;
+        maxActive = Math.max(maxActive, active);
+        await new Promise<void>((resolve) => {
+          releases.push(() => {
+            active--;
+            resolve();
+          });
+        });
+        return { result: 'ok' };
+      });
+
+      // engine from beforeEach is configured with maxConcurrency: 3.
+      engine.registerExecutor('generate', gatedExecutor);
+
+      // 6 independent branches (no edges) all eligible to run at once.
+      const workflow = makeWorkflow([
+        makeNode('n1', 'generate'),
+        makeNode('n2', 'generate'),
+        makeNode('n3', 'generate'),
+        makeNode('n4', 'generate'),
+        makeNode('n5', 'generate'),
+        makeNode('n6', 'generate'),
+      ]);
+
+      let done = false;
+      const runPromise = engine.execute(workflow).then((r) => {
+        done = true;
+        return r;
+      });
+
+      // Drain: each tick, release every currently-gated node so the scheduler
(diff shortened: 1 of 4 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**Scope:** `feat(workflow-engine): honor abortSignal cancellation + bound maxConcurrency` — 4 files, +342 lines, reviewed against `develop`. Multi-agent finder sweep across 7 maintainability dimensions, then adversarial verification; only confirmed findings appear below.

**Verdict: BLOCK.** Cancellation is wired in the engine but never reaches the database correctly — the feature ships half-applied. One hard blocker, one structural regression, plus simplifications the new scheduler should land *with*, not after.

### 🔴 Blocker

**1. Cancelled runs are persisted as `FAILED` — the change stops at the engine boundary**
`apps/server/api/src/collections/workflows/services/workflow-executor.service.ts:566,715,1011`

This PR makes `engine.execute()` return `status: 'cancelled'` for the first time (`engine.ts:360`). All three persistence sites map run status with an identical ternary that only handles `'completed'` and `'running'`, sending everything else to `FAILED`:

```ts
const finalStatus =
  result.status === 'completed'
    ? WorkflowExecutionStatus.COMPLETED
    : result.status === 'running'
      ? WorkflowExecutionStatus.RUNNING
      : WorkflowExecutionStatus.FAILED; // 'cancelled' lands here
```

`WorkflowExecutionStatus.CANCELLED` already exists (`packages/enums/src/workflow.enum.ts:65`) and this file already owns the canonical mapping helper `mapNodeStatus` (`:2527`). So from the moment this ships, every cancelled run is recorded as `FAILED`, corrupting audit/analytics data — and the PR never touches the service, leaving the engine change stranded. This is exactly the "related updates left half-applied" failure mode.

Secondary: `engine.ts:341` assigns `lastError = result.error` unconditionally on any node failure. After an abort, a drain-phase failure sets it before `currentStatus` is overwritten to `'cancelled'`, so a cancelled run's `error` field carries a misleading drain message into the DB record.

**Fix:** Add a `mapEngineRunStatus` switch mirroring `mapNodeStatus`, with `case 'cancelled': return WorkflowExecutionStatus.CANCELLED;`, and call it at all three sites. Guard the `lastError` assignment so a cancelled run's error field is not polluted by drain failures. Cancellation must round-trip engine → service → DB in this PR.

### 🟠 Structural regression

**2. Drain-after-abort emits spurious `failed` node events — asymmetric with the completed branch**
`packages/workflow-engine/src/execution/engine.ts:339-353`

The PR introduces an abort-suppression invariant (comment at `:307-310`): in-flight siblings drained after an abort must not emit user-facing status events. The completed branch honors it (`:311` guards with `currentStatus !== 'failed' && !wasAborted`). The failed branch does not — `emitNodeStatusChange({ newStatus: 'failed' })` fires unconditionally at `:344`. During an abort drain, some nodes silently disappear from the stream while failing siblings still push `running → failed` to subscribers, even though the run resolves to `cancelled`. Callers observing `onNodeStatusChange` get an inconsistent picture.

**Fix:** Apply the same guard to the failed branch — gate only the `emitNodeStatusChange` call; keep `failedNodes.add`, `lastError`, and `currentStatus` mutations unconditional. If drain-phase failures need to be observable, expose them through a structured result field, not the user-facing node-status stream.

### 🟡 Simplifications (land with the scheduler, not after)

**3. `workflows-contracts-shim.ts` is a 140-line manual copy, and this PR ratifies the double-write**
`packages/workflow-engine/src/workflows-contracts-shim.ts` vs `packages/workflows/src/contracts/execution.ts`
The shim has zero imports — it is a verbatim transcription of the 113-line canonical contract. This PR adds `abortSignal?: AbortSignal` to **both** files by hand, making the two-write obligation load-bearing with no compiler enforcement. `@genfeedai/workflows` has no dependency on the engine (no cycle), and the sibling `packages/workflow-saas` already demonstrates the correct pattern. **Fix:** add `@genfeedai/workflows` as a workspace dep, replace the shim body with `export * from '@genfeedai/workflows/contracts'`, and relocate or drop the three orphaned `SaaS*` types it carries.

**4. Drop the `wasAborted` shadow flag — it duplicates authoritative state and hides an edge case**
`engine.ts:197` (read at `220,276,311,359`)
`wasAborted` is set only inside the dispatch inner-loop (`:225`), which runs only while `remaining` is non-empty. If the signal fires *after* all nodes are dispatched but before in-flight work settles, `wasAborted` stays `false` and the run resolves to `completed` instead of `cancelled`. `AbortSignal.aborted` is monotonic and `ExecutionStatus` already includes `'cancelled'`. **Fix:** eliminate the shadow — either set `currentStatus = 'cancelled'` directly at `:225` (guarding the drain-failure assignment so abort wins over a late failure), or derive from `context.abortSignal?.aborted` via one `isAborted()` helper. Either collapses the four compound guards and closes the edge case.

**5. `execute()` is now 312 lines — extract the bounded-concurrency scheduler**
`engine.ts:68-379`
The new queue / `inFlight` map / drain loop is ~158 lines embedded inline. Siblings (`retry-handler.ts`, `resume-handler.ts`, `partial-execution.ts`) all export pure functions; this PR introduces comparable complexity without following that pattern. **Fix:** extract `concurrency-scheduler.ts` exposing `runScheduled(...)` that returns `{ nodeResults, completedNodes, failedNodes, aborted, lastError }`, passing the class methods as a small deps object, and co-locate the abort/cap tests there.

**6. The "run is healthy" guard is scattered across three sites in two phrasings**
`engine.ts:220,276,311` — `currentStatus !== 'failed' && !wasAborted` and its negation. Add a `'timed_out'` status later and all three need hand-edits; the `!== 'failed'` form is also accidentally-correct negative logic. **Fix:** extract one positive-form predicate (`isRunActive()`) and reuse it, so a new terminal status extends only the predicate.

**7. Dead branch**
`engine.ts:279-281` — `if (remaining.length === 0) break` inside the `inFlight.size === 0` block is unreachable: the outer `while` guarantees `remaining.length > 0` there. Delete it.

**8. `abortSignal` copied onto `ExecutionContext` widens the public executor contract**
`engine.ts:31-36,86` — the field is read only at `:224`, where `options.abortSignal` is already in scope. `ExecutionContext` is delivered to every executor; none honor cancellation, so the optional field advertises support that does not exist. **Fix:** read from `options`, drop the field; pass the signal explicitly to resolvers when executors genuinely opt in.

**9. Name the `inFlight` payload type**
`engine.ts:209-216` — `Map<string, Promise<{ node; nodeId; result }>>` is the scheduler's central unit yet left anonymous, inconsistent with `ExecutionContext` / `NodeExecutionResult`. Give it a `DispatchedNodeResult` interface.

---
*Automated thermo-nuclear review: 7-dimension finder sweep + adversarial verification (only confirmed findings shown). No tests were run — correctness claims are from static reading of the PR head. Item 1 is the one change required before merge.*

PR #505, opened 2026-06-10
refactor(config): consolidate service ConfigService classes onto createServiceConfig factory

### LINKED ISSUE #484: Backend services: consolidate 5 per-service ConfigService classes onto shared base
## Problem

Five backend services each define their own `ConfigService` class that extends a base and composes Joi validation schemas from shared fragments. The pattern is identical across all five; only the set of included schema fragments differs:

- `apps/server/workers/src/config/config.service.ts`
- `apps/server/notifications/src/config/config.service.ts`
- `apps/server/files/src/config/config.service.ts`
- `apps/server/mcp/src/config/config.service.ts`
- `apps/server/clips/src/config/config.service.ts`

Current consequence of the duplication: `clips/config/config.service.ts` is missing `microservicesSchema`, so `GENFEEDAI_MICROSERVICES_FILES_URL` is never validated at startup — the clips service silently falls back to `http://localhost:3012` in production (see related bug in `clips/services/transcription.service.ts`). The same drift risk exists for all five services whenever a new required env var is added.

Audit evidence: production-readiness audit, area "Backend services — workers, notifications, files, mcp, clips", consolidation section.

## Proposed approach

Extract a `createServiceConfig(schemaFragments[])` factory in `packages/config/` (the shared config package already exists). Each service calls the factory with its required fragments — eliminating the boilerplate class body and ensuring new required vars propagate to all services automatically.

## Acceptance criteria

- [ ] `createServiceConfig` factory (or equivalent) in `packages/config/`
- [ ] All 5 service `ConfigService` classes replaced with factory calls
- [ ] `clips` config includes `microservicesSchema` (fixing the silent localhost fallback)
- [ ] All services still pass `bun type-check` and startup validation
- [ ] No env var documented in any service schema is missing from any other service that logically needs it

VincentShipsIt:
Closes #484

## What

Replace 5 hand-rolled per-service `ConfigService` classes with a shared `createServiceConfig` factory in `@genfeedai/config`. The factory merges `baseSchema` + per-service Joi fragments + `extend` overrides via `Object.assign({}, baseSchema, ...schemas, extend)` and returns a constructable `BaseConfigService<T>` subclass.

## Changes

- **New factory** `createServiceConfig(options)` in `packages/config/src/services/create-service-config.ts`, exported from the package (with `CreateServiceConfigOptions`).
- **Migrated** `workers`, `notifications`, `files`, `mcp` config services onto the factory — each keeps its existing schema set and getters (notifications keeps its `logConfiguration()` constructor; files/workers keep `ingredientsEndpoint`).
- **clips bug fix**: clips previously extended the **unvalidated** `getDefaultConfigService()` (`get(k) => process.env[k] || ''`, no Joi), silently falling back to localhost for microservice URLs. It now extends the factory **with `microservicesSchema`**, so env is validated and a missing required key fails fast at boot. `PORT` is set on every clips deployment surface (`docker-compose.{staging,production,clips}.yml` → `PORT=${CLIPS_PORT:-3015}`), so this is intended fail-fast, not a regression.

## Acceptance criteria

- [x] `createServiceConfig` factory in `packages/config/`
- [x] All 5 service `ConfigService` classes replaced with factory calls
- [x] clips config includes `microservicesSchema`
- [x] All services pass `type-check` + startup validation

## Verification

- `bunx tsc --noEmit` clean on `@genfeedai/config` + all 5 services (only pre-existing TS5101/TS2688 unrelated to this change)
- Scoped specs green: factory **9**, base **14**, clips **10**, files **6**, mcp **5**, workers **11** = **55** tests
- `biome check` clean on all changed files; secretlint clean (lint-staged)

## Notes

- `notifications` config spec stays in its `vitest.config.ts` exclude list — pre-existing `RESEND_FROM_EMAIL` display-name vs `.email()` failure, identical in old code, out of scope.


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

Multi-agent review (5 parallel dimension finders + adversarial verification per finding; 27 agents total). Verified behavior preservation key-by-key for all 5 migrated services: the `Object.assign({}, baseSchema, ...schemas, extend)` precedence matches the old `Joi.object({ ...baseSchema, ...fragments, ...inline })` spreads exactly, the notifications `logConfiguration()` constructor path is intact, and the clips getters (`API_URL`, `FILES_URL`, `API_KEY`, `OPENROUTER_API_KEY`) are all preserved at the PR head. The factory earns its keep — it deletes real boilerplate rather than renaming a call. The findings below are about what the refactor left behind, ordered by severity.

### Major — structural

**1. The consolidation is incomplete and leaves three competing config patterns in the codebase.**

- `apps/server/{discord,slack,telegram,images,voices,videos}/src/config/config.service.ts` — six production services still extend the unvalidated `getDefaultConfigService()` from `packages/libs/config/default.config.ts` (`get(key) => process.env[key] || ''`, no Joi). That is the exact footgun this PR's own spec comment calls out as the clips pre-migration bug — and post-merge it still has six active callers while the fix covers one.
- `apps/server/api/src/config/config.service.ts:114-121` — the api service still uses the hand-rolled `Joi.object({ ...baseSchema, ... })` + `extends BaseConfigService` + constructor shape that `createServiceConfig` was built to kill. Migrating it is a mechanical transformation (specific schema object becomes `extend`, fragment imports become `schemas`).
- `packages/libs/config/default.config.ts:7` — carries no `@deprecated` tag and its JSDoc still recommends itself for "lightweight microservices (telegram, slack) that don't need full Joi validation," which is now actively misleading guidance that will onboard future services onto the footgun.

The issue #484 scope was five services, so the cut is understandable — but a half-applied migration is the most expensive state to live in. Concrete asks before or immediately after merge: (a) add `@deprecated Use createServiceConfig from @genfeedai/config` to `getDefaultConfigService` naming the six remaining callers, (b) open a tracking issue for migrating them plus api, and (c) treat the six unvalidated services as the priority — they have the bug this PR fixes for clips.

**2. `apps/server/clips/src/config/config.service.ts:20` — the `API_URL` localhost fallback is dead code in cloud mode and undocumented mode-dependent behavior in self-hosted.**

`return this.get('GENFEEDAI_API_URL') || 'http://localhost:3010'` made sense under the old unvalidated config. Now `genfeedaiMinimalSchema` makes the key `conditionalRequired`: in cloud mode (`GENFEED_CLOUD` set) the constructor throws before the getter can ever fall back — the `||` branch is unreachable and misleads readers into thinking the service can boot without the URL. In self-hosted mode the key is `.optional().allow('')`, so the fallback silently fires and becomes the sole guard for a gap validation deliberately left open. Two different behaviors per mode, zero comments distinguishing them. Either delete the fallback and let fail-fast validation do its job, or document the self-hosted-only reachability explicitly.

### Minor — simplification opportunities

**3. `apps/server/clips/src/config/config.service.ts:13` — `microservicesSchema` couples clips to keys it doesn't use.** Of its three keys, `GENFEEDAI_MICROSERVICES_MCP_URL` and `GENFEEDAI_MICROSERVICES_NOTIFICATIONS_URL` have zero references in the clips source tree; only `_FILES_URL` is read, and that getter carries its own `|| 'http://localhost:3012'` fallback, so the fragment adds no enforced contract. Drop the fragment and add the single `_FILES_URL` key to `extend` if validated coupling is ever actually wanted.

**4. `packages/config/src/services/create-service-config.ts:22-33` — `schemas` and `extend` are structurally identical options.** Both are plain `Record<string, Joi.Schema>` spreads differing only in merge position; nothing enforces the shared-fragment vs service-specific convention, and `extend` silently wins on collision. The override is documented and tested, which mitigates — but a cheap non-production collision guard (throw on duplicate keys between `schemas` entries and `extend`) would turn the silent precedence into an explicit decision.

**5. `packages/config/src/services/index.ts` — orphaned barrel.** Both the pre-existing and the newly added exports in `packages/config/src/index.ts` import service files by direct path, so the sub-barrel has zero consumers and the two files have already diverged. Delete it, or make `index.ts` consume it — pick one pattern.

**6. `apps/server/workers/src/config/config.service.ts:48-53` — `isLocalDevFlagEnabled` is duplicated verbatim as a private method in workers and `apps/server/api/src/config/config.service.ts:141-151`.** Pre-existing, but this boilerplate-elimination refactor was the natural moment to lift it onto `BaseConfigService` as `protected`. The next `GF_DEV_*` flag will produce a third copy.

**7. `apps/server/workers/src/config/config.service.spec.ts:264-276` — the new `ingredientsEndpoint` test enshrines a smell instead of fixing it.** `GENFEEDAI_CDN_URL` is absent from workers' schema and reaches the getter only via `allowUnknown` passthrough; missing env yields the string `undefined/ingredients` at runtime. Pre-existing gap, but the test now guards the passthrough as if it were a contract. The files service shows the correct pattern (`GENFEEDAI_CDN_URL: Joi.string().uri().required()` in `extend`) — follow it.

**8. `packages/config/src/services/create-service-config.spec.ts:452-481` — the two `workingDir` tests assert on `existsSync` path prefixes (`'../../'`, `'apps/server/sample/'`) that are internals of `BaseConfigService.loadEnvFiles`, not outputs of the factory.** A path-resolution change in the base class breaks the factory spec while the factory's forwarding contract is unchanged. Shallow coupling, but worth knowing it's there.

**9. `apps/server/clips/src/config/config.service.spec.ts` — cloud-mode validation is never exercised.** `IS_SELF_HOSTED_FLAG` is frozen at module import, so the suite only ever tests the self-hosted (everything-optional) branch of `genfeedaiMinimalSchema`. The required-in-cloud behavior of `GENFEEDAI_API_URL` — the headline of the clips fix — has no test. Add one that loads the module with `GENFEED_CLOUD` set and asserts construction throws without the URL.

### Verdict: **APPROVE**

No blockers. The factory is a genuine simplification — five hand-rolled schema/constructor stanzas collapse into declarative option objects, clips gains real fail-fast validation, and merge semantics are behavior-preserving (verified fragment-by-fragment). Approval is conditional in spirit on finding 1: ship the `@deprecated` tag and the tracking issue for the seven remaining services, or this PR's main achievement is adding a third config pattern rather than converging on one.

VincentShipsIt:
## 🔥 Thermo-Nuclear Review

Re-review of head `f1ea1529` (merge of develop), superseding the [previous review](https://github.com/genfeedai/genfeed.ai/pull/505#issuecomment-4668307964) of `7115183d`. Multi-agent pass: 5 parallel dimension finders (merge integrity, factory-vs-develop interaction, prior-finding status, fresh simplification, behavior preservation) with adversarial 3-lens verification per finding; 24 agents total, zero findings rejected. The merge resolution itself introduced a regression that flips the verdict.

### Blocker — new since the previous review

**1. `apps/server/workers/src/config/config.service.ts:18` — the merge resolution silently reverts develop's #472 fix by dropping `microservicesSchema` from workers.**

Develop commit `56bca2e9` ("fix(services): stop posting internal API calls to hardcoded localhost", #472) deliberately added `microservicesSchema` to the workers schema so `GENFEEDAI_MICROSERVICES_FILES_URL` gets a Joi `.default('http://localhost:3012')` in self-hosted mode and `.uri()` validation in cloud mode. The merged head's factory call reads `schemas: [postgresSchema, redisSchema, sentryOptionalSchema]` — the fragment is gone.

This is not cosmetic. `BaseConfigService.validateInput` runs Joi with `allowUnknown: true`, which passes through keys that are *present* in the environment but only injects defaults for keys *declared in the schema*. With the fragment dropped, `configService.get('GENFEEDAI_MICROSERVICES_FILES_URL')` returns `undefined` when the env var is unset, and two active consumers regress:

- `clip-factory.processor.ts:226-233` throws `GENFEEDAI_MICROSERVICES_FILES_URL is not configured — clip factory cannot reach the files service` whenever the value is falsy outside development. A self-hosted production deployment with the var unset worked on develop (Joi default) and now crashes `downloadAudio`.
- `clip-analyze.processor.ts:188-190` falls back to the hardcoded `'http://localhost:3012'` string — the exact silent void-posting behavior #472 was written to eliminate.

The merge commit message justifies the resolution with two claims that are both factually wrong: develop's `workersSchema` block did not reference "an undefined baseSchema" (it imports `baseSchema` from `@genfeedai/config`, same as every other service), and `microservicesSchema` was not "unused" (it is spread into `workersSchema` on develop). The factory's `extend` does **not** capture the same rules — it only carries `GF_DEV_ENABLE_SCHEDULERS`.

Fix is one line plus an import: `schemas: [microservicesSchema, postgresSchema, redisSchema, sentryOptionalSchema]`. Then add a workers spec test asserting that with `GENFEEDAI_MICROSERVICES_FILES_URL` unset, `get()` returns `'http://localhost:3012'` in self-hosted mode, so the next merge cannot drop it silently again.

### Major — carried over, still open

**2. The consolidation still leaves three competing config patterns** (previous review, finding 1). Verified unchanged at the merged head: discord, slack, telegram, images, voices, videos still extend the unvalidated `getDefaultConfigService()`; `apps/server/api/src/config/config.service.ts` still hand-rolls `Joi.object({...})` + `BaseConfigService`; `packages/libs/config/default.config.ts` still carries no `@deprecated` tag and still recommends itself for new services. The asks stand: deprecation tag, tracking issue, and priority on the six unvalidated services.

**3. `apps/server/clips/src/config/config.service.ts:20` — `API_URL` localhost fallback is dead code in cloud mode, undocumented sole guard in self-hosted** (previous review, finding 2). Unchanged.

### New minor findings (both in the notifications spec added by this PR)

**4. `apps/server/notifications/src/config/config.service.spec.ts:20` — the suite never sets `PORT`, so it only passes when the shell environment happens to provide one.** `baseSchema` declares `PORT: Joi.number().required()` and `createServiceConfig` always merges it; every other migrated spec sets `PORT` explicitly (clips `'3015'`, workers `'3000'`, files `'3002'`, mcp `'3014'`). On a clean CI runner without an ambient `PORT`, every `new ConfigService()` in this file throws a validation error. Add `process.env.PORT = '3011'` to `beforeEach` and delete it in `afterEach`, matching the sibling specs.

**5. `apps/server/notifications/src/config/config.service.spec.ts:4` — `vi.mock('fs')` never intercepts anything, because `BaseConfigService` imports `node:fs`.** The bare `'fs'` and protocol-prefixed `'node:fs'` specifiers are distinct module IDs to Vitest, so the declared mock is dead code; the test only works because line 17 imperatively mutates `existsSync` on the cached `node:fs` singleton. That survives today's shared module cache and breaks the moment isolation tightens. The same PR already contains the correct pattern — `vi.hoisted` + `vi.mock('node:fs', ...)` in `create-service-config.spec.ts:5-15`. Use it.

### Minor — carried over, all verified still valid at the merged head

Findings 3-9 from the previous review stand unchanged; one-line status for each:

- clips still pulls all three `microservicesSchema` keys while reading only `_FILES_URL` (finding 3)
- `schemas`/`extend` still collide silently with `extend` winning, no duplicate-key guard (finding 4)
- `packages/config/src/services/index.ts` is still an orphaned barrel with zero consumers (finding 5)
- `isLocalDevFlagEnabled` is still duplicated verbatim in workers and api (finding 6)
- the workers `ingredientsEndpoint` test still enshrines the `allowUnknown` passthrough instead of declaring `GENFEEDAI_CDN_URL` in the schema like the files service does (finding 7)
- the `workingDir` tests still assert on `loadEnvFiles` path internals rather than the factory's forwarding contract (finding 8)
- clips cloud-mode validation is still untested; the new `startup validation` block added via the merge only exercises the generic `PORT` requirement, not the `GENFEEDAI_API_URL` cloud-mode requirement that is the headline of the clips fix (finding 9)

### Verdict: **BLOCK**

Explicit blocker: finding 1. The factory refactor itself remains a genuine simplification (re-verified key-by-key at the merged head for clips, files, mcp, and notifications — all key sets, defaults, required-ness, and the notifications `logConfiguration()` constructor side-effect match develop exactly). But as merged, this branch ships a silent partial revert of a develop production fix: workers loses the schema half of #472 while keeping the processor half that depends on it. Restore `microservicesSchema` to the workers `schemas` array (and ideally fix the two notifications spec defects while in there), and this goes back to APPROVE.

PR #510, opened 2026-06-10
fix(workflow-engine): ReDoS-proof regex evaluation via recheck static analysis (#475)

### LINKED ISSUE #475: Security: ReDoS-proof regex evaluation in workflow condition executor
## Problem

Two files evaluate user-supplied or integration-supplied regex strings at runtime with no protection against catastrophic backtracking (ReDoS):

- `packages/workflow-engine/src/executors/condition-executor.ts` — interprets condition node regex patterns in workflow execution
- `apps/server/api/src/services/integrations/twitter/twitter-social.adapter.ts` — applies regex to tweet text matching

An interim length cap was added to reduce the attack surface but does not prevent polynomial backtracking on carefully crafted inputs.

Audit evidence: production-readiness audit, workflow engine area.

## Proposed remediation

Evaluate one of:

1. **`re2`** (Google's RE2 library, linear-time guarantees, Node bindings available) — drop-in for most patterns, no backtracking
2. **`safe-regex`** npm package — static analysis to reject dangerous patterns before execution

## Acceptance criteria

- [ ] `re2` or `safe-regex` evaluated and decision documented in a comment on this issue
- [ ] Chosen approach applied in `condition-executor.ts`
- [ ] Chosen approach applied in `twitter-social.adapter.ts`
- [ ] Unit tests covering known ReDoS patterns (e.g. `(a+)+b` on long `a` strings) pass within 10 ms
- [ ] Interim length cap kept or upgraded as part of the fix

VincentShipsIt:
## Summary

Replaces `safe-regex2`'s quantifier-count heuristic with [`recheck`](https://github.com/makenowjust-labs/recheck)'s automaton-based ReDoS decision procedure in `createSafeRegExp` (`packages/workflow-engine/src/utils/safe-regex.ts`).

Closes #475

### Why the heuristic had to go

Adversarial review of the safe-regex2 approach found it broken in both directions:

- **False negative (critical):** `'.*'.repeat(25) + 'Z'` is star-height 1 with exactly 25 repetition nodes — it passes the default `limit: 25` check, yet blocks the event loop for **127+ seconds** against a 10k-char non-matching subject (polynomial degree 26).
- **False positive:** `'a?'.repeat(26)` is completely harmless but gets rejected by the same counter.

No reps-limit value fixes both: k overlapping `.*`s cost ≈ n^(k-1) ops, so even k=4 at a 10k subject is 10^12 operations.

### What recheck does instead

`checkSync(pattern, flags, { checker: 'automaton', timeout: 500 })` runs a real ambiguity analysis and returns the attack complexity class. Policy:

| Verdict | Action |
|---|---|
| `safe` (constant/linear) | allow |
| `vulnerable` polynomial **degree ≤ 2** | allow — see below |
| `vulnerable` polynomial degree ≥ 3 | reject (null) |
| `vulnerable` exponential | reject (null) |
| `unknown` (invalid / timeout / unsupported) | reject (null) |

**Degree-2 allowance:** any pattern starting with an unbounded quantifier (e.g. `\d+ features`, email patterns) classifies as degree 2 because the engine re-scans from each start position — rejecting degree 2 would break virtually every realistic user pattern. The existing 10k-char subject cap (`SAFE_SUBJECT_MAX_LENGTH`) bounds the quadratic worst case to ~141ms measured.

### Implementation notes

- Verdicts cached per pattern+flags (FIFO, 1000 entries) — analysis runs once (~1–450ms warm), cached rejection <1ms. A **fresh RegExp instance** is returned per call so `g`/`y` flag `lastIndex` state is never shared between callers.
- Behavioral semantics unchanged at call sites: condition executor `matches` returns `false` for unsafe patterns; Twitter adapter regex matchMode falls back to `includes()`.
- `recheck@4.5.0` pinned exact. Heads-up: its `optionalDependencies` include platform-native binaries plus `recheck-jar` (~22MB) which always installs — pure-JS fallback, no native build step, Bun-compatible.
- Known limitation: a few complex patterns (e.g. `(\w+\s*,\s*){13,}`) return `unknown` even at 1000ms timeout and are rejected. Documented in JSDoc; fail-closed is the right default for user-supplied patterns.

## Verification

- `packages/workflow-engine/src/utils/safe-regex.spec.ts` — 13 passed (exponential + poly-chain + deg≥3 rejected; deg-2 + 26x `a?` allowed; fresh-instance-per-call; cached rejection <10ms)
- `packages/workflow-engine/src/executors/saas/condition-executor.spec.ts` — 27 passed
- `apps/server/api/.../twitter-social.adapter.spec.ts` — 16 passed
- `bunx turbo run build --filter=@genfeedai/workflow-engine` — 5 tasks successful
- api `tsc --project tsconfig.typecheck.json` — exit 0
- `biome check` on all changed files — clean (1 pre-existing `noExplicitAny` warning untouched)
- dist ESM smoke (node import of built output) — attack patterns null, safe patterns match


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

Multi-agent maintainability review of this PR: 6 parallel finders across review dimensions (abstraction quality, simplification, package boundaries, runtime risk, test quality, file-size/conditionals), followed by 3-lens adversarial verification of every finding. 17 raw findings were raised; 8 were refuted under verification and are not reported. The 9 confirmed findings are below, ordered by severity.

### Structural concern (should be addressed, not a merge blocker)

**1. `checkSync` blocks the event loop synchronously for up to 500ms per new unique pattern — and nothing in the code says so.**
[packages/workflow-engine/src/utils/safe-regex.ts:72](https://github.com/genfeedai/genfeed.ai/blob/issue-475-redos-safe-regex/packages/workflow-engine/src/utils/safe-regex.ts#L72)

`recheck`'s `checkSync` goes through synckit, which uses `Atomics.wait` to suspend the main thread until the worker responds or the 500ms timeout fires. Every cache miss freezes the entire event loop — no I/O callbacks, no microtasks, no timers — for up to `ANALYSIS_TIMEOUT_MS`. The old guard (`new RegExp()` in a try/catch) was instantaneous, so this is a qualitative behavioral change in a shared utility that runs inside the BullMQ workflow-execution processor (`concurrency: 3` — the other in-flight jobs stall too). A workflow with N distinct regex keywords pays N × 500ms of consecutive stalls on first evaluation. The PR description discusses analysis cost, but the `ANALYSIS_TIMEOUT_MS` JSDoc only says "this cost is paid once per unique pattern" without saying the cost is a synchronous event-loop block, and neither `createSafeRegExp`'s JSDoc nor either call-site comment mentions it.

To be clear about the trade-off: this is still strictly better than the pre-PR state, where a crafted pattern could block the loop for minutes. But the blocking behavior is the single most operationally surprising property of this util and it is invisible at every layer.

*Fix:* add an explicit warning to `createSafeRegExp`'s JSDoc ("first call per (pattern, flags) pair blocks the event loop for up to 500ms; subsequent calls hit the in-process cache"). Worth a follow-up: the Twitter adapter's polling path is already async — recheck's async `check()` with a promise-based cache would take the stall off the event loop entirely for that caller.

**2. The subject-length cap is part of the security contract but is enforced by every caller instead of the util.**
[packages/workflow-engine/src/utils/safe-regex.ts:26](https://github.com/genfeedai/genfeed.ai/blob/issue-475-redos-safe-regex/packages/workflow-engine/src/utils/safe-regex.ts#L26), condition-executor.ts (`String(actual).slice(0, SAFE_SUBJECT_MAX_LENGTH)`), twitter-social.adapter.ts (`text.slice(0, SAFE_SUBJECT_MAX_LENGTH)`)

Degree-2 polynomial patterns are only admitted *because* the subject is bounded — the JSDoc says exactly that. Yet `createSafeRegExp` returns a raw `RegExp` and trusts each call site to remember the slice. `SAFE_SUBJECT_MAX_LENGTH` is exported purely so callers can do the util's job for it. Any future third call site that imports `createSafeRegExp` and forgets the slice silently voids the safety guarantee, with no compile-time or lint signal.

*Fix:* add a `safeTest(pattern, subject, flags?): boolean` helper that compiles via `createSafeRegExp`, applies the slice internally, and returns `false` on null. Both existing call sites collapse to a single expression, and `SAFE_SUBJECT_MAX_LENGTH` can become module-private, tightening the public surface. (The Twitter adapter keeps its `includes()` fallback by checking `createSafeRegExp` for null first, or `safeTest` can be shaped to report rejection distinctly.)

**3. The five `performance.now() - start < 10` assertions test the host scheduler, not the cache.**
safe-regex.spec.ts, condition-executor.spec.ts (×2), twitter-social.adapter.spec.ts (×2)

The cached path is a `Map.get()` plus an early return — nanoseconds unloaded. On shared 2-core CI runners, GC pressure or a noisy neighbor can push any wall-clock measurement past 10ms; the pre-existing ReDoS test in condition-executor.spec.ts uses a 100ms budget for the same intent. Five 10ms assertions per run is a standing flake budget that will block PR gates for reasons unrelated to the code.

*Fix:* assert the behavior structurally — spy on `checkSync` and assert it is not called on the second invocation. That is deterministic under any CI load and tests the actual caching claim. (Failing that, raise the budget to 100ms to match the existing precedent.)

**4. `'exports correct constant values'` pins magic numbers — a snapshot of implementation shape.**
[packages/workflow-engine/src/utils/safe-regex.spec.ts:79-82](https://github.com/genfeedai/genfeed.ai/blob/issue-475-redos-safe-regex/packages/workflow-engine/src/utils/safe-regex.spec.ts#L79-L82)

`expect(SAFE_PATTERN_MAX_LENGTH).toBe(200)` adds no coverage — the boundary behavior is already exercised symbolically by the `'a'.repeat(SAFE_PATTERN_MAX_LENGTH + 1)` test. Its only effect is forcing whoever legitimately changes the cap to also edit this test. If the intent is a security floor, say so: `expect(SAFE_PATTERN_MAX_LENGTH).toBeGreaterThanOrEqual(200)`. Otherwise delete it.

**5. The `unknown`-verdict path — a security-critical rejection branch — has no test.**
packages/workflow-engine/src/utils/safe-regex.spec.ts

The spec covers `safe`, polynomial ≤2, polynomial ≥3, exponential, and the `SyntaxError` catch. But `status === 'unknown'` (analysis timeout / unsupported construct) is never exercised — the `[unclosed` test hits the catch block, not the unknown branch. If a recheck upgrade changes `unknown` handling, nothing regresses visibly. *Fix:* mock `checkSync` to return `{ status: 'unknown' }` and assert null.

### Nits

**6. Both call sites carry a 5-line ReDoS-guard comment that restates `createSafeRegExp`'s JSDoc.** Replace with a one-liner: `// ReDoS guard: see createSafeRegExp (#475)`. Rationale, policy table, and examples belong in one place.

**7. FIFO eviction in `verdictCache` carries a dead guard.** `keys().next().value` can only be `undefined` on an empty map, and the surrounding check guarantees ≥1000 entries. The six-line block collapses to `if (verdictCache.size >= VERDICT_CACHE_MAX_ENTRIES) { verdictCache.delete(verdictCache.keys().next().value!); }`.

**8. The warm-up + time + assert pattern is copy-pasted five times across three spec files.** This disappears naturally if finding 3 is fixed with spy-based assertions; otherwise extract a small local helper per file.

### Verified clean (raised and refuted under adversarial review)

- **Package ownership**: the Twitter adapter already imported from `@genfeedai/workflow-engine` pre-PR; adding `createSafeRegExp` to that relationship is not a layering change, and `packages/utils` would be a worse home (workflow-engine would gain a heavyweight dependency).
- **Null-contract divergence between callers** is deliberate, correct policy: `matches` is a boolean predicate (false is right), the adapter's `includes()` fallback is best-effort UX at the right boundary.
- **The try/catch around `new RegExp`** is legitimate defense-in-depth, not dead code — recheck's analysis is not a hermetic guarantee of JS-engine syntax validity.
- **Exact version pin** matches repo convention; **split type/value imports** match the existing style in `workflows.service.ts`; **fresh `RegExp` per call** matches pre-PR behavior and is the correct trade for `lastIndex` safety.
- **File sizes**: no changed file approaches the 1,000-line limit (largest post-PR: twitter-social.adapter.ts at ~690 lines; spec at ~522).

### Verdict: **APPROVE**

No structural blockers. The core design is sound and the PR is a strict improvement over the heuristic it replaces — the analysis policy is well-reasoned, the verdict cache with fresh-instance compilation is the right shape, and the adversarial test patterns (25× `.*` chain, 26× `a?`) show real rigor. Before merge, finding 1 deserves at minimum the JSDoc warning (the event-loop block is too surprising to leave undocumented), and finding 3 is worth fixing now rather than after the first CI flake. Findings 2, 4, 5 would tighten the module meaningfully; the nits are take-it-or-leave-it.

PR #516, opened 2026-06-10
refactor(pricing): consolidate pricing into @genfeedai/pricing single source of truth

### LINKED ISSUE #486: Pricing: single source of truth (helpers/core/config 3-way split + STRIPE_BYOK_FEE_PERCENTAGE env duplicate)
## Problem

Pricing values are defined in at least three separate locations with no canonical source:

1. `packages/helpers/src/` — pricing helper functions used by the frontend
2. `packages/core/src/` — core pricing constants
3. `packages/config/src/` — config-layer pricing (potentially sourced from env vars)

Additionally, `STRIPE_BYOK_FEE_PERCENTAGE` is defined as an env var but also appears as a hardcoded default in at least one of the above locations, creating a split where some code reads from the env and some uses the hardcoded fallback.

Audit evidence: production-readiness audit, pricing/config area.

## Risks of the current state

- A pricing change requires coordinated edits in 3+ files with no compile-time enforcement that they agree
- The env var vs hardcoded split means different services may charge different fee percentages silently
- Frontend may display a price that differs from what the backend charges

## Proposed fix

Establish one canonical pricing module (likely `packages/config/` or a new `packages/pricing/`):
- All price constants and fee percentages sourced from there
- `packages/helpers` and `packages/core` re-export from canonical source (no independent definitions)
- `STRIPE_BYOK_FEE_PERCENTAGE` read exclusively from env (with a documented default in config); no hardcoded duplicate

## Acceptance criteria

- [ ] Single canonical location for all pricing constants
- [ ] `packages/helpers` and `packages/core` pricing exports are re-exports, not independent definitions
- [ ] `STRIPE_BYOK_FEE_PERCENTAGE` read from env in exactly one place; all other usages import that value
- [ ] `bun type-check` passes
- [ ] A comment in the canonical file documents where to change prices and how to verify propagation

VincentShipsIt:
## Summary

Creates `packages/pricing` (`@genfeedai/pricing`) as the single canonical home for all pricing code, per issue #486:

- **`src/provider-pricing.ts`**: provider cost tables (`PRICING`), model types, node-type sets, and UI option arrays, moved verbatim from `packages/core/src/pricing.ts`
- **`src/plans-pricing.ts`**: BYOK constants, `applyMargin`, internal credit costs, website plans, credit packs, service/training offerings, and format helpers, moved verbatim from `packages/helpers/src/business/pricing/pricing.helper.ts`
- **`src/pricing-config.ts`**: `getPricingConfig` + `PricingConfig`, moved from `packages/config/src/pricing.ts`
- `src/index.ts` carries the canonical doc comment: where to change prices and how propagation is verified

The three old locations are now pure **named re-exports** with exact export-surface parity (16 symbols in core, 30 in helpers, 2 in config), so every existing import path keeps working unchanged.

### Duplicates eliminated

- `packages/config/src/schemas/stripe.schema.ts`: Joi defaults now reference `BYOK_FEE_PERCENTAGE` / `BYOK_FREE_THRESHOLD_CREDITS` instead of duplicated literals `5` / `500`
- `apps/server/api/.../byok-billing.service.ts`: removed 4 dead `|| 5` / `|| 500` fallbacks. Joi `.default()` guarantees values at startup, so the fallbacks were unreachable. **Edge-case behavior change**: an explicit env value of `0` is no longer silently coerced back to `5`/`500`.

### Wiring

`@genfeedai/pricing` path mappings and vitest aliases added everywhere `@genfeedai/core` or `@genfeedai/helpers` are compiled from source (apps/app, apps/website, apps/desktop, apps/server, and 8 packages). The new package follows the constants-style `tsc --build` composite pattern with zero runtime dependencies, so it is safe for both frontend and backend consumption.

Closes #486

## Verification

- [x] `packages/pricing` builds clean (`tsc --build`)
- [x] `turbo run build --filter=@genfeedai/helpers... --filter=@genfeedai/core...`: 8/8 tasks pass (incl. core tsup DTS)
- [x] `packages/core/src/pricing.spec.ts`: 20/20 pass (propagation check)
- [x] `packages/helpers/.../pricing.helper.test.ts`: 54/54 pass (propagation check)
- [x] `packages/config` builds + `__tests__/schemas.test.ts`: 97/97 pass (Joi defaults still 5/500 via canonical constants)
- [x] api typecheck: zero errors from changed files (note: `tsconfig.typecheck.json` has pre-existing failures on develop — missing `@genfeedai/tools`, `import.meta` module option — unrelated to this PR)
- [x] biome + secretlint via lint-staged pre-commit hooks
- [x] Adversarial multi-agent review: 6 candidate findings, 4 confirmed (missing vitest aliases in config/api, missing server tsconfig path), all fixed before commit

## Notes / follow-ups

- `apps/app/src/lib/replicate/client.ts:36` contains a separate local `PRICING` table duplicating provider costs; left untouched (out of scope for this issue, worth its own cleanup issue)
- `packages/config/src/pricing.ts` (`getPricingConfig`) currently has zero consumers; kept and re-exported rather than deleted to stay within issue scope
- Process note: implemented by scheduled autonomous run; worktree created manually via `git worktree add` (tooling limitation), branch `issue-486-pricing-sot` off `origin/develop`


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

This review was produced by parallel analysis across five dimensions (move parity, abstraction quality, complexity, wiring, residual duplication), with every finding adversarially verified before inclusion. Findings that turned out to be pre-existing debt or repo convention were rejected and are not listed, except where the PR's own stated goal makes them relevant.

### Verdict: BLOCK

The PR does what it says for the three locations it touches, and it does it carefully. But its stated goal is a *single source of truth* for pricing, and it leaves a fourth, divergent pricing table alive in production code. That contradiction is the blocker.

---

### Structural regressions

**1. Shadow `PRICING` table survives with divergent prices (blocker)**

`apps/app/src/lib/replicate/client.ts:36-73` contains an independent `PRICING` constant that this consolidation never absorbed. It is not a stale copy of the canonical table; it actively disagrees with it:

- `nano-banana-pro` at `2K`: **$0.20** in client.ts:45 vs **$0.15** in `packages/pricing/src/provider-pricing.ts:22`
- `llama`: **0.0001** in client.ts vs **0.0000095** canonical (a ~10x difference)

Both sides have tests asserting their incompatible values (`apps/app/src/lib/replicate/client.test.ts:60,75` on one side, the moved provider-pricing tests on the other), so the suite will happily green-light the divergence forever. This is not dead code: `apps/app/src/store/workflow/slices/persistenceSlice.ts:19` imports `calculateWorkflowCost` from client.ts, so this shadow table backs a live workflow cost path shown to users.

A PR titled "consolidate pricing into a single source of truth" that ships with two sources of truth disagreeing on real prices has not met its own bar. Suggested path: resolve each price discrepancy as a deliberate decision (is 2K nano-banana-pro $0.15 or $0.20?), delete the local `PRICING` const, import from `@genfeedai/pricing`, and update the client tests to assert against the canonical values.

**2. The canonical package doesn't own its primary types**

`PricingPlanProps` and `PricingOutputsProps` are declared as *private, unexported* interfaces in `packages/pricing/src/plans-pricing.ts:17,26`, while exported duplicates live in `packages/props/content/subscription.props.ts:25,38`, and the UI imports them from `@genfeedai/props` (e.g. `PricingCard.tsx:5`). To be fair: this duplication predates the PR (the old helpers file had the same private interfaces). But a consolidation PR that creates the canonical pricing package is exactly the moment to fix it, and instead the split ownership was moved verbatim into the new package. Suggested path: export both types from `@genfeedai/pricing` and have `@genfeedai/props` re-export them, so the package that computes pricing also defines its shape.

**3. `BYOK_CREDIT_VALUE_DOLLARS` exists but is bypassed by magic literals**

The constant is exported at `packages/pricing/src/plans-pricing.ts:100`, yet `creditPackPrice` computes `pack.credits * 0.01` at plans-pricing.ts:491, and `apps/server/api/src/services/byok-billing/byok-billing.service.ts` (a file this PR touches) uses raw `* 0.01` at lines 100 and 406. If the credit value ever changes, these three sites silently disagree with the named constant. Replace the literals with the constant.

**4. Named re-export lists in the shims are a maintenance trap**

The three compatibility shims enumerate symbols by hand: 16 names in `packages/core/src/pricing.ts:3-22`, 30 in `packages/helpers/src/business/pricing/pricing.helper.ts:3-36`, 2 in `packages/config/src/pricing.ts`. Verified the lists are currently exact, but every future addition to `@genfeedai/pricing` must now be mirrored in up to three places or consumers of the old paths silently miss it. `export * from '@genfeedai/pricing'` is safe here (no name collisions in any of the three shims) and removes the trap. Minor, but cheap to fix now.

**5. `packages/pricing` has no direct tests**

The package's behavior is only exercised through the old shim paths, and the comment in `packages/pricing/src/index.ts` saying propagation "is verified by" those tests slightly oversells it: `getPricingConfig` is untested anywhere (pre-existing gap, now inherited by the new package). Once the shims are eventually deleted, the package loses its coverage entirely. Suggest moving (or at least co-locating) the pricing tests into `packages/pricing` itself.

### What checked out clean

Credit where due, because the core mechanics of this PR are solid:

- The "moved verbatim" claims are true: blob IDs of the new files match the old ones exactly.
- Re-export parity is exact across all three shims (16/30/2 symbols, verified against the new package's exports).
- Removing the `|| 5` / `|| 500` fallbacks in `byok-billing.service.ts` is safe: the Joi schema's `.default()` values are applied via `schema.validate` in `BaseConfigService` (throws on error), and both the API and workers entrypoints construct config through that path. The change even fixes a real edge case where an explicit `0` was previously coerced to the fallback.
- tsup `external` entries, workspace deps, tsconfig path mappings, and `bun.lock` changes are all correctly scoped; no deep imports into the new package's internals; largest new file is 583 lines.

One aside, separate from this PR's scope: `scripts/manage-fal-models.ts` has its own inline margin math that also diverges from the canonical helpers; worth a follow-up ticket as part of the same consolidation effort.

### Blockers to clear

1. Absorb or explicitly justify the shadow `PRICING` table in `apps/app/src/lib/replicate/client.ts`, resolving the `nano-banana-pro` and `llama` price conflicts as deliberate decisions.
2. Export `PricingPlanProps` / `PricingOutputsProps` from `@genfeedai/pricing` so the canonical package owns its types (or state explicitly why the props package should remain the owner).

Items 3-5 are strongly recommended but not blocking. With the two blockers addressed this is an APPROVE: the migration mechanics themselves are careful and verifiably correct.

PR #518, opened 2026-06-10
feat(agent): non-zero creditCost floors for generation tools (#482)

### LINKED ISSUE #482: Agent tools: non-zero creditCost for generation tools
## Problem

Five agent tools that invoke paid generation APIs have `creditCost: 0` in the tool registry, meaning they are billed as free regardless of actual inference cost:

- `generate_image`
- `generate_video`
- `generate_music`
- `generate_voice`
- `generate_as_identity`

Source: `packages/tools/src/registry/source.ts`

This means every agent invocation of these tools incurs real compute / vendor API cost with zero credit deduction from the user's balance.

Audit evidence: production-readiness audit, agent tools area.

## Required action

1. **Pricing decision** — determine the credit cost per call for each generation tool (coordinate with billing/product)
2. **Implementation** — update `creditCost` in `packages/tools/src/registry/source.ts` for each tool
3. **Verification** — confirm the credit deduction flow fires for agent-initiated generation (the credit deduction BullMQ queue job is enqueued after tool execution)

## Acceptance criteria

- [ ] `generate_image`, `generate_video`, `generate_music`, `generate_voice`, `generate_as_identity` all have `creditCost > 0`
- [ ] Credit cost values documented (pricing decision logged in this issue or a linked doc)
- [ ] Integration test: agent tool call deducts the correct credit amount from org balance
- [ ] No generation tool has `creditCost: 0` in the registry unless it is explicitly a free-tier tool

VincentShipsIt:
Closes #482

## What

All five agent generation tools now carry a non-zero `creditCost` (minimum-charge floor), and the orchestrator no longer double-charges or silently skips the affordability gate.

### Credit cost decision (floors, 1 credit = $0.01)

| Tool | Floor | Basis |
|------|-------|-------|
| `generate_image` | 50 | `INTERNAL_CREDIT_COSTS.image` (4k variant bills 100 at the endpoint) |
| `generate_video` | 300 | `VIDEO_CREDIT_COSTS` minimum (4s clip) |
| `generate_music` | 10 | Conservative floor; actual = DB `Model.cost` per model |
| `generate_voice` | 17 | `INTERNAL_CREDIT_COSTS.voicePerMinute` |
| `generate_as_identity` | 100 | `INTERNAL_CREDIT_COSTS.avatarPerSecond` |

These are **floors for the pre-flight affordability gate**, not the billed amount. The downstream generation endpoints bill the real dynamic amount (CreditsGuard/CreditsInterceptor or direct deduction), so the orchestrator must not also deduct its flat cost.

### Changes

- **`packages/tools/src/registry/source.ts`**: the five generation tools get non-zero `creditCost` floors with comments documenting the basis.
- **`packages/interfaces`**: new optional `AgentToolResult.isBillingDelegated` flag.
- **`agent-tool-executor.service.ts`**: the six generation success paths (image incl. timeout fallback, video, music, voice, as_identity) return `isBillingDelegated: true`.
- **`agent-orchestrator.service.ts`** (both chat and stream paths):
  - Deduction guard: flat `creditCost` is only deducted when `result.success && creditCost > 0 && !result.isBillingDelegated`, preventing double-charging delegated generations.
  - **Gate bypass fix**: the `prepare_generation` remap mutated `toolName` *before* the `AGENT_CREDIT_COSTS` lookup, so `generate_image`/`generate_video`/`generate_as_identity` resolved a cost of 0 and skipped the affordability check entirely. The gate now uses `max(effective cost, requested tool's pre-remap cost)`; the deduction still uses the effective tool's cost (so the free prep card is never billed).

### Tests

- `agent-tool-registry.spec.ts`: asserts the five non-zero costs.
- `agent-orchestrator.service.spec.ts` (4 new integration tests):
  - blocks a generation tool call when the org cannot afford the flat cost (asserts `checkOrganizationCreditsAvailable(orgId, 10)`, no execution, tool error message)
  - gates on the requested tool's cost (50) when the call is remapped to `prepare_generation`
  - skips the flat deduction when the result is `isBillingDelegated` (turn cost only, `creditsUsed === 1`)
  - deducts the flat cost exactly once when billing is not delegated (`deduct(orgId, userId, 10, 'Agent tool: generate_music', …)`, `creditsUsed === 11`)

## Verification

- `bunx vitest run src/services/agent-orchestrator/tools/agent-tool-registry.spec.ts` — 4/4 pass
- `bunx vitest run src/services/agent-orchestrator/agent-orchestrator.service.spec.ts` — 38/38 pass
- `bunx biome check` on all 6 changed files — clean
- `packages/tools` and `packages/interfaces` `tsc --build` — clean (scoped typecheck)

## Out of scope (pre-existing, noted for follow-up)

- `generate_video`/`generate_music`/`generate_voice`/`generate_as_identity` executor handlers lack `generate_image`'s timeout fallback (try/catch returning `PROCESSING`). No wrong deduction is possible (the guard requires `success: true`), only reporting accuracy differs on timeout.
- `reframe_image` / `upscale_image` still have `creditCost: 0` while hitting paid endpoints.
- `videos.controller.ts` generate endpoint has a pre-existing double-deduct path; avatar endpoint deducts a hardcoded 1 credit (`avatar-video-generation.service.ts:173`).

### DIFF
--- a/apps/server/api/src/services/agent-orchestrator/agent-orchestrator.service.spec.ts
+++ b/apps/server/api/src/services/agent-orchestrator/agent-orchestrator.service.spec.ts
@@ -842,6 +842,269 @@ describe('AgentOrchestratorService', () => {
     );
   });
 
+  it('should block a generation tool call when the org cannot afford its flat credit cost', async () => {
+    organizationsService.findOne.mockResolvedValue({
+      onboardingCompleted: true,
+    } as never);
+    creditsUtilsService.checkOrganizationCreditsAvailable
+      .mockResolvedValueOnce(true) // turn-cost pre-check
+      .mockResolvedValueOnce(false); // generate_music flat-cost gate
+
+    llmDispatcher.chatCompletion
+      .mockResolvedValueOnce({
+        choices: [
+          {
+            message: {
+              content: null,
+              tool_calls: [
+                {
+                  function: {
+                    arguments: '{"prompt":"lofi beat"}',
+                    name: 'generate_music',
+                  },
+                  id: 'call-music-1',
+                },
+              ],
+            },
+          },
+        ],
+        usage: {
+          completion_tokens: 10,
+          prompt_tokens: 10,
+          total_tokens: 20,
+        },
+      } as never)
+      .mockResolvedValueOnce({
+        choices: [{ message: { content: 'Done!' } }],
+        usage: {
+          completion_tokens: 5,
+          prompt_tokens: 15,
+          total_tokens: 20,
+        },
+      } as never);
+
+    await service.chat(
+      { content: 'Make me a song' },
+      { organizationId: ORG_ID, userId: USER_ID },
+    );
+
+    expect(
+      creditsUtilsService.checkOrganizationCreditsAvailable,
+    ).toHaveBeenCalledWith(ORG_ID, 10);
+    expect(toolExecutorService.executeTool).not.toHaveBeenCalled();
+    expect(llmDispatcher.chatCompletion).toHaveBeenLastCalledWith(
+      expect.objectContaining({
+        messages: expect.arrayContaining([
+          expect.objectContaining({
+            content: expect.stringContaining(
+              'Insufficient credits. This tool requires 10 credits.',
+            ),
+            role: 'tool',
+          }),
+        ]),
+      }),
+      ORG_ID,
+    );
+  });
+
+  it('should gate on the requested tool cost when the call is remapped to prepare_generation', async () => {
+    organizationsService.findOne.mockResolvedValue({
+      onboardingCompleted: true,
+    } as never);
+    creditsUtilsService.checkOrganizationCreditsAvailable
+      .mockResolvedValueOnce(true) // turn-cost pre-check
+      .mockResolvedValueOnce(false); // generate_image pre-remap flat-cost gate
+
+    llmDispatcher.chatCompletion
+      .mockResolvedValueOnce({
+        choices: [
+          {
+            message: {
+              content: null,
+              tool_calls: [
+                {
+                  function: {
+                    arguments: '{"prompt":"a red fox"}',
+                    name: 'generate_image',
+                  },
+                  id: 'call-image-1',
+                },
+              ],
+            },
+          },
+        ],
+        usage: {
+          completion_tokens: 10,
+          prompt_tokens: 10,
+          total_tokens: 20,
+        },
+      } as never)
+      .mockResolvedValueOnce({
+        choices: [{ message: { content: 'Done!' } }],
+        usage: {
+          completion_tokens: 5,
+          prompt_tokens: 15,
+          total_tokens: 20,
+        },
+      } as never);
+
+    await service.chat(
+      { content: 'Make me an image' },
+      { organizationId: ORG_ID, userId: USER_ID },
+    );
+
+    // generate_image is remapped to prepare_generation (cost 0); the gate
+    // must still use the requested tool's cost of 50.
+    expect(
+      creditsUtilsService.checkOrganizationCreditsAvailable,
+    ).toHaveBeenCalledWith(ORG_ID, 50);
+    expect(toolExecutorService.executeTool).not.toHaveBeenCalled();
+    expect(llmDispatcher.chatCompletion).toHaveBeenLastCalledWith(
+      expect.objectContaining({
+        messages: expect.arrayContaining([
+          expect.objectContaining({
+            content: expect.stringContaining(
+              'Insufficient credits. This tool requires 50 credits.',
+            ),
+            role: 'tool',
+          }),
+        ]),
+      }),
+      ORG_ID,
+    );
+  });
+
+  it('should not deduct the flat credit cost when the tool delegates billing to its endpoint', async () => {
+    organizationsService.findOne.mockResolvedValue({
+      onboardingCompleted: true,
+    } as never);
+
+    llmDispatcher.chatCompletion
+      .mockResolvedValueOnce({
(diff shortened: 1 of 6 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

Reviewed at head `1f7621a`. The intent here is sound: generation tools need non-zero affordability floors so a broke user can't kick off a 300-credit video generation that the endpoint then fails to bill, and `isBillingDelegated` correctly prevents double-billing between the orchestrator's flat deduction and the endpoint's dynamic billing. The registry pin test on the five floor values is good. But the implementation has a confirmed gate-bypass bug, a key test that exercises a production-unreachable path, an entirely untested second copy of the billing logic, and that second copy should never have existed. Each finding below survived adversarial verification (3-lens review, majority confirm).

### Blockers

**1. The credit gate is silently bypassed for remapped generation tools — `agent-orchestrator.service.ts:1071-1109` (and the mirror in `runStreamLoop`)**

`preRemapToolName` is captured *after* unknown-tool recovery runs, not before. For any agent type whose `allowedToolNames` includes `PREPARE_GENERATION` but not the concrete generation tool — `AgentType.X_CONTENT` (`agent-type-config.constant.ts:83`) is a live example — a `generate_image` call is remapped to `prepare_generation` by `getUnknownToolRecovery` *before* the capture. So `preRemapToolName` is `prepare_generation`, `AGENT_CREDIT_COSTS['prepare_generation']` is 0, `preflightCreditCost` is 0, and the affordability gate this PR exists to add never fires for exactly the agent types most likely to hit it. The fix is small: `requestedToolName` is already in scope (lines 981 / 1738) and holds the genuinely pre-remap name — use `Math.max(creditCost, AGENT_CREDIT_COSTS[requestedToolName] ?? 0)` and delete `preRemapToolName`. Add a test with an `X_CONTENT`-style agent asserting the gate blocks an unaffordable `generate_image` request; none of the four new tests cover the remap path, which is why this shipped.

**2. The flat-deduction test exercises a path that cannot occur in production — `agent-orchestrator.service.spec.ts:1043-1105`**

"should deduct the flat credit cost exactly once when billing is not delegated" uses `generate_music` as its subject, but after this PR every real `generate_music` return site in `agent-tool-executor.service.ts` sets `isBillingDelegated: true` (lines 5767–7327 cover all six return sites across the generation tools). The test only passes because the mock omits the flag the real executor always sets. It is pinning behavior that is unreachable: a non-delegated `generate_music` does not exist. Rewrite it around a tool that genuinely takes the flat-deduction path (`create_post`, `ai_action` — creditCost 1, no flag). As written, the test would keep passing even if the flat-deduction branch were broken for the tools that actually use it.

**3. `runStreamLoop`'s billing logic has zero test coverage — all 4 new tests call `chat()` only**

The PR duplicates the full preflight-gate + deduction + delegation logic into `runStreamLoop` (~1773–1984), and not one test exercises it. The existing `chatStream` tests assert tool routing and stream-publisher calls but make no `creditsUtilsService` assertions. Elsewhere in this spec file the convention is symmetric chat/chatStream variants for shared behavior; billing — the one place where a silent regression costs real money — is the place that convention matters most. The streaming path is where the bug in finding 1 also lives, untested. Add streaming variants of the gate-block, delegated, and flat-deduction tests.

**4. ~25 lines of billing logic copy-pasted into both dispatch loops — `agent-orchestrator.service.ts:1068-1208` vs `1773-1984`**

The gate/deduction/delegation block is duplicated verbatim between `chat()` and `runStreamLoop()` in a file already at 6,607 lines. This file's own established pattern is shared private helpers for exactly this situation (`getGenerationPreparationOverride`, `buildUnknownToolRecoveryParams`, `getUnknownToolRecovery`). Findings 1–3 are direct consequences of the duplication: the bypass bug must now be fixed twice, and the test gap exists because the second copy was invisible. One honest caveat from verification: the two loops' `!canAfford` error paths genuinely differ (sync `recordToolCompleted` vs `publishStreamingToolCompletedEffect`), so don't force the error reporting into the helper — extract `resolveToolCreditBilling(requestedToolName, toolName)` returning `{ preflightCreditCost, creditCost }` and a `buildBillingDeduction(result, creditCost)` for the post-execution decision, and leave the divergent error emission at the call sites.

### Non-blocking

**5. The billing-delegation contract is split across three unlinked modules with no compile-time enforcement** — `source.ts` declares floors with a comment saying "billed dynamically by the endpoint", the executor sets `isBillingDelegated` at six return sites, and the orchestrator interprets the combination. Nothing stops a new generation tool from getting a floor in the registry but missing the flag in the executor (result: double-billing). A `billingMode: 'orchestrator_flat' | 'endpoint_delegated' | 'free'` discriminant on the tool definition would make the registry the single owner. Caveat: the runtime flag does carry information a static set can't — `generateImage` sets it `true` even on timeout/error paths — so if you keep the flag, at least add a registry-level test asserting every tool with a floor comment has executor return sites setting the flag.

**6. `AGENT_CREDIT_COSTS` is typed `Record<string, number>`** — any misspelled tool name silently resolves to the `?? 0` fallback, which is precisely the failure mode of finding 1. `Partial<Record<AgentToolName, number>>` would catch key typos at compile time and make the `?? 0` an explicit, visible decision.

### Verdict

**BLOCK.** Explicit blockers: (1) the preflight gate bypass for remapped generation tools — the central feature of this PR does not work for `X_CONTENT`-class agents; (2) the flat-deduction test pinning an unreachable path; (3) zero billing coverage on the streaming loop. Finding 4 (extract the shared billing helper) is strongly recommended in the same pass since fixing 1 in two places by hand is how the next divergence starts. Findings 1+2+3 are each individually cheap to fix; together they're a day of work, and this feature touches real money.

PR #522, opened 2026-06-11
fix(api): enable validation whitelist and protect DTO fields (#493)

### LINKED ISSUE #493: API: enable ValidationPipe whitelist (73 IsOptional-only DTO fields block the flip)
The custom ValidationPipe (`apps/server/api/src/helpers/pipes/validation.pipe.ts:90`) calls `validate(object)` with no options — unknown/extra body fields flow through to services untouched.

Flipping `whitelist: true` is currently unsafe: a sweep during the production-readiness remediation found **73 DTO fields whose only decorator is `@IsOptional()`** (e.g. `create-agent-strategy.dto.ts` policy objects, `create-workflow.dto.ts` nodes/edges, `create-reply-bot-config.dto.ts` configs) plus fully undecorated fields (`create-workflow-execution.dto.ts` `inputValues` — fixed in PR feat/fix-ssrf-redos-validation). class-validator strips fields with no validation metadata when whitelist is on — the flip would silently drop user data on ~25 endpoints.

Plan:
1. Add proper validators (or `@Allow()` where validation is genuinely deferred) to every flagged field — list reproducible via the sweep script in the audit notes.
2. Flip `validate(object, { whitelist: true })`.
3. Smoke: POST each affected endpoint with full payloads, assert round-trip.

From the 2026-06-09 production-readiness audit (validation cluster).

VincentShipsIt:
Closes #493

## What

Flips the API's custom global `ValidationPipe` to `validate(object, { whitelist: true })` so undeclared payload properties are stripped before reaching services, and hardens 31 DTO fields plus one query DTO so every previously-accepted payload keeps working.

## Key decisions

- **class-validator 0.15 semantics discovery**: the issue's premise (`@IsOptional`-only fields get stripped) held for 0.13/0.14 but **not** for the pinned 0.15.1 — `@IsOptional`-only properties now survive whitelist. Verified empirically and pinned in a spec test so a future class-validator bump that reverts this fails loudly. The decorator additions are therefore belt-and-braces plus genuine validation hardening, not strictly required for whitelist survival.
- **Decorator policy** (prime directive: payloads accepted before must still be accepted):
  - `@IsDate()` on `@Type(() => Date)` fields (agent-goals, agent-strategies incl. the issue-named `nextRunAt`/`dailyResetAt`/`dailyCreditResetAt`/`monthlyResetAt`)
  - `@IsBoolean()` only where the existing `@Transform` provably always returns a boolean (assets/gifs/images/videos query DTOs, posts `isRepeat`)
  - `@IsObject()`/`@IsString()`/`@IsNumber()`/`@IsArray()+@IsString({each:true})`/`@IsEnum()` where the wire type is unambiguous
  - `@Allow()` where the wire format differs from the declared type (Date fields without `@Type` arrive as strings, unions, `unknown`) — whitelists without over-constraining
  - No `@ValidateNested` added anywhere; no existing decorators removed
- **api-keys regression found by adversarial review**: `GET /api-keys` read `label`/`description` off an untyped cast of `BaseQueryDto` — whitelist would have silently dropped both and broken search. Fixed with a dedicated `ApiKeysQueryDto extends BaseQueryDto` (mirrors `AssetQueryDto` pattern) and removed the cast.
- **Intentionally not changed**: `UploadPostDto` (never imported by any controller — dead code), `RunEventEnvelopeDto` (swagger `@ApiResponse` only), `CreateWorkspaceTaskDto` (service-only, no controller). All three are unreachable from any controller param. Also `BaseCRUDController` generic bodies erase to `Object` at runtime so the pipe never validated them — pre-existing, unaffected by this PR.
- `forbidNonWhitelisted` deliberately **not** set: unknown keys strip silently rather than 400, keeping lenient behavior for existing clients.

## Verification

- `validation.pipe.spec.ts`: 20/20 pass, incl. new 7-test whitelist suite (unknown props stripped in plain + JSON:API paths, `@Allow`/`@IsOptional`-only/validated-optional kept, `@Type(() => Date)` converts, invalid values still 400)
- Biome clean on all 23 changed files
- Typecheck zero-delta vs develop baseline (stash-diff proof: sorted tsc output identical before/after, 0 changed lines; the 862 pre-existing errors are module-resolution noise from unbuilt workspace package dists)
- Deterministic AST sweep over `apps/server/api/src`: 0 controller-reachable DTO fields left without validation metadata (was 26)

### DIFF
--- a/apps/server/api/src/collections/agent-goals/dto/create-agent-goal.dto.ts
+++ b/apps/server/api/src/collections/agent-goals/dto/create-agent-goal.dto.ts
@@ -7,6 +7,7 @@ import { ApiProperty } from '@nestjs/swagger';
 import { Type } from 'class-transformer';
 import {
   IsBoolean,
+  IsDate,
   IsEnum,
   IsNumber,
   IsOptional,
@@ -48,11 +49,13 @@ export class CreateAgentGoalDto {
   @ApiProperty({ description: 'Target metric value' })
   targetValue!: number;
 
+  @IsDate()
   @Type(() => Date)
   @IsOptional()
   @ApiProperty({ description: 'Optional start date', required: false })
   startDate?: Date;
 
+  @IsDate()
   @Type(() => Date)
   @IsOptional()
   @ApiProperty({ description: 'Optional end date', required: false })
--- a/apps/server/api/src/collections/agent-goals/dto/update-agent-goal.dto.ts
+++ b/apps/server/api/src/collections/agent-goals/dto/update-agent-goal.dto.ts
@@ -6,6 +6,7 @@ import { ApiProperty } from '@nestjs/swagger';
 import { Type } from 'class-transformer';
 import {
   IsBoolean,
+  IsDate,
   IsEnum,
   IsNumber,
   IsOptional,
@@ -35,11 +36,13 @@ export class UpdateAgentGoalDto {
   @ApiProperty({ description: 'Target metric value', required: false })
   targetValue?: number;
 
+  @IsDate()
   @Type(() => Date)
   @IsOptional()
   @ApiProperty({ description: 'Optional start date', required: false })
   startDate?: Date;
 
+  @IsDate()
   @Type(() => Date)
   @IsOptional()
   @ApiProperty({ description: 'Optional end date', required: false })
--- a/apps/server/api/src/collections/agent-strategies/dto/create-agent-strategy.dto.ts
+++ b/apps/server/api/src/collections/agent-strategies/dto/create-agent-strategy.dto.ts
@@ -15,6 +15,7 @@ import { Type } from 'class-transformer';
 import {
   IsArray,
   IsBoolean,
+  IsDate,
   IsEnum,
   IsIn,
   IsNumber,
@@ -498,16 +499,19 @@ export class CreateAgentStrategyDto {
   })
   rankingPolicy?: RankingPolicyDto;
 
+  @IsDate()
   @Type(() => Date)
   @IsOptional()
   @ApiProperty({ description: 'Next scheduled run time', required: false })
   nextRunAt?: Date;
 
+  @IsDate()
   @Type(() => Date)
   @IsOptional()
   @ApiProperty({ description: 'Daily reset timestamp', required: false })
   dailyResetAt?: Date;
 
+  @IsDate()
   @Type(() => Date)
   @IsOptional()
   @ApiProperty({
@@ -516,6 +520,7 @@ export class CreateAgentStrategyDto {
   })
   dailyCreditResetAt?: Date;
 
+  @IsDate()
   @Type(() => Date)
   @IsOptional()
   @ApiProperty({ description: 'Monthly reset timestamp', required: false })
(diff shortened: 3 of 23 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

Reviewed at head `9731262`. This is a tight, well-aimed hardening change: enabling `whitelist: true` in the shared validation pipe and back-filling decorators so legitimate fields survive stripping is exactly the right shape, and the new pin test documenting class-validator 0.15's `@IsOptional`-only behavior is the kind of test that earns its keep. `ApiKeysQueryDto` deleting the `query as unknown as Record<string, string | undefined>` cast in the controller is a genuine contract improvement. The findings below were each independently confirmed by adversarial verification; none of them undo the value of the change, but two of them deserve action before or shortly after merge.

### Findings

**1. Production pipe is weaker than every test harness that claims to cover it — `apps/server/api/src/helpers/pipes/validation.pipe.ts:90`**

The pipe now runs `validate(object, { whitelist: true })`, but all five e2e harnesses construct their `ValidationPipe` with `{ whitelist: true, forbidNonWhitelisted: true }` (`e2e-test.utils.ts:200` and the auth/brands/tasks/organizations e2e specs). That means the e2e suite proves a *stricter* behavior (reject unknown props with a 400) than production actually has (silently strip them). Tests that pass against a stronger configuration than production runs are worse than no tests: they certify a contract that doesn't exist. Either add `forbidNonWhitelisted: true` to the production pipe (preferable for an API surface; silent stripping hides client bugs) or align the e2e harnesses down to match production. Pick one; the current split is the worst option.

**2. `@Allow()` on plain `Date` fields is a contract dodge, not validation — `create-workflow.dto.ts:329-351`, `create-context.dto.ts:50-56`**

`lastExecutedAt`, `startedAt`, `completedAt`, and `lastAnalyzed` get `@Allow()` (keep the property, validate nothing, coerce nothing) while the identically-typed `scheduledFor` four lines below gets the correct `@IsDate()` + `@Type(() => Date)` treatment. These DTOs are spread into Prisma, so a client can pass any JSON value into a `Date`-typed column path with zero checking. Decide what these fields actually are: if they're server-managed timestamps (which their names strongly suggest), they should not be client-settable on a create DTO at all — remove them from the DTO and let the service set them. If they are legitimately client-suppliable, validate and coerce them like `scheduledFor`. `@Allow()` is the worst of both worlds: it whitelists the field while explicitly opting out of the validation this PR exists to enforce.

**3. The boolean `@Transform` body is now copy-pasted in 11 DTO files — e.g. `gifs-query.dto.ts:101-121`**

The 12-line string-to-boolean transform appears verbatim across 11 query DTOs, and this PR adds `@IsBoolean()` to each copy rather than consolidating. Same story with the `lightweight` field, copy-pasted into 4 media query DTOs (`images-query.dto.ts:109-129` and siblings). To be fair: the duplication predates this PR, and `BaseQueryDto` is extended by 35+ non-media DTOs, so the right consolidation target is a shared `ToBoolean()` decorator helper plus a `BaseMediaQueryDto` for the media-specific fields — not cramming everything into the existing base. This PR touched all 11 sites anyway; that was the moment to extract the helper. Acceptable as a fast-follow, but it should actually follow.

**4. No regression test for the `@Allow()` escape hatch — `validation.pipe.spec.ts`**

The spec thoroughly covers whitelist stripping but has no test demonstrating that an `@Allow()`-decorated `Date` field accepts a non-Date value. Whichever way finding 2 is resolved, a test should pin the chosen contract — otherwise the next person to touch these DTOs has no signal that the distinction between `@Allow()` and `@IsDate()` is deliberate.

### Verdict

**APPROVE** — the core change is correct, well-tested where it matters most, and removes a real cast. Findings 1 and 2 are real contract gaps that should be resolved promptly (finding 1 ideally before merge, since it's a one-line decision); findings 3 and 4 are acceptable fast-follows.

PR #682, opened 2026-06-19
chore: add voice catalog reseed tooling

VincentShipsIt:
## Summary

Adds tooling to reseed the external voice catalog (ElevenLabs / HeyGen) plus two portable slash commands. **Purely additive** — no existing code, infra, or CI is modified.

- `scripts/migrations/reseed-voice-catalog.ts` — upsert-based reseed (idempotent, `ON CONFLICT DO UPDATE`), **dry-run by default** (requires `--live`)
- `scripts/migrations/reseed-voice-catalog.incontainer.mjs` — in-container variant for running against prod via `docker exec`
- `.claude/commands/bug.md`, `.claude/commands/release.md` — distribute the global slash commands into the repo so contributors get them
- `.codex/environments/environment.toml` — 1-line cosmetic display-name change

## Regression gate ✅

Adversarial review (3 lenses + skeptic verification) found **no regressions and no blockers**:
- No Terraform / CI / infra files touched → the merged Fargate routing, ECS image-wait, rollout-overlap fix, and retired-SSM-flag ignore are all unaffected.
- No files deleted; no existing symbols or references broken.
- All imports resolve (`normalizePgUrl`, `packages/prisma`, `@elevenlabs/elevenlabs-js`, `@prisma/adapter-pg`, `ExternalVoice` model + `externalProvider_externalId` unique key all confirmed present).
- **No secrets** — all credentials (`ELEVENLABS_API_KEY`, `HEYGEN_KEY`, `DATABASE_URL`) read from env at runtime.
- Reseed scripts are **non-destructive** (upsert only, no DELETE/TRUNCATE) and **not auto-run** by any CI/Dockerfile/postinstall hook.

## Reviewer notes (non-blocking)

1. **Live-mode asymmetry:** the `.ts` script defaults to dry-run, but `.incontainer.mjs` defaults to **live** (writes unless `--dry` is passed). Consider flipping the default or adding a confirm gate before merging if that's a concern.
2. The `.incontainer.mjs` run instructions reference `node _reseed.mjs` but omit the `docker cp` / rename step — an operator running it cold needs to know to copy + rename first.
3. `rejectUnauthorized: false` mirrors the existing `_pg-ssl.ts` / `sslmode=no-verify` RDS pattern (documented in `reference_postgres_rds.md`), just without the explanatory comment in the `.mjs` copy.

🤖 Opened via release-cleanup after confirming the branch's work was complete and not a regression.


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

Structural / maintainability pass on the voice-catalog reseed tooling. Methodology: parallel structural finders across all rubric axes, then adversarial verification of every finding against the checked-out branch — only confirmed findings appear below. These are dev/migration scripts, not app runtime, so severity is calibrated accordingly — but a reseed script that miscounts or silently drifts is still worth fixing.

None.

**1. Wrong-by-construction create/update counter — `reseed-voice-catalog.ts:240-247`.**
The "created vs updated" tally uses a timestamp-proximity heuristic: `Math.abs(result.createdAt.getTime() - result.updatedAt.getTime()) < 1000`. This is structurally unsound — it assumes both timestamps are written within 1ms, but any DB/network latency on a genuine create pushes the gap past the window and the row gets miscounted as an update. Worse, the sibling `.incontainer.mjs` already does this *correctly and more simply* via `RETURNING (xmax = 0) AS inserted`.
*Remedy:* drop the heuristic; use `$queryRaw` returning `xmax = 0` (matching the `.mjs`), or split into explicit create/update queries and count results. This is the "code-judo" case — the simpler form is also the correct one.

**2. Duplicated HeyGen provider logic across the two scripts — `reseed-voice-catalog.incontainer.mjs:56-76` vs `reseed-voice-catalog.ts:126-153`.**
The HeyGen fetch is a near-verbatim copy across both files: same URL, same header, same field-fallback chain (`voice_id ?? id`, `voice_name ?? name`, `preview_url ?? preview`), same `String()` coercions, same `providerData: { index }`, and the same shape-resolution branch (`Array.isArray(body?.data) ? body.data : (body?.data?.voices ?? [])`). The ElevenLabs split is justified (the container lacks the SDK), but the HeyGen path is a plain `fetch` with no such constraint — there's no reason for two copies that will silently diverge when the HeyGen response shape changes.
*Remedy:* extract the HeyGen fetch + normalization into one shared `.mjs` helper (e.g. `scripts/migrations/_voice-providers.mjs`) imported by both scripts, so the field mapping and shape resolution live in exactly one place.

**3. Single-use `createPrismaClient()` wrapper — `reseed-voice-catalog.ts:93-99`.**
A 6-line wrapper around `new PrismaPg(...)` + `new PrismaClient(...)`, called exactly once (line 168). It earns no reuse, no testability seam, and no invariant beyond what the two inline lines express.
*Remedy:* delete the function and inline the two lines in `main()`.

### Minor

**4. ElevenLabs field-name drift between the twins — `reseed-voice-catalog.incontainer.mjs:36-53`.**
The `.mjs` reads raw REST fields (`v.voice_id`, `v.preview_url`) while the `.ts` reads SDK-normalized camelCase (`voice.voiceId`, `voice.previewUrl`). Equivalent today, but the mapping is an undocumented assumption with no shared contract to catch a future REST-shape change.
*Remedy:* add a one-line comment in the `.mjs` asserting the REST↔SDK field equivalence, or fold it into the shared helper from finding #2.

**5. Uncommented cast chain — `reseed-voice-catalog.ts:137-143` and `:145`.**
Three successive `as` casts over the HeyGen JSON body with no type guard between them, plus a per-element `voice as Record<string, unknown>` in the `.map()`. None carry an explanatory comment.
*Remedy:* replace the chain with a single inline type guard (`if (!body || typeof body !== 'object') return []`) and derive from the narrowed value, or add a one-line comment on each cast stating the structural assumption.

### Approved axes (clean)

File size (all files well under threshold), spaghetti branching, canonical-layer discipline (the parallel-script duplication is justified by the container/SDK constraint, not a layer violation), non-atomic mutations, sequential orchestration, stack hygiene (no npm/npx/yarn/pnpm debris — clean) — no findings.

---

**Verdict: APPROVE.** No blockers; this is migration tooling. Before merge, fix finding #1 (the timestamp heuristic is wrong, and the correct form is shorter) and finding #2 (collapse the duplicated HeyGen path to one shared helper). Both are cheap and leave the tooling correct and single-sourced; the rest are minor cleanups.

PR #683, opened 2026-06-19
fix(mcp): MCP server end-to-end — transport + identity + approval queue (Phases 1, 2, 4)

VincentShipsIt:
The genfeed MCP server (`apps/server/mcp`, port 3014) was **non-functional end-to-end**. This PR delivers phases 1, 2, and 4 of the MCP audit fix — the server now serves requests, carries a real role-aware identity, *and* gates mutating tools behind human approval. (Phase 3 OAuth-discovery follows as a separate PR — it has a Clerk-dashboard ops prerequisite.)

## Phase 1 — HTTP transport (🔴 server was dead)

Every `POST /mcp` threw `'Transport did not generate a session ID'` and 500'd: the SDK transport was configured stateless (`sessionIdGenerator: undefined`) but the code read `transport.sessionId` and threw when falsy. Rewrote `StreamableHttpService` to the correct **stateless per-request** lifecycle (fresh `Server` + transport per request, `handleRequest(req, res, req.body)`, teardown in `finally`); removed the dead session map / GET-DELETE handlers / `getActiveSessionCount`.

## Phase 2 — identity foundation (🟠 role + tenant + connectivity)

Three layers were broken:

1. **`/v1` routing** — `GENFEEDAI_API_URL` is configured **without** the API's `/v1` global prefix in every env (`.env.example`, all docker-compose, terraform), yet both `ClientService` (every tool call) and `AuthService` called un-prefixed paths → **all 404'd**. Added a shared `resolveApiBaseUrl` that normalizes `/v1` exactly once.
2. **Identity** — `AuthService` hit a non-existent `/accounts` route and an `/api-keys/validate` that returned no identity. Unified **both** token types onto `GET /v1/auth/whoami` (the API's `CombinedAuthGuard` already handles `gf_` keys *and* Clerk JWTs); extended whoami to return the caller's org `role` via a `Member→Role` lookup.
3. **Role enforcement** — `main.ts` dropped `role` and the registry read a non-existent `clientService.currentRequest`, so the role was always `'user'` and the **14 admin-gated tools** (darkroom/training) were *always denied even for admins*. Now the role is threaded `whoami → authContext → buildServer → ToolRegistryService(requestRole)`; the dead cast is gone. `owner|admin → admin`, `superadmin → superadmin`, everything else (incl. no-membership) → `user` (deny-by-default for admin tools; the API re-enforces membership on actual calls).

**Tenant isolation:** each request's `ClientService` carries the caller's bearer token, so the API scopes every query to that token's org.

## Phase 4 — write-action approval queue (human-in-the-loop)

Mutating MCP tools no longer execute immediately. They persist a `PENDING` `McpApproval` and return an approval prompt; a human resolves it via the **admin-gated `resolve_approval`** tool, which executes the deferred action and persists its result. A user-tier agent cannot self-approve its own queued write.

- **prisma** — `McpApproval` model + `McpApprovalStatus` enum + migration (org/user/toolName/arguments/status/result; soft-delete; indexed on `(organizationId, status, isDeleted)` and `(organizationId, isDeleted, createdAt desc)`).
- **api** — new `mcp-approvals` collection (controller/service/DTOs/schema/module): org-scoped create/list/get/resolve; `createPending` publishes a best-effort `mcp_approval_pending` notification (publish failures swallowed, not fatal).
- **mcp** — `ToolRegistryService` approval gate + `resolve_approval` handler; `ClientService` create/get/resolve approval. Approving re-executes the original tool through normal dispatch (bypassing the gate) and persists the result on the approval row.
- **tools** — register `resolve_approval` (admin, credit cost 0).

Gated: `generate_content_batch`, `create_post`, `create_article`, `create_avatar`, `start_training`, `delete_dataset`, `control_comfyui`, `run_captioning`, `generate_face_test`, `generate_bootstrap`, `generate_pulid`, `generate_darkroom_content`.

## Verification

- ✅ `@genfeedai/mcp` — **131 tests** (incl. the Phase 4 approval queue: queue / decline / approve-execute-persist / refuse-already-resolved / missing-args)
- ✅ `@genfeedai/api` mcp-approvals — **18 tests** (service + controller: notification-failure swallow, cross-org NotFound, already-resolved BadRequest)
- ✅ `@genfeedai/mcp` — earlier phases (incl. a real-SDK transport integration test and a **real-guard** role-enforcement integration test proving a user is *denied* an admin tool and an admin is *allowed*)
- ✅ `@genfeedai/api` whoami — 20 tests · type-check + biome green
- ✅ Hardened after a **25-agent adversarial review** (6 confirmed findings fixed: tautological role tests, role-mapping/empty-role handling, and coverage gaps)

## Follow-ups (separate PRs)

- Phase 3 (OAuth discovery / RFC 9728 via Clerk — has a Clerk-dashboard ops prerequisite).
- Reviewer UI for the approval queue (REST endpoints are ready). A UI-driven `POST /mcp-approvals/:id/resolve` currently only flips status; tool execution happens through the MCP `resolve_approval` path — wiring API-side execution-on-approve would let a dashboard approve without the MCP round-trip.


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

Structural / maintainability pass on the MCP server end-to-end work (transport + identity + approval queue). Methodology: parallel structural finders across all rubric axes, then adversarial verification of every finding against the checked-out branch — only confirmed findings appear below. This review is orthogonal to correctness/test review; it targets structure, layering, and atomicity.

**1. Non-atomic approval resolution — `apps/server/api/src/collections/mcp-approvals/services/mcp-approvals.service.ts:113-138`.**
`resolve()` does a `findFirst` (to check `status === PENDING`) and then a separate `update`, with nothing between them. This is a classic check-then-act TOCTOU: two concurrent `resolve` requests for the same approval can both pass the PENDING check and both apply an update, double-resolving an approval that must resolve exactly once. For an *approval queue* — the security boundary that gates tool execution — a double-resolve is precisely the race you cannot ship.
*Remedy:* make it a single atomic write. Either wrap the read+update in a `$transaction`, or replace the read-check-write with an atomic `updateMany({ where: { id, status: 'PENDING' }, data: {...} })` and branch on the returned count (`0` → already-resolved → throw `BadRequestException`/`NotFoundException`). The second form is simplest and removes the race entirely.

**2. Serializer logic living in the controller — `mcp-approvals.controller.ts:24-53`.**
The `McpApprovalResponse` shape and its `toResponse()` DB-record-to-client mapping are defined inside the controller. This repo's invariant (CLAUDE.md, `system-patterns.md`) is explicit: serializers live in `packages/serializers/`, never in API modules, and you never return raw DB records.
*Remedy:* extract to `packages/serializers/src/server/collections/mcp-approvals/` following the `{name}.attributes.ts` / `{name}.config.ts` / `{name}.serializer.ts` triplet, then import into the controller.

**3. Catch-all index signature demotes the whole model to `unknown` — `mcp-approval.schema.ts:7`.**
`McpApprovalDocument` adds `[key: string]: unknown`. That index signature collapses every typed Prisma field (`arguments`, `result`, `resolvedAt`, `createdAt`) to `unknown` at every access site, which is exactly why the controller then has to `as`-cast each field back. This is deferred-`any` introduced wholesale by this PR.
*Remedy:* delete the index signature — the Prisma `McpApproval` model already supplies the typed fields. If genuine extra runtime keys exist, name them as explicit optional properties instead of a blanket `unknown` indexer. Fixing this removes the downstream casts for free.

**4. Inline response type inside a controller — `mcp-approvals.controller.ts:24-32`.**
`McpApprovalResponse` is an inline type alias in a controller file; project rules require response/prop shapes in `packages/props/` or a colocated `*.types.ts`. (Folds into finding #2 — moving the serializer takes this with it.)

**5. Single-use placeholder DTO — `dto/update-mcp-approval.dto.ts:1-5`.**
`UpdateMcpApprovalDto` exists only to fill the generic slot in `BaseService<…, CreateMcpApprovalDto, UpdateMcpApprovalDto>`. No controller route, no service method, no handler ever accepts one; `resolve()` builds its update payload inline.
*Remedy:* delete the file and collapse to `BaseService<McpApprovalDocument, CreateMcpApprovalDto>` (the `UpdateDto` generic defaults to `Partial<CreateDto>`, which is sufficient).

**6. Three passthrough HTTP wrappers — `streamable-http.service.ts:55-65`.**
`handlePost`, `handleGet`, and `handleDelete` are each a single `await this.processRequest(req, res)` with no verb-specific logic — and the class JSDoc itself states the SDK handles all verbs identically in stateless mode. The three stubs add indirection with zero behavioral distinction.
*Remedy:* make `processRequest` the public entry point (e.g. rename to `handleRequest`) and call it directly from `main.ts` for all three routes; the three stubs vanish.

**7. Single-call `textResult()` helper — `tool-registry.service.ts:227-246`.**
`textResult()` wraps a one-line literal (`{ content: [{ text, type: 'text' }] }`) and is called from exactly one place (the `decline` branch). It unifies nothing — `pendingApprovalResult` builds the identical envelope inline without delegating to it.
*Remedy:* delete `textResult()` and inline the literal at its single call site.

**8. Three-state ternary decode where a guard suffices — `tool-registry.service.ts:163-174`.**
`args.decision` is decoded into `'approve' | 'decline' | null`, then validated via `!decision`. The schema already constrains `decision` to `'approve' | 'decline'`; the `null` intermediate is an extra observable state the reader must track that behaves identically to any other invalid input.
*Remedy:* `const decision = args.decision as 'approve' | 'decline'; if (decision !== 'approve' && decision !== 'decline') throw …` — collapses decode and validate into one pass with no intermediate state.

### Minor

**9. Inert `@Optional()` decorator — `tool-registry.service.ts:70`.**
`@Optional()` on the `requestRole: McpRole` constructor param is a no-op: NestJS DI cannot inject `McpRole` (a TS string-union alias, not a class/token), so the TypeScript default `= 'user'` is what actually provides the fallback in both the manual-`new` and DI-singleton paths. The decorator only misleads a reader into thinking DI could override the role.
*Remedy:* drop the `@Optional()` import and decorator; keep the default parameter.

**10. Structurally-wrong cast without comment — `tool-registry.service.ts:199`.**
`result as Record<string, unknown>` casts `executeTool`'s `{ content, isError?, component? }` return into a plain string-keyed record with no comment explaining the deliberate mismatch.
*Remedy:* add a one-line comment that the cast intentionally stores the opaque MCP result blob, or widen `resolveApproval`'s parameter to a named result type.

**11. Double-cast through `unknown` without comment — `auth-whoami.controller.ts:96`.**
`member?.role as unknown as { key?: string } | undefined` has no comment explaining why the populated relation type can't be used directly.
*Remedy:* add a one-line comment, or declare a local populated-member type.

**12. `client.service.ts` now 1,770 lines — `apps/server/mcp/src/services/client.service.ts`.**
Over the 1,000-line threshold, but this PR only added ~70 lines (the new approval methods + a baseURL change) to a file that was already ~1,701 lines on master. Flagged for awareness, not as a blocker on this PR — the PR inherited the monolith rather than creating it.
*Remedy (opportunistic):* extract the new `createApproval`/`getApproval`/`resolveApproval` methods (lines ~167-229) into a dedicated `ApprovalClientService` wrapping the shared Axios client. That keeps the approval HTTP surface in one place and stops the file growing further.

### Cross-PR coordination note

This PR adds 24 lines to `packages/tools/src/registry/source.ts`. PR #695 **deletes** that file and splits it into per-module files. Whichever merges first forces a rebase of the other — coordinate so these new tool registrations land in #695's new structure rather than the deleted monolith.

### Approved axes (clean)

Spaghetti branching, sequential orchestration, stack hygiene (no npm/npx/tailwind-config/middleware regressions) — no findings.

---

**Verdict: BLOCK.**

Single hard blocker: the non-atomic `resolve()` (finding #1). It is a real concurrency race on the approval-queue security boundary and must become an atomic write before merge. Findings #2-#3 (serializer-in-controller, the `[key: string]: unknown` indexer) are strong request-changes that are cheap to fix and directly violate documented repo invariants — fold them in while you're in there, since #3's fix also removes the downstream casts. The remaining items are low-cost cleanups.

PR #693, opened 2026-06-19
feat(app): library mood board canvas

VincentShipsIt:
## What

A fullscreen, pannable/zoomable **mood board** for the library (Google Stitch-style): one big canvas that pulls **all of a brand's generated visual assets** (images, videos, gifs, avatars) together, with the arrangement **persisted per brand**.

## Why

Vincent wanted a single visual surface to see everything generated for a brand at a glance and arrange it freely — distinct from the per-type library list views.

## Approach

- **Canvas** — reuses the workflow editor's canvas tech (`@xyflow/react`) via a *thin, dedicated* canvas rather than the execution-coupled `WorkflowCanvas`/`useWorkflowStore`. A custom `mediaAsset` node bypasses `BaseNode` (same pattern as `BrandAssetNode`); clicking a tile opens the existing `MediaLightbox`.
- **Persistence** — new brand-scoped `MoodBoard` model (one per brand) stores **only `{ assetId, position }` layout**, never asset URLs. Content is fetched live from ingredients and merged by id, so the board self-heals as assets are generated/deleted. Drag autosaves on a 2s debounce (mirrors the workflow canvas).
- **Data** — no combined-all endpoint exists, so assets are loaded by fanning out parallel per-type ingredient fetches (images/videos/gifs/avatars), filtered to visual media.
- **Shell** — fullscreen route strips the app sidebar/topbar via a new `isMoodboardRoute` branch, gated by the `moodboard` feature flag.
- **Entry points** — "Mood board" button on the library toolbar + ⌘K command (*Go to Mood Board*).

## Layers touched

Prisma model + migration · serializer triplet · interfaces · client/domain models · NestJS `mood-boards` collection (find-or-create GET, PATCH layout) · client service + `useMoodBoard` · asset-fetch hook + layout util · canvas/node/autosave · route + shell strip + entry points.

## Verification

- **type-check:** 45/45 packages clean
- **tests:** 30 passing — utils 6, hooks 3, client 3, models 3, app 8, api 7
- **biome / secretlint / no-raw-html:** clean

### Notes for reviewers
- Migration `20260619140000_add_mood_board_model` is hand-authored (no DB in the worktree); CI/`migrate deploy` applies it.
- Requires the `moodboard` feature flag enabled (or `NEXT_PUBLIC_FEATURE_FLAG_DEFAULTS` unset → passthrough in dev).

### DIFF
--- a/apps/app/app/(protected)/[orgSlug]/[brandSlug]/library/moodboard/layout.tsx
+++ b/apps/app/app/(protected)/[orgSlug]/[brandSlug]/library/moodboard/layout.tsx
@@ -1,0 +1,8 @@
+'use client';
+
+import type { LayoutProps } from '@props/layout/layout.props';
+import FeatureGate from '@ui/guards/feature/FeatureGate';
+
+export default function MoodboardLayout({ children }: LayoutProps) {
+  return <FeatureGate flagKey="moodboard">{children}</FeatureGate>;
+}
--- a/apps/app/app/(protected)/[orgSlug]/[brandSlug]/library/moodboard/page.tsx
+++ b/apps/app/app/(protected)/[orgSlug]/[brandSlug]/library/moodboard/page.tsx
@@ -1,0 +1,21 @@
+import { createPageMetadata } from '@helpers/media/metadata/page-metadata.helper';
+import { Suspense } from 'react';
+import MoodBoardCanvasClient from '@/features/moodboard/MoodBoardCanvasClient';
+
+export const generateMetadata = createPageMetadata('Mood Board');
+
+function MoodBoardPageFallback() {
+  return (
+    <div className="flex h-screen w-full items-center justify-center bg-background text-sm text-foreground/60">
+      Loading mood board…
+    </div>
+  );
+}
+
+export default function LibraryMoodboardPage() {
+  return (
+    <Suspense fallback={<MoodBoardPageFallback />}>
+      <MoodBoardCanvasClient />
+    </Suspense>
+  );
+}
--- a/apps/app/packages/components/app-protected-layout.tsx
+++ b/apps/app/packages/components/app-protected-layout.tsx
@@ -111,6 +111,7 @@ function AppLayoutWithDynamicMenu({
     isFocusedOnboardingRoute,
     isLibraryLandingRoute,
     isLibraryRoute,
+    isMoodboardRoute,
     isOrgRoute,
     isPromptBarRoute,
     isSettingsRoute,
@@ -158,7 +159,7 @@ function AppLayoutWithDynamicMenu({
   );
 
   const menuComponent = useMemo(() => {
-    if (isFocusedOnboardingRoute || isEditorCanvasRoute) {
+    if (isFocusedOnboardingRoute || isEditorCanvasRoute || isMoodboardRoute) {
       return undefined;
     }
 
@@ -208,6 +209,7 @@ function AppLayoutWithDynamicMenu({
     isEditorRoute,
     isFocusedOnboardingRoute,
     isLibraryRoute,
+    isMoodboardRoute,
     isOrgRoute,
     isSettingsRoute,
     isStudioRoute,
@@ -225,7 +227,7 @@ function AppLayoutWithDynamicMenu({
   ]);
 
   const topbarComponent =
-    isEditorCanvasRoute || isFocusedOnboardingRoute
+    isEditorCanvasRoute || isFocusedOnboardingRoute || isMoodboardRoute
       ? undefined
       : isAdminRoute
         ? AdminTopbar
@@ -255,7 +257,7 @@ function AppLayoutWithDynamicMenu({
     </>
   );
 
-  if (isEditorCanvasRoute) {
+  if (isEditorCanvasRoute || isMoodboardRoute) {
     return (
       <CommandPaletteProvider>
         <CommandPaletteInitializer />
--- a/apps/app/packages/components/useAppProtectedLayout.ts
+++ b/apps/app/packages/components/useAppProtectedLayout.ts
@@ -142,6 +142,7 @@ export function useAppProtectedLayout(
   const hasSecondaryTopbar =
     !isAdminRoute && pathname.startsWith(APP_ROUTE_PREFIXES.STUDIO);
   const isEditorCanvasRoute = isProtectedEditorCanvasRoute(pathname);
+  const isMoodboardRoute = pathname === APP_ROUTES.LIBRARY.MOODBOARD;
   const isWorkflowsRoute =
     pathname.startsWith(APP_ROUTE_PREFIXES.WORKFLOWS) ||
     pathname.startsWith(APP_ROUTE_PREFIXES.ORCHESTRATION);
@@ -167,7 +168,10 @@ export function useAppProtectedLayout(
                   : 'workspace';
 
   const shouldMountAgentPanel =
-    isTerminalDockAvailable() && !isEditorCanvasRoute && !isConversationRoute;
+    isTerminalDockAvailable() &&
+    !isEditorCanvasRoute &&
+    !isMoodboardRoute &&
+    !isConversationRoute;
   const shouldInitAgentApiService =
     shouldMountAgentPanel || isConversationRoute;
 
@@ -490,6 +494,7 @@ export function useAppProtectedLayout(
     isFocusedOnboardingRoute,
     isLibraryLandingRoute,
     isLibraryRoute,
+    isMoodboardRoute,
     isOrgRoute,
     isPromptBarRoute,
     isSettingsRoute,
--- a/apps/app/src/features/moodboard/MediaAssetNode.test.tsx
+++ b/apps/app/src/features/moodboard/MediaAssetNode.test.tsx
@@ -8,10 +8,12 @@ import { MediaAssetNode } from '@/features/moodboard/MediaAssetNode';
 import type { MediaAssetNodeProps } from '@/features/moodboard/moodboard.types';
 
 vi.mock('next/image', () => ({
-  default: (props: React.ImgHTMLAttributes<HTMLImageElement>) => (
-    // biome-ignore lint/a11y/useAltText: alt is forwarded via props
+  default: ({
+    alt = '',
+    ...props
+  }: React.ImgHTMLAttributes<HTMLImageElement>) => (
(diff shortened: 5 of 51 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**Verdict: 🚫 BLOCK**

The PR introduces a solid mood board canvas feature but ships two hard repo-standards violations that must be fixed before merge: an exported interface declared inline in a component file (instead of the colocated `moodboard.types.ts` that already exists for exactly this purpose), and a relative sibling import in a DTO that must use the `@api` path alias like every peer DTO. Beyond those blockers it accumulates meaningful simplification debt — three more inline interface declarations, a `ReactFlowProvider` wrapper placed at the wrong abstraction level versus the workflow-canvas peers, a dead entity class, and a `NEGATIVE_INFINITY` sentinel encoding a two-state machine as one number. None of the simplifications block individually, but they should land to keep the new surface maintainable.

No file crosses the 1,000-line limit. Findings below survived adversarial verification (18 of 19 raw findings confirmed).

### 🚫 Blockers

1. **`apps/app/src/features/moodboard/MoodBoardCanvas.tsx:25-32` — exported `MoodBoardCanvasProps` declared inline.** A colocated `moodboard.types.ts` already exists and is the canonical home. **Fix:** move the interface there and `import type { MoodBoardCanvasProps } from '@/features/moodboard/moodboard.types';`.
2. **`apps/server/api/src/collections/mood-boards/dto/update-mood-board.dto.ts:2` — relative sibling import.** Every peer update DTO uses the `@api` alias; path aliases over relative imports is an explicit repo rule. **Fix:** `import { CreateMoodBoardDto } from '@api/collections/mood-boards/dto/create-mood-board.dto';`.

### Structural regressions

- **`MoodBoardCanvas.tsx:25-32`** — see blocker 1.
- **`update-mood-board.dto.ts:2`** — see blocker 2.

- **`use-mood-board-canvas.ts:15-27`** — `UseMoodBoardCanvasParams`/`UseMoodBoardCanvasResult` declared inline in the hook source. Move to `moodboard.types.ts` and import.
- **`MoodBoardCanvasClient.tsx:13-19`** — `CanvasMessage` uses an anonymous object-literal param type. Promote to `CanvasMessageProps` in `moodboard.types.ts`.
- **`use-brand-media-assets.ts:25-36`** — the public `UseBrandMediaAssetsResult` contract is declared inline; extract to a colocated `use-brand-media-assets.types.ts` and re-export from `index.ts`. The file-private `PageableService` (single internal cast) can stay inline or become a `type` alias — extracting a one-use private alias would be over-engineering.
- **`use-brand-media-assets.ts:52-80`** — `fetchAllPages` and `isVisualMediaIngredient` are pure functions with no React dependency embedded in a 194-line hook. If reuse is likely (e.g. a future `use-brand-video-assets`), extract to `packages/utils/moodboard/mood-board-assets.util.ts` alongside the existing layout util; otherwise leaving them colocated is defensible.
- **`MoodBoardCanvas.tsx:130-138`** — the outer `MoodBoardCanvas` exists solely to wrap `MoodBoardCanvasInner` in `<ReactFlowProvider>` and spread props through, with zero logic. The workflow-canvas peers (`WorkflowDetailPageClient`, `WorkflowNewPageClient`) put the provider at the page-level client component. Collapse to one `MoodBoardCanvas` and move `<ReactFlowProvider>` into `MoodBoardCanvasClient.tsx`.
- **`packages/models/content/mood-board.model.ts:1-8`** — the subclass adds a constructor that only calls `super(partial)` with no hydration. Peer models with no relationship deserialization are empty class bodies. Reduce to `export class MoodBoard extends BaseMoodBoard {}`.
- **`apps/app/packages/components/app-protected-layout.tsx:162-260`** — `isEditorCanvasRoute || isMoodboardRoute` appears three times independently (memo, ternary, early-return); the next canvas route needs three edits. Hoist `const isFullscreenCanvasRoute = isEditorCanvasRoute || isMoodboardRoute;` and reuse.
- **`MoodBoardCanvas.tsx:118-125`** — `MediaLightbox` is guarded by `lightboxIndex !== null` yet also passed `open={lightboxIndex !== null}`, which is tautologically `true` inside the guard. Either use the boolean shorthand `open` to make the unmount-on-close intent explicit, or drop the outer guard and let the lightbox manage open/close.
- **`mood-board-layout.util.ts:76-86`** — `Number.NEGATIVE_INFINITY` is used as a "no live saved items yet" sentinel, then disambiguated with `Number.isFinite`, encoding a two-state machine in one number. Replace with explicit `hasSavedLiveItem: boolean` + `maxSavedY: number`.
- **`MoodBoardCanvas.tsx:45-51`** — imperative `assets.forEach((a, i) => map.set(a.id, i))` is the long form of `new Map(assets.map((a, i) => [a.id, i]))`. Collapse.
- **`mood-board-layout.util.ts:109-114`** — conditional spreads `...(seed.width !== undefined ? { width } : {})` for optional fields are noise; the target fields are already optional, so assigning `width: seed.width` directly is equivalent. Inline the four optional fields.
- **`apps/server/api/src/collections/mood-boards/entities/mood-board.entity.ts:1-18`** — `MoodBoardEntity` has zero importers; the service uses the Prisma-derived `MoodBoardDocument` directly. Dead code is a defect under the green-field policy. Delete the file and its `entities/` directory.

---
*Automated thermo-nuclear review — multi-agent finders across 5 quality dimensions, each finding adversarially verified before inclusion (19 raw → 18 confirmed). Read-only: no tests, builds, or typechecks were run.*

VincentShipsIt:
## 🔥 Thermo-Nuclear Review

Structural / maintainability pass on the library mood board canvas. Methodology: seven parallel structural finders (file size, abstraction, branching, duplication, layer/type discipline, incidental complexity, async/atomicity) followed by adversarial verification of every finding against the checked-out branch (head `5d008c90f`, which already resolved the earlier review blockers) — only confirmed findings appear below.

**Headline:** a clean, well-tested feature. The "thin dedicated canvas" claim holds up — it reuses `@xyflow/react` directly and does *not* drag in the execution-coupled `WorkflowCanvas`/`useWorkflowStore`, and `MediaAssetNode` follows the established `BrandAssetNode`-bypasses-`BaseNode` pattern. Largest file is ~186 lines, so no size concerns anywhere. Every finding below is confined to a handful of new files and is cheap to fix; none requires redesign.

None.

### Request Changes (significant)

**1. TOCTOU race in `findOrCreateByBrand` — `mood-boards.service.ts:23-48`.** `findFirst` then `create` is non-atomic. Two concurrent first-loads of the same brand's board (a double-mount, a rapid re-navigation, two tabs) race, and the second `create` throws on the unique constraint — a 500 on an ordinary read path. This is the one I would most want fixed before merge.
*Remedy:* `upsert` on the brand-unique key, or catch the unique-violation and re-read.

**2. Inline interface violates the canonical-layer rule — `use-brand-media-assets.ts:26-28`.** `PageableService` is an inline interface duplicating the canonical `FindAllService<T>` already exported from `@genfeedai/interfaces`. Repo rule: no inline interfaces outside `packages/interfaces`. Import the existing one and delete the local copy.

**3. Duplicated service wiring — `use-brand-media-assets.ts:88-100`.** Three of the four `useAuthedService` getters (Images, Videos, GIFs) reimplement service access that `useIngredientServices()` already provides inside the same `@genfeedai/hooks` package. Reuse the existing hook instead of hand-wiring the services.

**4. Media-type union re-expanded — `use-brand-media-assets.ts:30-42`.** `isVisualMediaIngredient` hand-expands the IMAGE/VIDEO/GIF/avatar union that `isImageIngredient` + `isVideoIngredient` (from `ingredient-type.util`) already encapsulate. A new visual category would then have to be added in two places.
*Remedy:* collapse to `(isImageIngredient(i) || isVideoIngredient(i)) && Boolean(i.ingredientUrl || i.thumbnailUrl)`.

**5. Dead `MoodBoardEntity` — `entities/mood-board.entity.ts`.** A scaffold-generated class that is never imported anywhere; the schema already re-exports the Prisma type directly. Delete the file (CLAUDE.md: delete dead code aggressively).

**6. `knownIds` is not ref-stable — `use-mood-board-canvas.ts:41-44, 65-72`.** `knownIds` sits in `onNodeDragStop`'s dependency array even though it is only read inside the debounce timeout closure. Every `assets` refresh recreates the callback and can stale an in-flight timer. The same file already solves this exact problem for `onPersist` via `onPersistRef`.
*Remedy:* mirror that pattern with a `knownIdsRef`; the callback's deps then collapse to `[]`.

**7. Save callback has no in-flight guard — `use-mood-board.ts:37-51`.** Concurrent patch calls can land out of order, and `invalidateQueries` after *every* save forces a full refetch and node reset mid-interaction. Guard concurrent saves and avoid the blanket invalidation on the optimistic path.

### Minor

- **`MoodBoardsModule` imports nothing** (`mood-boards.module.ts:5-10`) though its controller and service both inject `LoggerService`. It works because `LoggerModule` is `@Global()`, but the rest of the codebase lists `LoggerModule` explicitly for a legible dependency graph — match that.
- **`error` from `useBrandMediaAssets` is unhandled** (`MoodBoardCanvasClient.tsx:30-35`). A failed fetch leaves `isLoading=false` and `assets=[]`, so the canvas shows "No assets to arrange yet" instead of an error state — misleading for a brand that actually has assets.
- **Dead `layout` computation** (`mood-board-layout.util.ts:118-123`). The `layout` field of `MergeMoodBoardLayoutResult` is built on every call but consumed only by tests — the production caller destructures only `seeds`. Drop it (or stop computing it on the hot path).
- **Heavier conditional-spread form** (`mood-board-layout.util.ts:119-120`) uses `...(x !== undefined ? { k: x } : {})` where the short-circuit `...(x !== undefined && { k: x })` is the lighter idiom. Minor, and the only such usage in the package.
- **Unreachable fallback** in `handleClose` (`MoodBoardCanvasClient.tsx:55-62`): both slugs are always present on this route, so the `router.back()` branch is dead.

### Approved axes (clean)

Canvas thinness (genuinely decoupled from the workflow store), serializer triplet, file sizes, Prisma schema/migration alignment, test coverage — no findings.

---

**Verdict: APPROVE.** No blocker gates merge. Land the `findOrCreateByBrand` upsert (#1) and the inline-interface/duplication cleanups (#2–#4) before or immediately after merge — none require redesign, and the head commit already cleared the earlier review round.

PR #694, opened 2026-06-19
feat(pages): Fastlane studio page package

VincentShipsIt:
## Summary

- Adds `packages/pages/studio/fastlane/` — the complete frontend package for the Fastlane studio mode (brand data → ideas → generate → blitz review → schedule)
- Wires to the existing route wrapper at `apps/app/.../studio/fastlane/page.tsx` which imports `@pages/studio/fastlane` (default export)
- Adds `@testing-library/react` alias to `packages/pages/vitest.config.ts` so hook tests can use `renderHook`

## Architecture

3-step wizard in `FastlaneLayout`:
1. **Ideas** (`FastlaneIdeaSelector`) — format toggles (image/video/avatar), count stepper (3–9), optional angle, credit estimate, idea preview cards
2. **Review** (`FastlaneBlitz`) — Tinder-style CSS-transform swipe (translate+rotate+opacity, no GSAP). ←/→ arrow key support. Failed asset count badge.
3. **Schedule** (`FastlaneSchedulePanel`) — per-asset editable caption, per-credential datetime picker, "post now" or scheduled.

## Hooks

- `useFastlaneIdeas` — wraps `BrandsService.generateFastlaneIdeas`; AbortController; 400 errors surface back to step 1
- `useFastlaneGeneration` — fan-out to `ImagesService`/`VideosService`/`HeyGenService`; WS subscription per pending ID via `useSocketManager`; `Promise.allSettled` so per-asset failures don't abort the batch; full cleanup on unmount
- `useFastlaneSchedule` — one `PostsService.post` per approved asset × credential; shared `groupId`; `source: 'fastlane'`; partial failure surface

## Guards

- `useFastlaneEnabled` — org-level flag; renders "not available" empty state before any content loads
- `FastlaneBrandGate` — checks voice tone, ≥1 reference image, ≥1 short-form social; avatar format requires `defaultAvatarIngredientId` + `defaultVoiceId`

## Test plan

- [ ] `bun run test --filter=@genfeedai/pages` → 33 tests pass (6 suites)
- [ ] `bunx turbo run type-check --filter=@genfeedai/app` → 17 tasks successful
- [ ] `bunx biome check --write packages/pages/studio/fastlane` → 10 files fixed, 0 errors
- [ ] Pre-commit hooks (lint-staged, secretlint, lint-no-raw-html) all pass

### DIFF
--- a/apps/app/app/(protected)/[orgSlug]/[brandSlug]/studio/fastlane/page.tsx
+++ b/apps/app/app/(protected)/[orgSlug]/[brandSlug]/studio/fastlane/page.tsx
@@ -1,0 +1,14 @@
+import { createPageMetadata } from '@helpers/media/metadata/page-metadata.helper';
+import FastlanePageContent from '@pages/studio/fastlane';
+import LazyLoadingFallback from '@ui/loading/fallback/LazyLoadingFallback';
+import { Suspense } from 'react';
+
+export const generateMetadata = createPageMetadata('Fastlane');
+
+export default function StudioFastlanePage() {
+  return (
+    <Suspense fallback={<LazyLoadingFallback variant="grid" />}>
+      <FastlanePageContent />
+    </Suspense>
+  );
+}
--- a/apps/app/app/(protected)/admin/organization/components/organization-settings-table.tsx
+++ b/apps/app/app/(protected)/admin/organization/components/organization-settings-table.tsx
@@ -64,6 +64,11 @@ const SETTING_GROUPS: SettingGroup[] = [
         label: 'Auto Evaluation',
         type: 'boolean',
       },
+      {
+        key: 'isFastlaneEnabled',
+        label: 'Fastlane',
+        type: 'boolean',
+      },
     ],
   },
   {
--- a/apps/app/packages/components/useAppProtectedLayout.ts
+++ b/apps/app/packages/components/useAppProtectedLayout.ts
@@ -30,6 +30,7 @@ import {
   STUDIO_CATEGORY_CONFIG,
   useEnabledCategories,
 } from '@hooks/data/organization/use-enabled-categories/use-enabled-categories';
+import { useFastlaneEnabled } from '@hooks/data/organization/use-fastlane-enabled/use-fastlane-enabled';
 import { useFeatureFlag } from '@hooks/feature-flags/use-feature-flag';
 import { useOrgUrl } from '@hooks/navigation/use-org-url';
 import { useMenuItems } from '@hooks/ui/use-menu-items';
@@ -321,6 +322,8 @@ export function useAppProtectedLayout(
   const role = useUserRole();
   const { enabledCategories, isLoading: isEnabledCategoriesLoading } =
     useEnabledCategories();
+  const { isEnabled: isFastlaneEnabled, isLoading: isFastlaneLoading } =
+    useFastlaneEnabled();
 
   // Sync route context into the agent store
   useAgentPageContext(role);
@@ -376,6 +379,15 @@ export function useAppProtectedLayout(
       const studioCategory = categoryByHref.get(item.href);
 
       if (!studioCategory) {
+        // Fastlane is not a generation category — gate it on its own org flag.
+        // Hide while the flag is still loading to avoid a flash of the item.
+        if (
+          item.href === APP_ROUTES.STUDIO.FASTLANE &&
+          (!isFastlaneEnabled || isFastlaneLoading)
+        ) {
+          return items;
+        }
+
         items.push({
           ...item,
           href: withTaskContextHref(item.href, taskContextSearchParams),
@@ -395,7 +407,13 @@ export function useAppProtectedLayout(
 
       return items;
     }, []);
-  }, [enabledCategories, isEnabledCategoriesLoading, taskContextSearchParams]);
+  }, [
+    enabledCategories,
+    isEnabledCategoriesLoading,
+    isFastlaneEnabled,
+    isFastlaneLoading,
+    taskContextSearchParams,
+  ]);
 
   const composeMenuItems = useMemo(
     () =>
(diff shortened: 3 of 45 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**Verdict: 🚫 BLOCK**

The Fastlane page package ships a solid wizard skeleton with good test coverage and correct `AbortController` discipline throughout, but three structural issues must be resolved before merge. The split-brain `approvalMap` pattern in `FastlaneLayout` creates permanently divergent state between the hook's internal asset list and the layout's enriched view — `updateAsset` already exists inside `useFastlaneGeneration` but is deliberately withheld from the return type, forcing a parallel state machine that any code reading `assets` directly will observe as stale. The inline `CredentialTarget` interface violates the repo's no-inline-interfaces rule, and two raw `` tags suppress the project-wide `next/image` lint rule rather than conforming to it.

No file crosses the 1,000-line limit. Findings below survived adversarial verification (25 of 28 raw findings confirmed).

### 🚫 Blockers

1. **`FastlaneLayout.tsx:82-119` — split-brain `approvalMap` state.** `useFastlaneGeneration` already defines `updateAsset` internally (line 132) but omits it from `UseFastlaneGenerationReturn` (line 348), forcing the layout to maintain a parallel `approvalMap`. `useFastlaneSchedule` re-filters against the merged `enrichedAssets` view, so any code reading `assets` directly observes stale approval statuses — a leaky abstraction with permanent correctness risk. **Fix:** add `updateAsset` to the hook return type and statement, destructure it in the layout, drive `handleApprove`/`handleReject` through `updateAsset(ideaId, { status })`, then delete the `approvalMap` `useState` and `enrichedAssets` `useMemo` and use `assets` directly.
2. **`FastlaneSchedulePanel.tsx:26-32` — inline `CredentialTarget` interface.** Non-prop interface declared inside a component file; `typescript-strict.md` prohibits this. The module already has a `types.ts`. **Fix:** move `CredentialTarget` to `packages/pages/studio/fastlane/types.ts` and `import type { CredentialTarget } from '../types';`.
3. **`FastlaneBlitz.tsx:136` and `FastlaneSchedulePanel.tsx:191` — raw `` with `eslint-disable @next/next/no-img-element`.** Two suppressions in one PR normalise an escape hatch the codebase enforces against project-wide (see `packages/pages/studio/generate/hooks/useTableColumns.tsx:30`). **Fix:** use `Image` from `next/image` in both spots and drop the disable comments.

### Structural regressions

- **`FastlaneLayout.tsx:82-119`** — see blocker 1.
- **`FastlaneSchedulePanel.tsx:26-32`** — see blocker 2.
- **`FastlaneBlitz.tsx:136`** — raw ``; replace with `<Image fill className="object-cover" />` inside a `relative` wrapper.
- **`FastlaneSchedulePanel.tsx:191`** — same raw `` pattern; replace with `<Image width={48} height={48} className="rounded-md object-cover shrink-0" />`.

- **`FastlaneIdeaSelector.tsx:45-48`** — `selectedFormats` is owned by both the selector (internal `useState`) and the layout (external `useState`); every `onGenerate` overwrites the layout copy, and the two diverge between toggles. Make the selector controlled: lift state to the layout, pass `selectedFormats` + `onFormatsChange` props, remove the reconciliation write at layout line 71.
- **`FastlaneBlitz.tsx:51-59`** — `triggerSwipe` closes over `onApprove`/`onReject` but the keydown `useEffect` suppresses exhaustive-deps, capturing stale callbacks if identities change. Wrap `triggerSwipe` in `useCallback([currentAsset, swipeDir, onApprove, onReject])`, depend on it, drop the disable.
- **`types.ts:13-16, 53-55`** — `UseFastlaneIdeasParams` and `UseFastlaneScheduleParams` are dead interfaces; the hooks take a positional string / no params. Delete both.
- **`useFastlaneGeneration.ts:163-186`** — three `if/else-if` branches in `subscribeToAsset` differ only by service getter; the two extract lines are copied verbatim. Collapse to a ternary that picks the service, then extract once.
- **`useFastlaneGeneration.ts:246-274`** — `image` and `video` branches in `dispatchIdea` share structure (base payload → builder → blacklist spread → `service.post`). Extract a `dispatchMediaPost<T>(getService, buildPayload, promptData, brandId)` helper; keep the `avatar` branch separate (it uses `service.generate()`).
- **`useFastlaneGeneration.ts:35`** — `FASTLANE_CREDIT_COSTS` typed `Record<string, number>` loses exhaustiveness. Tighten to `Record<FastlaneFormat, number>`.
- **`useFastlaneGeneration.ts:10`** — `FASTLANE_CREDIT_COSTS` is plain data imported by a component, inverting dependency direction and pulling the hook's import graph into the component bundle. Move it to `constants.ts`.
- **`FastlaneBlitz.tsx:100-116`** — the swipe `transition` string repeats in every ternary branch and the idle branch sets redundant CSS defaults. Hoist `SWIPE_TRANSITION` and collapse to `swipeDir ? {...} : {}`.
- **`useFastlaneGeneration.ts:50-54`** — outer `.trim()` on line 52 is a no-op (every branch already returns trimmed/empty); lines 82-83 evaluate `(idea.speechText ?? idea.caption ?? '').trim()` twice. Remove the outer trim and hoist the shared expression.
- **`FastlaneLayout.tsx:107`** — `approvalMap` `useState` is declared after the `handleApprove`/`handleReject` that reference its setter, breaking the declaration-before-use ordering the rest of the component follows. (Moot if blocker 1 removes it.)
- **`FastlaneLayout.tsx:121-126`** — `handleSchedule` is a pure-indirection `useCallback` that only `void`s `scheduleApproved`. Inline as `onSchedule={(params) => void scheduleApproved(params)}`.

---
*Automated thermo-nuclear review — multi-agent finders across 5 quality dimensions, each finding adversarially verified before inclusion (28 raw → 25 confirmed). Read-only: no tests, builds, or typechecks were run.*

PR #696, opened 2026-06-20
refactor(api): decompose DarkroomService god object into facade + peer services (#690)

### LINKED ISSUE #690: Decompose the DarkroomService god object in apps/server/api/src/endpoints/admin/darkroom/darkroom.service.ts (2,107 lines)
## Summary

`apps/server/api/src/endpoints/admin/darkroom/darkroom.service.ts` is a single 2,107-line `@Injectable()` class with **17 injected constructor dependencies** (lines 93-111) conflating at least seven domains: type-coercion helpers (122-224), character/persona CRUD (226-295), asset review/listing (297-398), training lifecycle (400-565), AWS infrastructure — EC2/CloudFront, owning its own `EC2Client` + `CloudFrontClient` (567-818), social publishing (820-959), image generation + job orchestration (961-1451), dataset ingest/scraping (1100-2055), stats/campaigns (1453-1635), and lip-sync + voice/TTS (1637-1833, 2057-2106). The 17-dependency constructor is the clearest symptom of a god object. The sibling `services/darkroom-training.service.ts` (405 lines) already demonstrates the intended peer-service pattern for this folder, so this decomposition follows existing house style rather than introducing a new one.

## Confirmed findings (verified against source)

- **Oversized file**: 2,107 lines, ~51 methods, 17 deps, 7+ domains.
- **Dead method**: `getqueryStats` (lines 1623-1635) is a pure pass-through to `getPipelineStats` with zero callers anywhere (the controller and spec call `getPipelineStats` directly). Delete outright per the repo's green-field "delete dead code aggressively" rule.
- **Unused parameter**: `_brandId: string` in `startTraining` (line 449) is never referenced in the body (468-564); remove it and update the single controller caller.
- **Oversized methods (>80 lines)**: `startTraining` 446-565 (~120); `generateImage` 967-1098 (~132, with 9 sequential `if` param assignments at 1011-1034); `publishAsset` 825-959 (~135, per-platform `switch` with full upload flow inlined per case); `ingestTrainingDataForPersonaDocument` 1835-1929 (~95, nested triple loop); `listCampaigns` 1458-1539 (~82).
- **Duplicated persona lookup**: the identical "find darkroom character by slug or throw `NotFoundException`" block (`{ isDarkroomCharacter: true, isDeleted: false, organization, slug }`) is copy-pasted at lines 248-257, 468-477, 994-1003, 1106-1115, 1174-1183, 1251-1260, 1660-1669, with a near-identical no-throw variant at 2065-2070 — eight sites that must all change together on any schema change.
- **Duplicated AWS try/catch boilerplate**: `getEC2Status` (576-611), `ec2Action` (624-659), and `invalidateCloudFront` (729-763) each wrap their SDK call in the same log-error-and-rethrow shape.
- **Nested-ternary status derivation**: `mapIngredientToGenerationJob` (1422-1433) is a 5-level nested ternary; `updateGenerationJob` (1321-1333) repeats the pattern.
- **Direct cross-collection Prisma access**: `listCampaigns` (1471) and `getPipelineStats` (1575-1591) reach into `this.ingredientsService.prisma.ingredient` / `this.trainingsService.prisma.training` directly, bypassing those collections' query surfaces and coupling to their ORM internals.

## Restructuring plan (behavior-preserving)

Split into a thin facade plus focused collaborators under `endpoints/admin/darkroom/services/` (matching the existing `darkroom-training.service.ts`), target no file over ~350 lines:

- `DarkroomCharacterService` (226-295), `DarkroomAssetService` (297-398 + `syncApprovedAssetToDataset`), move `startTraining` into the existing `DarkroomTrainingService`.
- `DarkroomInfraService` (567-818) — **owns `EC2Client`/`CloudFrontClient`**, removing the AWS clients from the main constructor, with a single `runAws(label, fn)` wrapper replacing the duplicated try/catch.
- `DarkroomPublishingService` (820-959), `DarkroomGenerationService` (961-1451), `DarkroomIngestService` (1100-2055 ingest/scrape/dataset), `DarkroomStatsService` (1453-1635, minus the dead `getqueryStats`), `DarkroomMediaService` (lip-sync + voice).
- A shared `requirePersonaBySlug(slug, orgId)` helper replaces all eight duplicated persona-lookup blocks; type-coercion helpers (122-224) move to a `DarkroomValueReader` util.
- Replace nested-ternary status derivations with a lookup map/switch.
- Add `groupBy`/`count` methods to `IngredientsService`/`TrainingsService` and call those instead of reaching into `.prisma.*`.

## Acceptance criteria

- `darkroom.service.ts` reduced to a facade under ~200 lines; no resulting file exceeds ~350 lines.
- `getqueryStats` deleted (repo grep returns zero references); unused `_brandId` removed.
- The eight duplicated persona-lookup blocks replaced by one `requirePersonaBySlug` helper with identical `NotFoundException` behavior.
- `EC2Client`/`CloudFrontClient` instantiation moved out of the main constructor into `DarkroomInfraService`.
- Nested-ternary status derivations replaced by a lookup map/switch covered by a unit test over every `(status, stage)` input.
- `listCampaigns`/`getPipelineStats` no longer access `*.prisma.*` directly.
- Existing `darkroom.controller.spec.ts` passes unchanged; `bun type-check` and `bunx turbo lint` green.

*Filed by the automated code-quality sweep. Findings verified against current source; line numbers may shift as the file changes.*

VincentShipsIt:
## Why

`apps/server/api/src/endpoints/admin/darkroom/darkroom.service.ts` was a **2,107-line `@Injectable()` god object** with **17 constructor deps** spanning 7+ domains (character CRUD, asset review, AWS infra, social publishing, image generation/jobs, dataset ingest, stats, lip-sync, voice/TTS). This splits it into a **thin delegating facade** plus focused peer services under `services/`, following the existing `darkroom-training.service.ts` precedent. Closes #690.

## Behaviour-preserving

The `DarkroomController` and **`darkroom.controller.spec.ts` are unchanged** — the facade keeps the exact public method surface the controller calls, so the controller spec passes verbatim. Every method body was moved, not rewritten.

## New collaborators (`endpoints/admin/darkroom/services/`)

| Service | Owns |
|---|---|
| `DarkroomCharacterService` | persona CRUD + the single shared `findPersonaBySlug` / `requirePersonaBySlug` |
| `DarkroomAssetService` | asset listing/review + approved-asset dataset sync |
| `DarkroomInfraService` | **owns the EC2 + CloudFront clients** (removed from the main constructor); single `runAws` wrapper |
| `DarkroomPublishingService` | publish approved assets to platforms |
| `DarkroomGenerationService` | ComfyUI generate + durable generation jobs |
| `DarkroomIngestService` | dataset upload + scrape-and-ingest pipeline |
| `DarkroomStatsService` | campaigns + pipeline stats |
| `DarkroomMediaService` | lip-sync + voice/TTS |
| `DarkroomTrainingOrchestratorService` | training reads + `startTraining` |
| `DarkroomValueReader` (util) | shared pure value-coercion + mapping helpers |

Facade is now **264 lines** (from 2,107); no peer service exceeds ~331 lines.

## Acceptance criteria (#690)

- [x] `darkroom.service.ts` reduced to a thin facade; implementation in peer services
- [x] **8 duplicated persona-lookup blocks → one `requirePersonaBySlug`** (per-call-site `NotFoundException` messages preserved exactly, incl. the slug-specific `getCharacterBySlug` message and the no-throw `findPersonaBySlug` used by `resolveFleetReferenceAudio`)
- [x] **`EC2Client`/`CloudFrontClient` moved out of the main constructor** into `DarkroomInfraService`; duplicated AWS try/catch collapsed into `runAws`
- [x] **Dead `getqueryStats` deleted** (`grep` → 0 references)
- [x] Unused `brandId` dropped from the training implementation (see note)
- [x] **Nested-ternary status derivations → explicit switch functions** (`deriveGenerationJobState`, `ingredientStatusForJobState`), covered by a unit test over the full status/stage matrix
- [x] `listCampaigns`/`getPipelineStats` **no longer touch `.prisma.*`** — new aggregation methods (`countPersonaAssets`, `groupPersonaAssetsByStatus`, `groupPersonaAssetsByReviewStatus`, `groupPersonaAssetCampaigns`, `countTrainingsByOrganization`, `groupTrainingsByStage`) added to `IngredientsService`/`TrainingsService`, returning domain-neutral typed shapes (Prisma generics stay internal)

## Notes / deliberate deviations

- **`startTraining` lives in a new `DarkroomTrainingOrchestratorService`, not the existing `DarkroomTrainingService`.** The issue suggested the latter, but it is already 405 lines and moving `startTraining` into it would push it past the ~350-line cap **and** force breaking changes to its passing spec. The new orchestrator keeps the GPU-client service (and its spec) untouched.
- **`_brandId` placeholder on the facade.** The issue asks to remove the unused `brandId`; the controller spec asserts the 4-arg `startTraining('org','user','brand',dto)` call. To satisfy *both* "remove unused brandId" and "controller spec unchanged", the brandId is dropped from the orchestrator implementation while the facade keeps a lint-clean `_brandId` placeholder to preserve the HTTP/controller call shape.
- A biome pre-commit auto-fix added a `node:process` import to `trainings.service.ts` for its pre-existing `process.cwd()` usage (repo-mandated lint rule), riding along with the aggregation-method addition.

## Verification

- ✅ Scoped type-check: `tsc -p apps/server/api/tsconfig.typecheck.json --noEmit` → clean (only the pre-existing `baseUrl` deprecation notice)
- ✅ `biome check` → clean
- ✅ Tests: `darkroom.controller.spec.ts` + `darkroom-training.service.spec.ts` (unchanged) + new `darkroom-value-reader.util.spec.ts` + `darkroom-character.service.spec.ts` → **70 passed**
- Diff scoped to `apps/server/api/src/endpoints/admin/darkroom/**` + the two collection-service aggregation methods. Heavy/integration tests deferred to CI.

### DIFF
--- a/apps/server/api/src/collections/ingredients/services/ingredients.service.ts
+++ b/apps/server/api/src/collections/ingredients/services/ingredients.service.ts
@@ -565,4 +565,83 @@ export class IngredientsService extends BaseService<
 
     return where;
   }
+
+  /**
+   * Count persona-scoped darkroom assets for an organization. Encapsulates the
+   * raw Prisma aggregation so darkroom callers don't reach into `.prisma.*`.
+   */
+  async countPersonaAssets(organizationId: string): Promise<number> {
+    return this.prisma.ingredient.count({
+      where: { isDeleted: false, organizationId, personaId: { not: null } },
+    });
+  }
+
+  /**
+   * Group persona-scoped darkroom assets by ingredient status.
+   */
+  async groupPersonaAssetsByStatus(
+    organizationId: string,
+  ): Promise<Array<{ status: string | null; count: number }>> {
+    const groups = await this.prisma.ingredient.groupBy({
+      _count: { id: true },
+      by: ['status'],
+      where: { isDeleted: false, organizationId, personaId: { not: null } },
+    });
+
+    return groups.map((group) => ({
+      count: group._count.id,
+      status: group.status,
+    }));
+  }
+
+  /**
+   * Group persona-scoped darkroom assets by review status.
+   */
+  async groupPersonaAssetsByReviewStatus(
+    organizationId: string,
+  ): Promise<Array<{ reviewStatus: string | null; count: number }>> {
+    const groups = await this.prisma.ingredient.groupBy({
+      _count: { id: true },
+      by: ['reviewStatus'],
+      where: { isDeleted: false, organizationId, personaId: { not: null } },
+    });
+
+    return groups.map((group) => ({
+      count: group._count.id,
+      reviewStatus: group.reviewStatus,
+    }));
+  }
+
+  /**
+   * Group persona-scoped darkroom assets by campaign + review status, with the
+   * earliest creation timestamp per group.
+   */
+  async groupPersonaAssetCampaigns(organizationId: string): Promise<
+    Array<{
+      campaign: string | null;
+      reviewStatus: string | null;
+      count: number;
+      earliestCreatedAt: Date | null;
+    }>
+  > {
+    const groups = await this.prisma.ingredient.groupBy({
+      _count: { id: true },
+      _min: { createdAt: true },
+      by: ['campaign', 'reviewStatus'],
+      orderBy: { campaign: 'asc' },
+      where: {
+        campaign: { not: null },
+        isDeleted: false,
+        organizationId,
+        personaId: { not: null },
+      },
+    });
+
+    return groups.map((group) => ({
+      campaign: group.campaign,
+      count: group._count.id,
+      earliestCreatedAt: group._min.createdAt ?? null,
+      reviewStatus: group.reviewStatus,
+    }));
+  }
 }
--- a/apps/server/api/src/collections/trainings/services/trainings.service.ts
+++ b/apps/server/api/src/collections/trainings/services/trainings.service.ts
@@ -1,5 +1,6 @@
 import fs from 'node:fs';
 import path from 'node:path';
+import process from 'node:process';
 import { CreateTrainingDto } from '@api/collections/trainings/dto/create-training.dto';
 import { UpdateTrainingDto } from '@api/collections/trainings/dto/update-training.dto';
 import type { TrainingDocument } from '@api/collections/trainings/schemas/training.schema';
@@ -282,5 +283,33 @@ export class TrainingsService extends BaseService<
     return typeof value === 'string' && value.trim().length > 0
       ? value
       : undefined;
+  }
+
+  /**
+   * Count an organization's trainings. Encapsulates the raw Prisma aggregation
+   * so darkroom callers don't reach into `.prisma.*`.
+   */
+  async countTrainingsByOrganization(organizationId: string): Promise<number> {
+    return this.prisma.training.count({
+      where: { isDeleted: false, organizationId },
+    });
+  }
+
+  /**
+   * Group an organization's trainings by stage.
+   */
+  async groupTrainingsByStage(
+    organizationId: string,
+  ): Promise<Array<{ stage: string | null; count: number }>> {
+    const groups = await this.prisma.training.groupBy({
+      _count: { id: true },
+      by: ['stage'],
+      where: { isDeleted: false, organizationId },
+    });
+
+    return groups.map((group) => ({
+      count: group._count.id,
+      stage: group.stage,
+    }));
   }
 }
--- a/apps/server/api/src/endpoints/admin/darkroom/darkroom.module.ts
+++ b/apps/server/api/src/endpoints/admin/darkroom/darkroom.module.ts
@@ -9,7 +9,16 @@ import { PersonasModule } from '@api/collections/personas/personas.module';
 import { TrainingsModule } from '@api/collections/trainings/trainings.module';
 import { DarkroomController } from '@api/endpoints/admin/darkroom/darkroom.controller';
 import { DarkroomService } from '@api/endpoints/admin/darkroom/darkroom.service';
+import { DarkroomAssetService } from '@api/endpoints/admin/darkroom/services/darkroom-asset.service';
+import { DarkroomCharacterService } from '@api/endpoints/admin/darkroom/services/darkroom-character.service';
+import { DarkroomGenerationService } from '@api/endpoints/admin/darkroom/services/darkroom-generation.service';
+import { DarkroomInfraService } from '@api/endpoints/admin/darkroom/services/darkroom-infra.service';
+import { DarkroomIngestService } from '@api/endpoints/admin/darkroom/services/darkroom-ingest.service';
+import { DarkroomMediaService } from '@api/endpoints/admin/darkroom/services/darkroom-media.service';
+import { DarkroomPublishingService } from '@api/endpoints/admin/darkroom/services/darkroom-publishing.service';
+import { DarkroomStatsService } from '@api/endpoints/admin/darkroom/services/darkroom-stats.service';
 import { DarkroomTrainingService } from '@api/endpoints/admin/darkroom/services/darkroom-training.service';
+import { DarkroomTrainingOrchestratorService } from '@api/endpoints/admin/darkroom/services/darkroom-training-orchestrator.service';
 import { IpWhitelistGuard } from '@api/endpoints/admin/guards/ip-whitelist.guard';
 import { FilesClientModule } from '@api/services/files-microservice/client/files-client.module';
 import { ComfyUIModule } from '@api/services/integrations/comfyui/comfyui.module';
@@ -44,6 +53,15 @@ import { forwardRef, Module } from '@nestjs/common';
   ],
   providers: [
     DarkroomService,
+    DarkroomAssetService,
+    DarkroomCharacterService,
+    DarkroomGenerationService,
+    DarkroomInfraService,
+    DarkroomIngestService,
+    DarkroomMediaService,
+    DarkroomPublishingService,
+    DarkroomStatsService,
+    DarkroomTrainingOrchestratorService,
     DarkroomTrainingService,
     IpWhitelistGuard,
     ModelRegistrationService,
--- a/apps/server/api/src/endpoints/admin/darkroom/darkroom.service.ts
+++ b/apps/server/api/src/endpoints/admin/darkroom/darkroom.service.ts
@@ -1,2107 +1,264 @@
-import { BrandsService } from '@api/collections/brands/services/brands.service';
-import { CreatorScraperService } from '@api/collections/content-intelligence/services/creator-scraper.service';
-import { CredentialsService } from '@api/collections/credentials/services/credentials.service';
-import { type IngredientDocument } from '@api/collections/ingredients/schemas/ingredient.schema';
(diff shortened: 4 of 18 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**Scope:** decomposition of the 2,107-line `DarkroomService` god object into a 264-line facade plus 10 peer services (+2,874 / −1,989, 18 files). Structural/maintainability lens only. Every finding below was checked against the actual code by an adversarial verifier; only confirmed defects are listed.

**Headline:** the core refactor lands well. The facade is 264 lines, every peer service is well under the 1,000-line rule (largest touched file is `ingredients.service.ts` at 647), module wiring resolves, and no circular-dependency or behavior-fidelity regression survived verification — the decomposition achieves its stated goal. What it does not yet deliver is *clean* boundaries: the PR declares `DarkroomCharacterService` the owner of persona reads/writes, then routes around it in two places, and it carries several non-atomic write sequences into the new services. None are blocker-severity, but the cluster below is real debt and most of it is cheap to close.

### Requested changes — structural debt the PR introduces or hardens

**1. Owner-service bypass: ingest writes personas directly — `darkroom-ingest.service.ts:257-260` (axis 4, Canonical Layer Discipline)**
`DarkroomIngestService` calls `this.personasService.patch(...)` to write `darkroomSources` back onto the persona, bypassing `DarkroomCharacterService` — the very service this PR introduces to own persona mutations. `characterService` is already injected here, so this is a one-line substitution. Add `updateDarkroomSources(personaId, sources)` to the owner service, call it, and drop the now-redundant `PersonasService` injection.

**2. Owner-service bypass: orchestrator writes lora-state directly — `darkroom-training-orchestrator.service.ts:176-180` (axis 4, Canonical Layer Discipline)**
Same anti-pattern: the orchestrator calls `personasService.patch(...)` to set `loraStatus = TRAINING` instead of going through the declared owner. Combined with the legacy `DarkroomTrainingService.updatePersonaLoraState`, persona lora-state is now written from three places. Migrate the mutation into `DarkroomCharacterService.updateLoraState(...)` and route both call sites through it. This is the finding that most directly contradicts the PR's own thesis — fix it here, or the "owner service" abstraction is decorative.

**3. Non-atomic ingest asset creation — `darkroom-ingest.service.ts:284-313` (axis 6, Non-Atomic Mutations)**
`createDarkroomIngestAsset` does DB create (placeholder `s3Key`) → external S3 upload → DB patch (real `s3Key`), with no transaction or compensation. An S3 failure orphans a row with a stale `cdnUrl`; a patch failure leaves the row permanently pointing at the wrong key. **Restructure:** upload to S3 first (no DB side-effects), then a single `ingredientsService.create` with the resolved `s3Key`/`cdnUrl` — the patch disappears entirely.

**4. Non-atomic and non-retryable dataset sync — `darkroom-asset.service.ts:146-175` (axis 6, Non-Atomic Mutations)**
`syncApprovedAssetToDataset` runs two S3 uploads → `syncDataset` → DB patch with no compensation. Worse, `reviewStatus` is set to `APPROVED` earlier (line 95), so the `!hasReviewStatus(APPROVED)` guard at line 110 blocks any retry — a failure mid-sequence leaves dangling S3 objects and an ingredient stuck without `generationCompletedAt`, requiring manual remediation. Accumulate the S3 keys, run `syncDataset`, then stamp the ingredient in a single trailing write, and make the path idempotent so a retry can complete it.

**5. Non-atomic training start — `darkroom-training-orchestrator.service.ts:156-179` (axis 6, Non-Atomic Mutations)**
`startTraining` does `trainingsService.create` (QUEUED) then `personasService.patch` (loraStatus) as two unwrapped writes. If the second fails, a QUEUED Training exists while the persona's status is stale. Narrowest blast radius of the three, but the same family of fix: make the persona update best-effort/at-most-once after the pipeline kicks off, so the only request-time guaranteed write is the Training create.

**6. Facade API expressed as positional parameter indices — `darkroom.service.ts:43-251` (axis 10, Directness vs Magic)**
The facade types 12 method parameters as `Parameters<PeerService['method']>[N]` (indices up to `[5]`). The worst, `options: Parameters<DarkroomGenerationService['generateImage']>[5]` (line 100), forces a reader to open another file, count to the sixth argument, and decode an unexported 10-field inline object. A facade exists to present a clean surface; positional-index types defeat that and break the moment a signature grows. Export named param types from each peer service (e.g. `export type GenerateImageOptions = { ... }`) and have the facade import them.

**7. `voiceId` cast without a runtime guard — `darkroom-media.service.ts:59-61` (axis 5, Type Structural Discipline)**
`(persona as Record<string, unknown>).voiceId as string | undefined` narrows `unknown` to `string` behind only a truthiness check — a truthy non-string (ObjectId, number) passes and reaches `elevenLabsService.generateAndUploadAudio` with a false type. `PersonaDocument` already has a `[key: string]: unknown` index signature, so the outer cast is redundant. Either add a `typeof rawVoiceId === 'string'` guard or, better, declare `voiceId?: string` on `PersonaDocument` and delete the cast.

### Minor

**8. Single-use interface file — `interfaces/darkroom-infra.interface.ts:1-7` (axis 2, Abstraction Earns Its Keep)**
`EC2InstanceStatus` is a 7-line file with exactly one importer (`darkroom-infra.service.ts`), and it duplicates `IEC2Instance` already exported from `packages/interfaces`. Reuse the shared type or inline it as a module-local type, and delete the file.

**9. Uncommented `as unknown as` cast — `darkroom-media.service.ts:104-109` (axis 5, Type Structural Discipline)**
`this.heyGenService as unknown as { getVideoStatus?: ... }` probes for an optional method; the intent is clear from the adjacent `typeof ... === 'function'` guard, but the cast carries no explaining comment. Add one line, or declare the optional method on `HeyGenService`.

**10. Redundant `notFoundMessage` arguments — `darkroom-character.service.ts:39-53` (axis 10, Directness vs Magic)**
Four callers pass `notFoundMessage` strings identical to the parameter's own default. Drop the redundant third argument at those call sites (generation ×2, media, orchestrator); keep the optional param, since `getCharacterBySlug` legitimately uses a different message.

**11. Character-list query duplicated — `darkroom-ingest.service.ts:122-128` (axis 3, Spaghetti Branching)**
`ingestTrainingDataForAllEnabledCharacters` calls `personasService.findAllByOrganization(org, { isDarkroomCharacter: true })` directly instead of `characterService.getCharacters(org)` (already injected and wrapping the identical query). Substitute it — same direction as #1, and it lets you drop `PersonasService` from this service entirely.

### Verdict: APPROVE — no structural blockers

The decomposition is a genuine improvement and the file-size, wiring, and fidelity axes are clean. Before merge, close the two owner-bypasses (#1, #2): they are one-line substitutions against already-injected services, and without them the "owner service" boundary this PR sells is fiction. The three non-atomic sequences (#3-#5) are real consistency hazards — note several appear carried over largely intact from the original god object rather than newly introduced, so at minimum file follow-up issues if they are not fixed here. Clean the facade's positional-index types (#6) while the surface is still fresh.

---
*Automated structural review — parallel axis-finders across the 10-axis structural rubric, then adversarial verification; confirmed findings only. Read-only: no tests, builds, or typechecks were run.*

PR #697, opened 2026-06-20
feat(workflows): workflow-backed daily trends digest + overview trends widget

VincentShipsIt:
## What

Ships the proactive **research trends agent** as the first workflow-backed conversion, plus the reusable primitives it needs.

### 1. Overview trends widget (no backend, no credits)
- New `OverviewTrendsPanel` (`@ui/overview`) rendered on **both** the Overview page and the Workspace dashboard, reading the already-collected global trend corpus via `useTrends()`. Always visible; "View all" → `/research/discovery`.

### 2. Daily Trends Digest = a predetermined, per-org **workflow** (toggled from the automation UI)
- **No new hard-coded cron** — driven by the canonical `WorkflowSchedulerService` per the dynamic-scheduling ADR.
- New reusable engine nodes:
  - `sendEmail` — generic email action (wired to `NotificationsService` via the adapter).
  - `trendDigest` — fetch + deterministic rank of the corpus (no scrape, no LLM), owner-email resolution, **durable per-day idempotency marker** (Redis `SET NX`, multi-replica safe), credit pre-check, branded render via the shared builder.
- `daily-trends-digest` template (`trendDigest → sendEmail` via explicit edge handles), registered in the executor registry, **node registry**, credit map, and template catalog.
- **Credit charged exactly once per confirmed send** via an adapter post-run hook, guarded by a separate durable Redis marker (no double-charge on retry/replica).
- Seeded **OFF** per org at the org-bootstrap chokepoint (`OrganizationSettingsService.create`, covering all 3 creation paths) + idempotent backfill seed script for existing orgs.
- Workflow-list **ON/OFF toggle** wired to the existing `setSchedule` endpoint.

### 3. Shared builder
- Pure trend-digest HTML/summary builder extracted to `@genfeedai/helpers`; the existing per-user trend-summary cron now consumes it.

## Root-cause scheduler/executor fixes (required for **any** workflow-backed schedule)
- Scheduler filtered `status: 'ACTIVE'` (uppercase) but the column is canonically **lowercase** → fixed to `WorkflowStatus.ACTIVE`. Durable scheduling was previously broken.
- Scheduled runs no longer flip `Workflow.status` to `COMPLETED`/`FAILED` (which dropped them from the scheduler after a single run) — they stay `ACTIVE`.

## Cron audit (answer to "are there crons that should be workflows?")
~24 tenant-product crons are workflow candidates (all default-ON); ~22 stay platform crons. This PR converts only the trends digest; the rest is a separate migration program (tracking issue to follow).

## Tests / verification
- `@genfeedai/workflow-engine`: 688 pass (incl. new `sendEmail`/`trendDigest` specs + registry/credit coverage).
- `@genfeedai/api`: adapter charge-hook (deduct-once + double-charge guard), template node-coverage, workflows/org-settings services, module — all green; `tsc --noEmit` clean.
- `@genfeedai/helpers`/`ui`/`app`: builder + overview panel + toggle specs green.
- type-check + lint green across all touched packages.

## Not in scope (follow-ups)
- Global workflow credit-deduction stays OFF (would bill every existing workflow).
- The broader ~24-cron → workflow migration.

### DIFF
--- a/apps/app/app/(protected)/[orgSlug]/[brandSlug]/overview/overview-dashboard-sections.test.tsx
+++ b/apps/app/app/(protected)/[orgSlug]/[brandSlug]/overview/overview-dashboard-sections.test.tsx
@@ -1,5 +1,6 @@
 import { AgentExecutionStatus, AgentExecutionTrigger } from '@genfeedai/enums';
 import type { IAgentRun } from '@genfeedai/interfaces';
+import type { TrendItem } from '@genfeedai/props/trends/trends-page.props';
 import type { PlatformTimeSeriesDataPoint } from '@props/analytics/charts.props';
 import { render, screen } from '@testing-library/react';
 import type { Key, ReactNode } from 'react';
@@ -10,6 +11,7 @@ import {
   OverviewPublishingInboxSection,
   OverviewStatusBadge,
   OverviewTopStatStrip,
+  OverviewTrendsPanel,
 } from './overview-dashboard-sections';
 
 vi.mock('next/link', () => ({
@@ -289,5 +291,63 @@ describe('Overview dashboard sections', () => {
       'href',
       '/posts/review',
     );
+  });
+
+  it('renders the overview trends panel with top trends and view-all link', () => {
+    const trends: TrendItem[] = [
+      {
+        createdAt: '2026-06-01T00:00:00.000Z',
+        expiresAt: '2026-06-08T00:00:00.000Z',
+        growthRate: 1.5,
+        id: 'trend-a',
+        isCurrent: true,
+        mentions: 1200,
+        platform: 'instagram',
+        requiresAuth: false,
+        topic: 'Reels growth',
+        viralityScore: 950,
+      },
+      {
+        createdAt: '2026-06-01T00:00:00.000Z',
+        expiresAt: '2026-06-08T00:00:00.000Z',
+        growthRate: 1.1,
+        id: 'trend-b',
+        isCurrent: true,
+        mentions: 800,
+        platform: 'tiktok',
+        requiresAuth: false,
+        topic: 'AI comedy skits',
+        viralityScore: 720,
+      },
+    ];
+
+    render(
+      <OverviewTrendsPanel
+        isLoading={false}
+        trends={trends}
+        viewAllHref="/research/discovery"
+      />,
+    );
+
+    expect(screen.getByText('Social Trends')).toBeInTheDocument();
+    expect(screen.getByText('Research Trends')).toBeInTheDocument();
+    expect(screen.getByRole('link', { name: /view all/i })).toHaveAttribute(
+      'href',
+      '/research/discovery',
+    );
+    expect(screen.getByText('Reels growth')).toBeInTheDocument();
+    expect(screen.getByText('AI comedy skits')).toBeInTheDocument();
+  });
+
+  it('renders the overview trends panel empty state', () => {
+    render(
+      <OverviewTrendsPanel
+        isLoading={false}
+        trends={[]}
+        viewAllHref="/research/discovery"
+      />,
+    );
+
+    expect(screen.getByText('No trends yet.')).toBeInTheDocument();
   });
 });
--- a/apps/app/app/(protected)/[orgSlug]/[brandSlug]/overview/overview-dashboard-sections.tsx
+++ b/apps/app/app/(protected)/[orgSlug]/[brandSlug]/overview/overview-dashboard-sections.tsx
@@ -11,12 +11,15 @@ import type { TableColumn } from '@props/ui/display/table.props';
 import type { OverviewBootstrapPayload } from '@services/auth/auth.service';
 import Card from '@ui/card/Card';
 import AppTable from '@ui/display/table/Table';
+import { OverviewTrendsPanel } from '@ui/overview/OverviewTrendsPanel';
 import { WorkspaceSurface } from '@ui/overview/WorkspaceSurface';
 import { Button } from '@ui/primitives/button';
 import dynamic from 'next/dynamic';
 import Link from 'next/link';
 import { useMemo } from 'react';
 import { HiOutlineArrowRight } from 'react-icons/hi2';
+
+export { OverviewTrendsPanel };
 
 const PlatformTimeSeriesChart = dynamic(
   () =>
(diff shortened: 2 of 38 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

Structural / maintainability pass on the workflow-backed trends digest. Methodology: seven parallel structural finders (file size, abstraction, branching, duplication, layer/type discipline, incidental complexity, async/atomicity) followed by adversarial verification of every finding against the checked-out branch — only confirmed findings appear below. The two blockers were additionally re-verified by hand against the schema and `wc -l`.

**Headline:** the feature *direction* is right — it replaces a hard-coded cron with a per-org workflow via the canonical `WorkflowSchedulerService` (per the dynamic-scheduling ADR), and the new executors/helpers carry real test coverage. But it ships a digest email with a broken video mapping, and it bolts 263 lines of cohesive new logic onto an adapter that is already 2.3× over the file-size ceiling. Both must be addressed before merge.

**1. Broken video field mapping — `workflow-engine-adapter.service.ts:795-797` (functional regression).**
The video branch of `buildDigestTrends` reads `video.url` and `video.views || video.playCount`. `TrendingVideoDocument` has none of those fields — the canonical names are `videoUrl` (or `playUrl`) and `viewCount`. Because the schema declares `[key: string]: unknown`, the compiler never flags it and **both resolve to `undefined` at runtime**: every digest email renders its videos with no link and no view count. The sibling branches are correct (sounds use `playUrl`/`usageCount`, hashtags use `postCount`/`viewCount`) — only video is wrong, which is exactly why it slipped through.
*Remedy:* `url: video.videoUrl`, `usageCount: video.viewCount`. Then delete the `RawTrend` local type and the `as unknown as RawTrend[]` cast (lines 754-769) and type the three fetcher lambdas with their real return types — the cast is what hid this mismatch from the compiler, and it will hide the next one.

**2. File-size ceiling breach — `workflow-engine-adapter.service.ts` is now 2,625 lines.**
Pre-PR it was 2,362; this PR adds 263 (`registerTrendDigestExecutor`, `registerSendEmailExecutor`, `buildDigestTrends`, `applyScheduledDigestCharge`, `digestUtcDateKey`). That new logic is self-contained and cohesive — it is a *digest* concern, not generic adapter plumbing. The pre-existing bloat is not this PR's job to fix, but new cohesive logic must not inflate a file that is already 2.6× over the 1,000-line ceiling.
*Remedy:* extract the digest registration + helpers into a dedicated collaborator (e.g. `TrendDigestWorkflowService`, or a focused adapter submodule) and have the adapter delegate to it. This also makes blocker #1's mapping logic unit-testable in isolation.

**3. `platforms` config is dead — `adapter:714-716` / `:749-753`.** The provider contract declares `platforms: string[]` and the executor passes it, but the wiring closure destructures only `{ topN, minViralScore }`, and `buildDigestTrends` never accepts or applies it. The per-org platform selection advertised in the template and node-registry has zero effect. Thread it through to the `trends.get*` calls, or remove it from the contract/template/registry — a documented dead knob is worse than a missing one.

**4. Idempotency lock consumed on transient skips — `trend-digest-executor.ts:194-215`.** The lock is acquired *before* the credit check and trends fetch. A `no-trends` skip (corpus not yet ingested for the day) permanently burns the day's run slot for ~26h, silently dropping the digest even if trends populate an hour later. Acquire the lock only at the "about to send" point, or explicitly do not consume it on the `no-trends` path.

**5. Charge lock never released on failure — `adapter:865-882`.** In `applyScheduledDigestCharge` the charge-guard lock is held for 26h even when `deductCreditsFromOrganization` throws and the error is swallowed. The email was already sent, so the result is a silent zero-charge with no retry path for the rest of the day. Release the lock in the `catch` (the sibling `withLock` helper already shows the try/finally pattern), or escalate the error instead of absorbing it.

**6. node-registry missing `skipped`/`reason` ports — `node-registry.ts:1198-1244`.** `trendDigest` omits its `skipped`/`reason` *outputs* and `sendEmail` omits the matching *inputs*, yet the `daily-trends-digest` template wires edges for all four and the `sendEmail` executor reads them. `validateConnection()` returns `false` for handles it can't find, so the visual editor marks those template edges invalid. Add the four port declarations so the registry matches the executors and template.

**7. Duplicated trend fetch/map + node-type literals.** `RawTrend` (`adapter:754-769`) is a field-for-field duplicate of `TrendRecord` in the cron this PR supersedes, and the getViral*/getTrending* → map logic is reimplemented rather than shared. Separately, `applyScheduledDigestCharge` (`adapter:840-844`) matches `'trendDigest'`/`'sendEmail'` by inline string literal instead of the executors' exported `nodeType` — rename a node type and the charge hook silently stops firing. Consolidate the fetcher; reference the constants.

### Minor (DRY drift introduced within this PR)

- **Four duplicated source-of-truth values:** `TrendDigestEntry` duplicates `TrendDigestItem` (`@genfeedai/helpers`); `digestUtcDateKey()` duplicates `utcDateKey()`; `DEFAULT_CREDIT_COST = 5` duplicates `TREND_DIGEST_CREDIT_COST`; the `93_600` TTL literal duplicates `IDEMPOTENCY_TTL_SECONDS`. The two UTC-date-key copies are the dangerous one — a one-character drift between the `workflow-digest:` and `workflow-digest-charged:` keys would double-send or double-charge. Share each from a single location.
- **`process.env` read directly** at `cron.trend-summary-notifications.service.ts:183` (`GENFEEDAI_APP_URL`) — repo rule is ConfigService only, which the adapter honors (`:683`). Plus a dead `import process` at `workflows.service.ts:1`.
- **Repeated three-way `type === 'video'|'hashtag'|'sound'` split** in both helper renderers (`trend-digest.helper.ts:66-68`, `:116-118`) and in `buildDigestTrends` — table-drive it so a fourth trend type is one edit, not four.

### Approved axes (clean)

Dynamic-scheduling discipline (no new hard-coded cron — `WorkflowSchedulerService` per ADR), test coverage on the new executors/helpers, frontend async hygiene on the overview widget, import order elsewhere — no findings.

---

**Verdict: BLOCK.** Two blockers: (1) the video field mapping ships a visibly degraded digest email and must be corrected (and the masking cast removed); (2) the 263 lines of new digest logic must be extracted rather than piled onto the 2,625-line adapter. The billing/idempotency lock-retention issues (#4, #5) and the dead `platforms` config (#3) are close behind and should land in the same pass — they are correctness gaps in the feature this PR exists to ship.

Issue #1, opened 2026-02-22
Add Prometheus Grafana Monitoring


VincentShipsIt:
# PRD: Add Prometheus Grafana Monitoring

## Executive Summary
Deploy Prometheus + Grafana for metrics and dashboards. The work is bounded to observable shared product surface behavior, clear infrastructure ownership, and verification evidence that can be reviewed from the issue or implementing PR.

## Problem Statement
Genfeed does not yet have a verified shared product surface contract for Prometheus Grafana Monitoring. Without this work, users or operators cannot rely on "Expose the Prometheus Grafana Monitoring capability through the affected shared product surface" being implemented, recoverable, and reviewable from public issue or PR evidence.

## Goals
- Expose the Prometheus Grafana Monitoring capability through the affected shared product surface
- Handle the primary success path and blocked/error state without private session context
- Capture verification evidence in the implementing PR or completion comment before release or closure

## Non-Goals
- Expand beyond this issue and linked parent epic.
- Introduce unrelated refactors or compatibility layers.
- Move secrets, private credentials, or private paid content into public issue text.

## User Stories
- As a Genfeed user or operator, I want to expose the Prometheus Grafana Monitoring capability through the affected shared product surface so that the shared product surface capability is reliable without hidden session context.
  **Acceptance:** WHEN the primary scenario is exercised, THE SYSTEM SHALL satisfy "Expose the Prometheus Grafana Monitoring capability through the affected shared product surface" on the affected shared product surface; IF the scenario cannot complete, THE SYSTEM SHALL expose a recoverable blocked, empty, or error state.
- As a maintainer, I want the scope, dependencies, and verification evidence captured in the issue so that progress decisions are reviewable.
  **Acceptance:** WHEN implementation or deferral is proposed, THE SYSTEM SHALL reference the Success Criteria and Verification Plan; IF scope changes, THE SYSTEM SHALL update this PRD before the issue moves forward.

## System Specification
- Observable states: The affected shared product surface exposes ready, empty, blocked/error, and completed states where they apply to this workflow.
- Data contract: Durable records, events, configuration, generated assets, or operational evidence required by the functional requirements have one canonical owner and stable identifiers.
- Permissions and trust boundary: User, organization, brand, workspace, and local-only boundaries are enforced at API, worker, desktop, CLI, or UI entry points before data is read or mutated.
- Integration points: The implementation uses existing Genfeed app/API/package boundaries and serializers for public response contracts rather than adding parallel shortcuts.
[...]

Issue #574, opened 2026-06-13
P0: v0.1.2 prod deploy blocked — EE billing circular-dep TDZ crashes api on boot (CI never boots EE bundle)


VincentShipsIt:
## Summary

The `v0.1.2` production deploy fails: the **api** service crashes on boot in the **EE/SaaS image** with a circular-dependency **temporal-dead-zone (TDZ)** error introduced by the ee/billing decouple (#563). The deploy's health-check auto-rolled api back to `v0.1.1`, so **production is serving and there is no outage**, but prod is **split-version** and `v0.1.2` cannot fully ship until the cycle is fixed.

## Impact / current production state

| Service | Image | Status |
|---|---|---|
| api | `ee8a682e4` (v0.1.1) | ✅ auto-rolled-back, healthy |
| notifications / mcp / files | `39498164d` (v0.1.2) | ✅ healthy |
| redis | v0.1.2 | ✅ healthy |
| workers / clips / discord / slack / telegram | `ee8a682e4` (v0.1.1) | ✅ healthy (waves 3–4 never ran) |
| Vercel frontends (app/docs/website/admin) | prior | ⛔ not deployed (skipped — needs backend) |

## Crash

```
ReferenceError: Cannot access 'userSubscriptions' before initialization
  at get userSubscriptions (main.js)
  at apps/server/api/src/collections/user-subscriptions/user-subscriptions.module.ts
  at apps/server/api/src/services/integrations/stripe/stripe.module.ts
  at ee/packages/billing/src/billing.providers.ee.ts
  at apps/server/api/src/collections/subscriptions/subscriptions.module.ts
  at apps/server/api/src/collections/organizations/organizations.module.ts
Exited with code 1
```

## Root cause

ESM imports are hoisted and fully evaluated before a module body runs, so:

```
organizations.module → subscriptions.module → @billing-providers (billing.providers.ee.ts)
  ↳ top-level  import { StripeModule }   (eager, billing.providers.ee.ts:35)
      → stripe.module → user-subscriptions.module
          → @Module(userSubscriptions)  reads the `userSubscriptions` const
            …but billing.providers.ee.ts's body has not run yet → TDZ
```

[...]

Issue #476, opened 2026-06-09
Self-hosted: first-boot experience — postgres + prisma migrate + mcp in one compose up


VincentShipsIt:
## Problem

The self-hosted Docker Compose path has several gaps that break a cold first-boot ("stranger path"):

- `prisma migrate deploy` must run before the API starts, but compose service ordering does not guarantee this
- MCP service startup ordering relative to API is not guarded
- No single documented command gets a fresh install from `git clone` to a running stack
- The compose fix for service ordering ships in PR `feat/fix-devops-compose`; this issue tracks docs and end-to-end smoke-testing

Audit evidence: production-readiness audit, area "Self-hosted / devops".

## Scope

This is a **verification and UX tracker**, not a code fix:

1. Confirm `feat/fix-devops-compose` actually resolves the race (`postgres` → `prisma migrate deploy` → `api` → `mcp`)
2. Write or update `docs/self-hosting.md` with the single-command quick-start
3. Smoke-test the stranger path end-to-end (fresh machine, `docker compose up`, app reachable with seed data)

## Acceptance criteria

- [ ] `docker compose up` on a clean checkout brings all services up without manual intervention
- [ ] `prisma migrate deploy` runs successfully before API accepts traffic
- [ ] `docs/self-hosting.md` includes a verified quick-start section
- [ ] Smoke-test checklist completed: login, create content, trigger a workflow, check MCP reachable
- [ ] CI job or manual sign-off documents passing stranger path

VincentShipsIt:
Verified done — shipped as part of the devops/compose work in #487. `docker-compose.selfhosted.yml` adds `postgres:16-alpine` with a `pg_isready` healthcheck and `genfeed depends_on: postgres (service_healthy)` (the ordering race); `selfhosted-entrypoint.sh` runs `bunx prisma migrate deploy` (5-attempt retry) **before** any service starts, then launches api/workers/files/notifications/mcp(3014)+app; `docs/self-hosting.md` carries the single-command `docker compose up` Quick Start. The stranger-path smoke test is automated and **green on master** via `e2e-selfhosted-release.yml` (latest success 2026-06-19). All acceptance criteria met — closing (board → Done).

PR #710, opened 2026-06-21
test(api): guard raw-SQL column-case bug class across server raw-SQL sites (#543)

### LINKED ISSUE #543: Audit OSS/cloud/self-hosted/desktop boundary + hunt sibling raw-SQL column-case bugs
**Status: P0**

Follow-up from a debugging session. Two production-breaking bugs were found and **already fixed** on branch `codex/remove-clerk-publishable-key` (pending PR → develop):

- `842e8105b` fix(build): externalize `recheck` — webpack bundled `recheck → synckit → @pkgr/core`, which reads `require.extensions` (stubbed to `(void 0)`), crashing **every** server on boot with `Object.keys(undefined)`.
- `32ac1080e` fix(analytics): raw `$queryRaw` used snake_case columns (`engagement_rate`, `brand_id`, …) but Prisma created camelCase columns (no `@map` on `PostAnalytics` fields) → every analytics endpoint failed with `42703 column does not exist`. Verified fixed against live DB.

Both are implementation defects, not design flaws. The local/cloud/self-hosted/desktop split itself is sound (`packages/config/src/mode.ts`, ADR-PLG-BOUNDARY). But the rest of the boundary was **not** audited. This issue tracks that verification + hunting for the same bug classes elsewhere.

## Scope / checklist

- [ ] **Sibling raw-SQL bugs** — grep all `$queryRaw` / `Prisma.raw` across `apps/server/**` for snake_case quoted identifiers that should be camelCase (same class as the analytics fix). Likely suspects: any service hitting `post_analytics`, `posts`, `brands` via raw SQL.
- [ ] **recheck externalization holds for all 12 server bundles** — confirm no other server (clips, workers, mcp, …) still bundles `recheck`/`@pkgr/core`. The fix is in shared `webpack.base.config.js`; verify each `main.js`.
- [ ] **LOCAL mode boots standalone** — `GENFEED_CLOUD` unset + `CLERK_SECRET_KEY` unset: no required path hard-depends on Clerk or the Genfeed API (auth guards, bootstrap, `@Public()` coverage).
- [ ] **Self-hosted single-command boot** — `docker compose -f docker/docker-compose.selfhosted.yml up` stands alone: postgres + `prisma migrate deploy` + services, no external hard deps.
- [ ] **ee/ multi-tenancy boundary** — all multi-tenant data-access code lives under `ee/` or imports from `ee/packages/`; OSS core single-tenant safe.
- [ ] **Desktop local-first** — BYOK generation works without Clerk/API; PGlite isolation; no silent cloud upload (per `desktop_byok_generation` memory).

## Acceptance criteria

- Each box above either ✅ verified clean with evidence, or 🐞 spun into its own fix issue.
- Any raw-SQL column-case bug found is fixed and verified against the live schema (not just type-check).

## Related

- #476 self-hosted first-boot experience
- #227 define OSS/EE boundary
- #184 OSS production readiness
- #262 verify offline desktop packaging

VincentShipsIt:
## What

Implements **#543** — *Audit OSS/cloud/self-hosted/desktop boundary + hunt sibling raw-SQL column-case bugs*. Adds a **schema-derived regression guard** that makes the production-breaking raw-SQL column-case bug class (the #574 sibling hunt) un-reintroducible, after auditing every server raw-SQL site and finding the current state clean.

## Background

Prisma maps **tables** to snake_case (`@@map("post_analytics")`) but leaves **columns** camelCase with **no `@map`**, so Postgres stores them as quoted camelCase identifiers (`"engagementRate"`, `"brandId"`). A raw-SQL string that double-quotes a column in snake_case (`"engagement_rate"`, `"brand_id"`) throws `42703 column does not exist` at runtime — invisible to typecheck/lint/vitest (raw SQL is an opaque string). That exact defect took down every analytics endpoint in production (fixed in `32ac1080e`). This issue asked to hunt siblings and harden the boundary.

## Audit findings (issue checklist)

- ✅ **Sibling raw-SQL bugs** — **clean**. Verified all 8 server raw-SQL files via an 8-agent cross-reference workflow + a deterministic schema sweep. Every quoted snake_case identifier in raw SQL is a legitimate `@@map` table (`post_analytics`, `credit_transactions`, `ad_performance`, `trend_remix_lineages`, `trend_source_references`); **zero** snake_case column refs; all columns are correctly quoted camelCase matching real schema fields; no `@map` columns mis-referenced; no `$queryRawUnsafe` string-building; the one dynamic column helper is typed `'brandId' | 'organizationId'`. No code fix needed — this PR locks that state in with a permanent guard.
- ✅ **recheck externalization** — clean (config-level): `webpack.base.config.js` externalizes `recheck` via the shared base used by all 12 server builds; dist confirms `"recheck"` as an external module. Runtime per-bundle confirmation is the EE-boot CI gate added in #575.
- ◻️ **LOCAL / self-hosted / desktop boot** — static signals positive (no `CLERK_SECRET_KEY` hard-dep in `main.ts`/`app.module.ts`, 32 `@Public()` sites, webpack aliases EE→OSS billing stub so the community image never bundles EE). Live single-command boot verification needs Docker/runtime and is out of scope for the laptop-scoped automation session; it remains covered by **#476** (self-hosted first-boot) and **#262** (offline desktop packaging).

## The guard (`raw-sql-column-case.spec.ts`)

Anchors on each raw-SQL call site (`$queryRaw` / `$executeRaw` / `Prisma.sql` / `Prisma.raw`), reads the following template literal / string with full `${…}` + nested-backtick awareness, and asserts every double-quoted snake_case identifier is a real `@@map` table or `@map` column from `schema.prisma` — never an un-mapped snake_case column reference. Allowlisted: mapped tables/columns, Prisma implicit-relation artifacts (`"A"`/`"B"`, `_junction`), camelCase columns, and double-quoted output aliases. Non-vacuity floors fail loudly if the extractor ever degrades to a no-op.

Hardened after an adversarial multi-agent review of the extractor that surfaced (then closed) blind spots: keyword-less `Prisma.sql` fragment helpers, nested-backtick splitting, and lazy-regex overrun across adjacent `Prisma.raw` calls — all replaced by the marker-anchored char scanner. Interpolated identifiers (`Prisma.raw(\`"${column}"\`)`) are guarded by their TS string-literal union type, noted inline.

## Verification

- `bunx vitest run src/__tests__/raw-sql-column-case.spec.ts` → **3 passed**.
- `bunx biome check` on the file → clean.
- Negative test: temporarily planting `"brand_id"` into a `Prisma.sql` fragment helper makes the guard fail with the exact file + snippet, then reverted — proving it covers the previously-missed fragment-helper vector.
- Scoped to the touched package per laptop resource limits; full suite runs on master in CI.

Closes #543

### DIFF
--- a/apps/server/api/src/__tests__/raw-sql-column-case.spec.ts
+++ b/apps/server/api/src/__tests__/raw-sql-column-case.spec.ts
@@ -1,0 +1,334 @@
+import { readdirSync, readFileSync, statSync } from 'node:fs';
+import { dirname, join, relative } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { describe, expect, it } from 'vitest';
+
+/**
+ * Regression guard for the raw-SQL column-case bug class (#574 sibling hunt, #543).
+ *
+ * Prisma tables ARE snake_case (every model carries `@@map("snake_table")`), but
+ * almost all COLUMNS are camelCase with NO `@map`, so Prisma creates them as
+ * double-quoted camelCase identifiers (e.g. `"engagementRate"`, `"brandId"`).
+ * Postgres folds UNQUOTED identifiers to lowercase and a camelCase column is only
+ * reachable when written double-quoted EXACTLY (`"engagementRate"`). Therefore a
+ * raw-SQL string that double-quotes a column in snake_case (`"engagement_rate"`,
+ * `"brand_id"`) throws `42703 column does not exist` at runtime — invisible to
+ * typecheck, lint and vitest (raw SQL is an opaque string).
+ *
+ * That exact defect took down every analytics endpoint in production (fixed in
+ * `32ac1080e`). This test makes the bug class un-reintroducible: it scans every
+ * raw-SQL site under `apps/server/**` and asserts that any double-quoted
+ * snake_case identifier is a real DB table (`@@map`) or a `@map`'d column from
+ * the Prisma schema — never an un-mapped snake_case column reference.
+ *
+ * Scope note: SQL is collected by anchoring on the raw-SQL call site itself
+ * (`$queryRaw`/`$executeRaw`/`Prisma.sql`/`Prisma.raw`) and reading the template
+ * literal or string that follows, with full `${…}`/nested-backtick awareness.
+ * That captures keyword-less fragment helpers (`Prisma.sql\`AND "brandId" = …\``)
+ * and nested fragments alike, while ignoring ordinary log templates. The one
+ * shape it can't see through is an *interpolated* identifier
+ * (`Prisma.raw(\`"${column}"\`)`); that column is always a typed string-literal
+ * union (e.g. `'brandId' | 'organizationId'`), so the TS type — not this test —
+ * is the guard there.
+ */
+
+const SRC_ROOT = join(dirname(fileURLToPath(import.meta.url)), '..');
+// apps/server/api/src -> apps/server
+const SERVER_ROOT = join(SRC_ROOT, '..', '..');
+// apps/server/api/src -> repo root -> packages/prisma/prisma/schema.prisma
+const SCHEMA_PATH = join(
+  SRC_ROOT,
+  '..',
+  '..',
+  '..',
+  '..',
+  'packages',
+  'prisma',
+  'prisma',
+  'schema.prisma',
+);
+
+const RAW_SQL_MARKERS = [
+  '$queryRaw',
+  '$executeRaw',
+  'Prisma.sql',
+  'Prisma.raw',
+];
+
+// Longest alternatives first so `$queryRawUnsafe` is not partially matched by
+// `$queryRaw`.
+const MARKER_PATTERN =
+  /\$queryRawUnsafe|\$executeRawUnsafe|\$queryRaw|\$executeRaw|Prisma\.sql|Prisma\.raw/g;
+
+interface SqlViolation {
+  file: string;
+  identifier: string;
+  snippet: string;
+}
+
+/**
+ * Every legitimate snake_case identifier the raw SQL is allowed to double-quote:
+ * `@@map("table")` table names and `@map("column")` mapped column names. Built
+ * straight from the schema so it can never drift from the real DB shape.
+ */
+function buildSnakeAllowlist(): Set<string> {
+  const schema = readFileSync(SCHEMA_PATH, 'utf-8');
+  const allow = new Set<string>();
+  const mapPattern = /@@?map\(\s*"([^"]+)"\s*\)/g;
+  let match: RegExpExecArray | null;
+  while ((match = mapPattern.exec(schema)) !== null) {
+    allow.add(match[1]);
+  }
+  return allow;
+}
+
+function walkTsFiles(dir: string): string[] {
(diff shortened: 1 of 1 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**Scope:** `apps/server/api/src/__tests__/raw-sql-column-case.spec.ts` (+334, single new file). Structural/maintainability lens only (the thermo-nuclear rubric) — correctness, security, and coverage belong to other harnesses. Every finding below was checked against the actual code by an adversarial verifier; only confirmed defects are listed.

**Headline:** structurally sound. Well under the 1,000-line rule (334 lines), good diagnostic output, no layering or abstraction smells. One real defect undercuts the guard's own purpose; two cosmetic cleanups.

### Requested changes

**1. Marker pre-filter silently drops `$queryRawUnsafe` / `$executeRawUnsafe` sites — `raw-sql-column-case.spec.ts:51-56` (axis 10, Directness vs Magic)**
`RAW_SQL_MARKERS` (the file pre-filter at line 283) lists only `$queryRaw`, `$executeRaw`, `Prisma.sql`, `Prisma.raw`. `MARKER_PATTERN` (lines 60-61) additionally matches `$queryRawUnsafe` and `$executeRawUnsafe`. Because the pre-filter gates file selection *before* `extractSqlRegions` ever runs, any source file whose only raw-SQL calls are the `Unsafe` variants is excluded from analysis entirely — the guard becomes a silent no-op for exactly the call shape most prone to injection and column-case bugs. The two parallel representations of "what counts as raw SQL" will keep drifting; the comment on line 58 even acknowledges the `Unsafe` variants exist but still omits them from the array.
**Fix:** delete `RAW_SQL_MARKERS` and drive the pre-filter from the single source of truth — `MARKER_PATTERN.test(content)` (reset `lastIndex = 0` first, since it is a global regex). One marker set, no drift, no blind spot.

### Minor

**2. Dead assertion after manual throw — `raw-sql-column-case.spec.ts:320-338` (axis 9, Design Purity)**
The `if (violations.length > 0) { throw new Error(...) }` block makes the trailing `expect(violations).toEqual([])` unreachable on failure and vacuously true on success — it never does independent work. Drop the manual throw and pass the formatted report as vitest's second `expect` argument: `expect(violations, reportString).toEqual([])`. Same diagnostic, one branch deleted.

**3. Double disk read per matched file — `raw-sql-column-case.spec.ts:281-291` (axis 2, Abstraction Earns Its Keep)**
Each file that passes the marker filter is `readFileSync`'d twice — once in the `.filter()` callback (result discarded) and again in `.map()`. Collapse to a single `flatMap`/`reduce` that reads once and returns `{ file, regions }` when `regions.length > 0`. Test-only and runs once per CI cycle, so impact is small — but it is free to remove.

### Verdict: APPROVE — no structural blockers

File size, abstraction shape, and layering are clean. Fix #1 before merge: a regression guard that silently ignores `*Unsafe` raw-SQL is a guard with a hole in exactly the place it matters most. #2 and #3 are quick cleanups that can ride the same change.

---
*Automated structural review — parallel axis-finders across the 10-axis structural rubric, then adversarial verification; confirmed findings only. Read-only: no tests, builds, or typechecks were run.*

PR #717, opened 2026-06-21
fix(app): repoint workspace-shell E2E to the reworked brand switcher (#570)

### LINKED ISSUE #570: 🚦 Self-hosted release E2E failed — 2026-06-13
**Self-hosted release E2E failed** on `2026-06-13`.

- Image tag under test: `ghcr.io/genfeedai/genfeed.ai:latest`
- Failed run: https://github.com/genfeedai/genfeed.ai/actions/runs/27463517769
- Artifacts: `release-e2e-playwright-report`, `release-e2e-compose-logs`

This means the shipped container does not boot + serve cleanly in LOCAL mode.
Triage to **P0 status** on the project board (P0 is a status, not a label).

VincentShipsIt:
Closes #570.

## Root cause
The release/shell E2E specs bound to `desktop-sidebar-rail` → `workspace-switcher-trigger`, but **#667 reworked the switchers into the protected topbar** (`MenuBrandSwitcher` + `AppSwitcher`) and left the old `menus/workspace-switcher` **and** `topbars/workspace-switcher` `WorkspaceSwitcher` components **orphaned** (rendered nowhere — only a stale `vi.mock` referenced one). So:
- `release/workspace-loads.spec.ts` **failed** — "workspace switcher must mount" (the testid never renders).
- `shell/navigation-interactions` + `shell/chrome-interactions` silently **no-op'd** behind their `if (isVisible())` guards (vacuous passes).

## Fix
- **Add a stable `brand-switcher-trigger` testid** to the real shell switcher (`MenuBrandSwitcher` trigger), which renders the seeded brand label.
- **Repoint all three specs** to it:
  - `workspace-loads` asserts the seeded **"Default Brand"** label — the live-integration proof (seed: `self-hosted-seed.service.ts` org `Default Workspace` / brand `Default Brand`).
  - the two shell specs exercise the per-brand **"Open … settings"** action (the affordance that only renders in the open dropdown) and stay resilient.
- **Delete the orphaned dead code**: `menus/workspace-switcher` (7 files) + `topbars/workspace-switcher` (5 files), and the stale `TopbarWorkspaceSwitcher` `vi.mock`.
- **Guard the testid contract** with a `MenuBrandSwitcher` unit test, so a refactor that drops it fails fast in unit tests — not only in the nightly E2E.

Net **−1467 LOC**.

## Validation
- type-check ui+app **17/17**, biome lint + format **clean**, pre-commit guards (secretlint / no-raw-html) **pass**
- unit: `app-protected-layout` **23/23**, `AppLayout` **18/18**, `MenuBrandSwitcher` **5/5** (incl. the new testid contract test)
- The frontend E2E is nightly (no PR trigger), so the live spec runs there / on the next release image; assertions were derived from the components' actual rendered DOM (`SwitcherDropdown` trailing-action `ariaLabel`, confirmed by the existing `MenuBrandSwitcher` test).


VincentShipsIt:
Addressed in d4e14404b — replaced the `if (isVisible())` guards on the brand-switcher trigger with a mandatory `toBeVisible` assertion in both shell specs. The route is brand-scoped and `brand-1` is mocked in `api-mocks.fixture` (the sibling `page-context-contract` spec already asserts shell chrome mandatorily), so the switcher always mounts; masking its absence is exactly what let #570 hide. The per-brand "Open … settings" action stays soft — it's genuinely omitted for brands without a slug.

PR #725, opened 2026-06-21
fix(scripts): reseed voice catalog fails closed on missing API key

VincentShipsIt:
## Summary

Fixes a LOW-severity operator footgun in the in-container voice-catalog reseed script from [#682](https://github.com/genfeedai/genfeed.ai/commit/e3a7bb5d5) (`e3a7bb5d5`).

`fetchElevenLabs` / `fetchHeyGen` logged a warning and `return []` when their API key was missing. `main()` has no zero-result guard, so a missing-credential run printed "0 voices" and **exited 0** — a false success that, in live mode, writes nothing while reporting success.

## Fix

Throw on a missing key (caught by `main()` → `exit 1`), matching the `.ts` variant's fail-closed behavior.

## Verification

- `node --check`: syntax OK.
- Biome: clean.
- No test framework covers this standalone migration script; the change is a guard inversion (`return []` → `throw`) verified by reading the `main()` try/catch → `process.exit(1)` path.

---

_From the automated daily commit review (2026-06-21). Related: #718, #720, #721, #723, #724._

### DIFF
--- a/scripts/migrations/reseed-voice-catalog.incontainer.mjs
+++ b/scripts/migrations/reseed-voice-catalog.incontainer.mjs
@@ -36,8 +36,11 @@ function genId() {
 async function fetchElevenLabs() {
   const key = process.env.ELEVENLABS_API_KEY;
   if (!key) {
-    console.error('ELEVENLABS_API_KEY missing — skipping ElevenLabs');
-    return [];
+    // Throw rather than return [] so a missing credential fails the run loudly
+    // (caught by main() -> exit 1), instead of silently reporting "0 voices" as
+    // success and — in live mode — writing nothing while exiting 0. Matches the
+    // .ts variant's fail-closed behavior.
+    throw new Error('ELEVENLABS_API_KEY is not set');
   }
   const res = await fetch('https://api.elevenlabs.io/v1/voices', {
     headers: { 'xi-api-key': key, accept: 'application/json' },
@@ -56,8 +59,8 @@ async function fetchElevenLabs() {
 async function fetchHeyGen() {
   const key = process.env.HEYGEN_KEY;
   if (!key) {
-    console.error('HEYGEN_KEY missing — skipping HeyGen');
-    return [];
+    // Fail closed on a missing credential (see fetchElevenLabs).
+    throw new Error('HEYGEN_KEY is not set');
   }
   const res = await fetch('https://api.heygen.com/v2/voices', {
     headers: { 'X-Api-Key': key, accept: 'application/json' },


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**Verdict: BLOCK**

The stated goal — fail closed on a missing API key — is undermined by coupling both provider fetches inside a single `Promise.all` with one shared catch/exit. A missing `HEYGEN_KEY` aborts the ElevenLabs result and writes **zero** voices even when ElevenLabs is fully available — a net regression from the old soft-skip behavior. Worse, the explanatory comment falsely claims parity with the `.ts` variant, which uses per-provider `try/catch` + `continue` (the opposite of all-or-nothing). The intent (fail loud on a missing key) is sound; the implementation needs per-provider error isolation before the `exit 1` signal.

### 🚫 Blockers

**1. `scripts/migrations/reseed-voice-catalog.incontainer.mjs:82-89` — `Promise.all` couples independent providers under one catch + `process.exit(1)`**
A missing key in either provider rejects the whole settled value, discarding any voices successfully fetched from the other. The script exits `1` before any DB writes even when the working provider returned data. This is a correctness regression — the old code at least seeded the available provider.

Run each fetch in its own `try/catch`, collect per-provider results, complete the DB upserts with whatever was gathered, then `process.exit(1)` after writes if any provider was skipped:
```js
const voices = [];
let anyFailed = false;
try { voices.push(...await fetchElevenLabs()); } catch (e) { console.error(e.message); anyFailed = true; }
try { voices.push(...await fetchHeyGen()); }     catch (e) { console.error(e.message); anyFailed = true; }
// ... upsert `voices` ...
if (anyFailed) process.exit(1);
```
Preserves fail-loud signaling for the missing key without discarding the working provider's data.

**2. `scripts/migrations/reseed-voice-catalog.incontainer.mjs:39-42` — comment factually wrong**
The comment states the throw *"Matches the .ts variant's fail-closed behavior."* The `.ts` variant (`reseed-voice-catalog.ts:197-200`) wraps each provider in its own `try/catch` with `continue` — it **skips** the failing provider and proceeds. The `.mjs` all-or-nothing semantic diverges from its stated reference, and the comment will mislead every future operator into thinking partial-key environments are handled identically. Replace with an accurate description, e.g.: *"Unlike the .ts variant (which catches per-provider and continues), this script is all-or-nothing: both provider keys must be present or the entire run aborts."* (Best paired with the fix in blocker 1, which makes per-provider isolation the actual behavior.)

**1. `scripts/migrations/reseed-voice-catalog.incontainer.mjs:37-43, 60-63` — duplicated key-guard logic**
`fetchElevenLabs` and `fetchHeyGen` contain identical `if (!key) throw new Error(...)` guards differing only in the env-var name; the second even cross-references the first (`see fetchElevenLabs`) — a textbook signal they should be one function. Extract `requireEnvKey(name)` with a JSDoc preserving the fail-closed rationale, and call `requireEnvKey('ELEVENLABS_API_KEY')` / `requireEnvKey('HEYGEN_KEY')`. Single canonical place for the intent.

---
*Read-only review: failure-mode coupling, fail-closed semantics, comment accuracy, and cross-script parity assessed by reading the diff, the `.mjs`, and the reference `.ts` at PR head. No scripts were run. Findings adversarially verified — 4/4 raw findings survived.*

PR #726, opened 2026-06-21
fix(ci): guard empty ECS task ARN and bound the stop-wait in prod deploy

VincentShipsIt:
## Summary

Hardens the prod ECS deploy from [#713](https://github.com/genfeedai/genfeed.ai/commit/93caab4ff) (`93caab4ff`). Both the **migrate** and **boot-smoke** steps did:

```bash
TASK_ARN=$(aws ecs run-task ... --query 'tasks[0].taskArn' --output text)
aws ecs wait tasks-stopped --cluster "$CLUSTER" --tasks "$TASK_ARN"
```

Two LOW-severity failure modes:
- If `run-task` launched **no task** (capacity / IAM / quota), the query yields the literal `None`, and `wait ... --tasks None` failed with a cryptic parameter-validation error.
- The bare `wait tasks-stopped` polls **silently for ~10 min** then fails opaquely — no intermediate signal, no ECS stop reason.

## Fix (both steps)

- Guard `TASK_ARN` against empty/`None` with an actionable error (capacity/IAM/quota).
- Replace the bare wait with a **deadline-bounded `describe-tasks` poll** (10 min) that logs `lastStatus` each tick and surfaces `stoppedReason` on timeout.

Happy path is unchanged; only failure diagnosability and the timeout bound improve.

## Verification

- `python yaml.safe_load`: workflow parses.
- `bash -n` on both modified `run:` blocks: OK.
- secretlint: clean.

---

_From the automated daily commit review (2026-06-21). Related: #718, #720, #721, #723, #724, #725._

### DIFF
--- a/.github/workflows/deploy-ecs.yml
+++ b/.github/workflows/deploy-ecs.yml
@@ -241,7 +241,28 @@ jobs:
             --launch-type FARGATE \
             --network-configuration "awsvpcConfiguration={subnets=[$SUBNETS],securityGroups=[$SG],assignPublicIp=DISABLED}" \
             --query 'tasks[0].taskArn' --output text)
-          aws ecs wait tasks-stopped --cluster "$CLUSTER" --tasks "$TASK_ARN"
+          # run-task emits the literal "None" when the tasks[] array is empty
+          # (capacity / IAM / quota failure). Guard so we fail with an actionable
+          # message instead of a cryptic "wait" parameter-validation error.
+          if [ -z "$TASK_ARN" ] || [ "$TASK_ARN" = "None" ]; then
+            echo "::error::ECS run-task launched no migrate task (capacity, IAM, or quota failure)"; exit 1
+          fi
+          # Bound the wait: bare 'aws ecs wait tasks-stopped' polls silently for
+          # up to ~10 min then fails opaquely. Poll describe-tasks with a deadline
+          # and surface lastStatus / stoppedReason for fast diagnosis.
+          DEADLINE=$((SECONDS + 600))
+          while :; do
+            STATUS=$(aws ecs describe-tasks --cluster "$CLUSTER" --tasks "$TASK_ARN" \
+              --query 'tasks[0].lastStatus' --output text)
+            [ "$STATUS" = "STOPPED" ] && break
+            if [ "$SECONDS" -ge "$DEADLINE" ]; then
+              REASON=$(aws ecs describe-tasks --cluster "$CLUSTER" --tasks "$TASK_ARN" \
+                --query 'tasks[0].stoppedReason' --output text)
+              echo "::error::migrate task did not stop within 10m (last status: $STATUS, reason: $REASON)"; exit 1
+            fi
+            echo "  migrate task status: $STATUS ($((DEADLINE - SECONDS))s remaining)"
+            sleep 6
+          done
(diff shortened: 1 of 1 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**Verdict: APPROVE**

The PR correctly fixes both failure modes it targets: the `None`-ARN guard closes the cryptic parameter-validation error path, and the deadline-bounded poll replaces the silent 10-minute black-box wait with per-tick status logging and a surfaced `stoppedReason`. Logic is sound, happy path unchanged. No blockers. One structural smell and two simplifications worth tracking before this pattern spreads.

### ⚠️ Structural (recommended, non-blocking)

**1. `.github/workflows/deploy-ecs.yml:244-265` (migrate step) and `288-309` (boot-smoke step) — verbatim copy-paste of the 21-line guard+poll**
The ARN-empty guard and bounded `describe-tasks` poll loop are duplicated across both steps; the only difference is the label string (`migrate task` vs `boot smoke task`). Any future change to poll interval, deadline arithmetic, error-message format, or early-exit logic (e.g. a `TASK_FAILED` short-circuit) must be applied in two places with no enforcement — one copy will drift.

Extract to a composite action at `.github/actions/wait-ecs-task/action.yml` with inputs `task-arn`, `cluster`, `label`. Each `run:` step in the composite must set `shell: bash` explicitly, and input expansions must be double-quoted in the shell body (`"${{ inputs.task-arn }}"`) to preserve the ARN-empty guard semantics. Keeps per-step status granularity in the Actions UI while killing the duplicate. `.github/actions/` already exists, so this is low-friction.

**1. `.github/workflows/deploy-ecs.yml:258-261` (migrate timeout path) and `302-305` (boot-smoke timeout path) — `stoppedReason` is always `None` on timeout**
On timeout the loop exits while the task is still `RUNNING`/`DEACTIVATING`, so the follow-up `describe-tasks` fetch of `stoppedReason` returns the literal `None` (that field only populates after the task stops). The error reads `reason: None` — less informative than `STATUS` alone, plus an extra API call on an already-failing path. Drop the call and emit status directly:
```bash
echo "::error::migrate task did not stop within 10m — last status: $STATUS"; exit 1
```
Or keep it with an inline comment noting `stoppedReason` is `None` while running, so future readers aren't confused by the CI logs.

**2. `.github/workflows/deploy-ecs.yml:240-249` (migrate ARN guard) and `284-293` (boot-smoke ARN guard) — discarded `failures[0].reason`**
When `run-task` returns `tasks: []` (capacity, IAM, or quota failure), `failures[]` holds the specific AWS reason (e.g. `RESOURCE:MEMORY`, `AGENT`, `ATTRIBUTE`). The current guard discards it and emits a hardcoded generic message, forcing a console trip to diagnose. Capture the JSON once and surface the reason without an extra call:
```bash
RUN_OUT=$(aws ecs run-task ... --output json)
TASK_ARN=$(echo "$RUN_OUT" | jq -r '.tasks[0].taskArn // empty')
FAIL_REASON=$(echo "$RUN_OUT" | jq -r '.failures[0].reason // empty')
# error message:
echo "::error::run-task started no task — ${FAIL_REASON:-capacity, IAM, or quota failure}"
```
Preserves the existing fail-closed behavior.

---
*Read-only review: shell-step control flow, ARN/empty-string guards, poll-loop bounds, and failure-mode coverage assessed by reading the diff and workflow at PR head. No CI was run. Findings adversarially verified — 6/6 raw findings survived.*

PR #727, opened 2026-06-21
fix(app): mood-board auto-placement overlap + cancellable asset paging

VincentShipsIt:
## Summary

Two LOW-severity issues in the library mood board from [#693](https://github.com/genfeedai/genfeed.ai/commit/e90926db8) (`e90926db8`).

### 1. New-asset auto-placement overlapped tall saved tiles
New (un-positioned) assets were placed at `maxSavedY + MOOD_BOARD_TILE_HEIGHT + GAP`, where `maxSavedY` is the **top-left y** of the bottommost saved tile. But tiles are fixed-width and keep the asset's aspect ratio, so a portrait tile renders **taller** than `MOOD_BOARD_TILE_HEIGHT` (e.g. a 4:5 asset is 350 tall, not 320) — the new row overlapped its lower edge.

**Fix:** compute each saved tile's **true bottom** (`y + rendered height`, derived from the ingredient aspect ratio, mirroring `MediaAssetNode`) and start the new band below the lowest bottom. Extracted `moodBoardTileHeight()` + `MOOD_BOARD_DEFAULT_ASPECT_RATIO` and pointed `MediaAssetNode` at the shared default so the two sizings can't drift.

### 2. Asset paging wasn't cancellable
`useBrandMediaAssets.fetchAllPages` checked `signal.aborted` *between* pages but never forwarded the signal to the HTTP request, so the in-flight page kept running after unmount (wasted bandwidth; no stale-state bug — the result guards already prevent that).

**Fix:** `BaseService.findAll` now accepts an optional `AbortSignal` and forwards it to axios (backward-compatible — existing callers unaffected); the hook threads its controller signal through.

## Tests

- Layout: new-asset placement clears a tall (portrait) saved tile's real bottom (regression: previously overlapped).
- `BaseService.findAll` forwards the `AbortSignal` to the axios request config.

## Verification

- `vitest`: utils **7**, services **52** (+1), hooks **3**, MediaAssetNode **2** — all pass.
- `tsc --noEmit` (utils/services/hooks): **0** non-noise errors (remaining `TS2307`s are pre-existing unbuilt-workspace-dep noise in this checkout).
- Biome + secretlint + no-raw-html: clean.

---

_From the automated daily commit review (2026-06-21). Related: #718, #720, #721, #723, #724, #725, #726._

### DIFF
--- a/apps/app/src/features/moodboard/MediaAssetNode.tsx
+++ b/apps/app/src/features/moodboard/MediaAssetNode.tsx
@@ -1,12 +1,15 @@
 'use client';
 
 import { isVideoIngredient } from '@genfeedai/utils/media/ingredient-type.util';
-import { MOOD_BOARD_TILE_WIDTH } from '@genfeedai/utils/moodboard/mood-board-layout.util';
+import {
+  MOOD_BOARD_DEFAULT_ASPECT_RATIO,
+  MOOD_BOARD_TILE_WIDTH,
+} from '@genfeedai/utils/moodboard/mood-board-layout.util';
 import Image from 'next/image';
 import { memo } from 'react';
 import type { MediaAssetNodeProps } from '@/features/moodboard/moodboard.types';
 
-const DEFAULT_ASPECT_RATIO = 4 / 5;
+const DEFAULT_ASPECT_RATIO = MOOD_BOARD_DEFAULT_ASPECT_RATIO;
 
 function resolveAspectRatio(width?: number, height?: number): number {
   if (typeof width === 'number' && typeof height === 'number' && height > 0) {
--- a/packages/hooks/data/moodboard/use-brand-media-assets/use-brand-media-assets.ts
+++ b/packages/hooks/data/moodboard/use-brand-media-assets/use-brand-media-assets.ts
@@ -24,7 +24,10 @@ const DISPLAYABLE_STATUSES = [
 ];
 
 interface PageableService {
-  findAll(query: Record<string, unknown>): Promise<IIngredient[]>;
+  findAll(
+    query: Record<string, unknown>,
+    signal?: AbortSignal,
+  ): Promise<IIngredient[]>;
 }
 
 function isVisualMediaIngredient(ingredient: IIngredient): boolean {
@@ -53,13 +56,18 @@ async function fetchAllPages(
       break;
     }
 
-    const batch = await service.findAll({
-      brand: brandId,
-      lightweight: true,
-      status: DISPLAYABLE_STATUSES,
-      limit: PAGE_SIZE,
-      page,
-    });
+    const batch = await service.findAll(
+      {
+        brand: brandId,
+        lightweight: true,
+        status: DISPLAYABLE_STATUSES,
+        limit: PAGE_SIZE,
+        page,
+      },
+      // Forward the abort signal so an in-flight page request is cancelled on
+      // unmount, not just checked between pages.
+      signal,
+    );
 
     collected.push(...batch);
 
--- a/packages/services/core/base.service.test.ts
+++ b/packages/services/core/base.service.test.ts
@@ -377,6 +377,19 @@ describe('BaseService', () => {
       });
     });
 
+    it('forwards an AbortSignal to the request config so the fetch is cancellable', async () => {
+      const mockResponse = { data: { data: [] } };
+      service.getInstanceForTest().get.mockResolvedValue(mockResponse);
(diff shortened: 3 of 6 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**PR #727 — `fix(app): mood-board auto-placement overlap + cancellable asset paging`**

Two fixes in one PR: (a) auto-placement now keys off each saved tile's rendered **bottom** edge via a new `moodBoardTileHeight` util, fixing overlap below tall tiles; (b) an `AbortSignal` is threaded through `BaseService.findAll` and the `use-brand-media-assets` pager so an in-flight page request cancels on unmount. Both fixes are sound and well-tested. The structural concerns below are about *how the abstractions landed*, not about whether the bugs are fixed. Findings confirmed by a 3-skeptic adversarial pass.

### Structural findings

**1. Aspect-ratio resolution is now duplicated and kept in sync only by a comment.** — `mood-board-layout.util.ts:15-32` and `MediaAssetNode.tsx:21-27`
`moodBoardTileHeight` and `MediaAssetNode.resolveAspectRatio` contain the *identical* guard: `typeof width === 'number' && typeof height === 'number' && height > 0 ? width / height : MOOD_BOARD_DEFAULT_ASPECT_RATIO`. Hoisting the constant was the right first move, but the resolution logic itself is copy-pasted, and the util's JSDoc literally says "Mirrors the `MediaAssetNode` sizing." That is coupling-by-comment — the exact failure mode this PR is fixing. If anyone tightens the guard in one place (a `width > 0` check, a `NaN` guard, a different fallback), the layout math silently diverges from the rendered height and tiles overlap again.
*Fix:* export a single `resolveAspectRatio(width?, height?): number` from `mood-board-layout.util.ts` (where the shared constants already live). `moodBoardTileHeight` calls it internally; `MediaAssetNode` imports it instead of reimplementing it. This deletes the sync comment and makes the renderer and the layout util provably consistent — one source of truth for tile geometry.

**2. Cancellation works, but by accident — aborts still fire a spurious `logger.error` on every unmount-during-fetch.** — `base.service.ts:226-235`
When the signal fires, Axios raises `CanceledError`; the response interceptor rejects with a plain `{ isCancelled: true, silent: true }` object. That object flows into `executeWithErrorHandling`'s catch → `handleOperationError`, which calls `logger.error(...)` **unconditionally** and then throws a wrapped `Error`. The hook's `catch` then checks `controller.signal.aborted` and returns silently, so the UI stays clean — but only because two independent guards happen to chain. The log noise still fires, and if `handleOperationError` is later changed to not wrap, or the hook's `aborted` check is dropped, aborts become user-visible errors.
*Fix:* add an early-exit for the cancellation shape before logging, e.g. in `executeWithErrorHandling`/`handleOperationError`: `if ((error as { isCancelled?: boolean }).isCancelled) throw error;`. That makes the abort path explicit instead of relying on the hook to absorb a logged-and-rethrown error.

**3. `PageableService` is a hand-maintained shadow of `BaseService.findAll`.** — `use-brand-media-assets.ts:30-36`
The local `interface PageableService { findAll(query, signal?) }` is a structural copy that must track `BaseService.findAll` by hand — this PR already had to edit both to add `signal?`. The cast `service as PageableService` only checks the local stub, so a future signature change (third param, different return) drifts silently. Another hook in the same package (`use-crud-modal.ts`) already imports `BaseService` directly, so the boundary allows it.
*Fix:* derive instead of copy — `type PageableService = Pick<BaseService<IIngredient>, 'findAll'>` — so the compiler enforces parity automatically.

### Simplification / hygiene

**4. Redundant module-level re-alias.** — `MediaAssetNode.tsx:19`
`const DEFAULT_ASPECT_RATIO = MOOD_BOARD_DEFAULT_ASPECT_RATIO` is a one-use alias of an already well-named import, pure indirection.
*Fix:* use `MOOD_BOARD_DEFAULT_ASPECT_RATIO` directly — and it disappears entirely once finding 1's shared helper lands.

### Verdict: **APPROVE**

Both bugs are genuinely fixed and the new tests assert the right invariants (the `> 560` tall-tile placement test and the signal-forwarding test are exactly on point). It clears merge as-is. That said, the overlap fix re-introduces the same duplicate-source-of-truth pattern it's curing (finding 1), and the abort path is correct only by coincidence (finding 2) — both are ~10-line refactors and strongly recommended before this becomes the reference implementation others copy. Finding 1 in particular should land before merge; it's cheap insurance against this exact bug recurring.

PR #728, opened 2026-06-21
fix(api): surface provisionDefaultWorkflows failures to Sentry

VincentShipsIt:
## Summary

Closes the observability gap in the org-bootstrap chokepoint from [#711](https://github.com/genfeedai/genfeed.ai/commit/f45b3869b) (`f45b3869b`).

All org-creation paths (Clerk webhook, `OrganizationsController`, `UserSetupService`) funnel through `OrganizationSettingsService.create`, which seeds the Daily Trends Digest workflow in a `try/catch`. The catch swallowed any failure with only a `logger.error` — so a DI / module-graph regression (the very class of bug this code works around: the org-settings↔workflows TDZ cycle) could break workflow seeding for **every new org silently**, with nothing but a log line.

## Fix

Report the swallowed error to **Sentry** (`@sentry/nestjs`, the pattern already used in the API's exception filters) in addition to logging. Org creation still never fails on this non-critical step.

## Tests

The service's first spec:
- provisioning failure → `logger.error` **and** `Sentry.captureException` called; the method still resolves (never throws).
- both early-return guards (no org id; no owner `userId`) skip without alerting.

## Verification

- `vitest`: org-settings service spec **3 passed**.
- `tsc --noEmit` (api): **0 errors**.
- Biome + secretlint: clean.

---

_From the automated daily commit review (2026-06-21). Final PR in the set. Related: #718, #720, #721, #723, #724, #725, #726, #727._

### DIFF
--- a/apps/server/api/src/collections/organization-settings/services/organization-settings.service.spec.ts
+++ b/apps/server/api/src/collections/organization-settings/services/organization-settings.service.spec.ts
@@ -1,0 +1,68 @@
+import { OrganizationSettingsService } from '@api/collections/organization-settings/services/organization-settings.service';
+import type { LoggerService } from '@libs/logger/logger.service';
+import type { ModuleRef } from '@nestjs/core';
+import * as Sentry from '@sentry/nestjs';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('@sentry/nestjs', () => ({
+  captureException: vi.fn(),
+}));
+
+describe('OrganizationSettingsService.provisionDefaultWorkflows', () => {
+  const organization = { findFirst: vi.fn() };
+  const prisma = { organization, organizationSetting: {} };
+  const logger: Partial<LoggerService> = {
+    debug: vi.fn(),
+    error: vi.fn(),
+    log: vi.fn(),
+    warn: vi.fn(),
+  };
+  const moduleRef = { get: vi.fn() };
+
+  let service: OrganizationSettingsService;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    service = new OrganizationSettingsService(
+      prisma as never,
+      logger as LoggerService,
+      moduleRef as unknown as ModuleRef,
+    );
+  });
+
+  // Private side-effect method; exercised directly to cover the bootstrap path.
+  const provision = (settings: unknown) =>
+    (
+      service as unknown as {
+        provisionDefaultWorkflows: (s: unknown) => Promise<void>;
+      }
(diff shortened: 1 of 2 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**PR #728 — `fix(api): surface provisionDefaultWorkflows failures to Sentry`**

Scope: one production line (`Sentry.captureException(error)` added to an intentionally-swallowed catch in `OrganizationSettingsService.provisionDefaultWorkflows`) plus a new spec covering the bootstrap path. Findings confirmed by a 3-skeptic adversarial pass.

The production change is clean and correct: keeping the swallow so non-critical workflow seeding never aborts org creation, while reporting to Sentry so the failure stops being a log line nobody watches. The explanatory comment earns its place. All findings below are on the new test file and are minor.

### Structural findings

None.

### Simplification / hygiene

**1. The Sentry assertion checks the count but not the payload.** — `organization-settings.service.spec.ts:55`
`expect(Sentry.captureException).toHaveBeenCalledTimes(1)` would still pass if the implementation regressed to `captureException(null)`/`undefined`. For the one test whose entire point is "the error is forwarded," that's the assertion that matters.
*Fix:* `expect(Sentry.captureException).toHaveBeenCalledWith(expect.objectContaining({ message: 'db down' }))` (or at least `expect.any(Error)`).

**2. The private reach-through erases the real parameter type with `unknown`.** — `organization-settings.service.spec.ts:40-45`
The `provision` helper casts to `(s: unknown) => Promise<void>`, while the real method takes `OrganizationSettingDocument`. Since this is the sole spec exercising the method, it's also the only type-level guard on its contract — and `unknown` lets the signature drift (added param, renamed type) without a compile error.
*Fix:* type the cast as `(s: OrganizationSettingDocument) => Promise<void>` and cast the partial fixtures at each call site (`{ organizationId: 'org-1' } as unknown as OrganizationSettingDocument`).

**3. Three unused logger stubs.** — `organization-settings.service.spec.ts:22-24`
The method only calls `logger.error`; `debug`, `log`, and `warn` are stubbed but never invoked or asserted. `Partial<LoggerService>` is satisfied with `error` alone.
*Fix:* keep only `error: vi.fn()`.

**4. Duplicated `'org-1'` literal.** — `organization-settings.service.spec.ts:51,61`
Same string in two independent `it` blocks.
*Fix:* hoist a `const ORG_ID = 'org-1'` alongside the other fixtures.

### Verdict: **APPROVE**

A correct, well-scoped observability fix. The only assertion worth tightening before merge is finding 1 (verify the captured argument); 2–4 are housekeeping. Nothing blocks merge.

PR #729, opened 2026-06-21
test(api): cover darkroom asset/publishing/training-orchestrator guards

VincentShipsIt:
## Summary

Adds unit-test coverage for the Darkroom peer services from the [#690](https://github.com/genfeedai/genfeed.ai/commit/a03e671a0) (`a03e671a0`) god-object decomposition. That refactor shipped 7 behaviour-bearing services with **no specs** — only `DarkroomCharacterService`, `DarkroomTrainingService`, and the value reader were covered, so a regression in any moved method would go uncaught locally.

This is the **first tranche**, covering the three highest-risk guard paths the review flagged:

- **`DarkroomAssetService.reviewAsset`** — multi-tenant `NotFound` guard; a freshly-approved IMAGE syncs into the training dataset; non-IMAGE and rejection do **not** sync.
- **`DarkroomPublishingService.publishAsset`** — `NotFound`; not-approved and missing-CDN `BadRequest` guards; Instagram dispatch success; an unsupported platform fails only that platform (not the whole call).
- **`DarkroomTrainingOrchestratorService`** — `getTraining` `NotFound`; **empty-dataset guard** refuses to start and creates no training record.

The remaining peer services (`generation`, `ingest`, `media`, `infra`, `stats`) are lower-risk and can follow in a second tranche.

## Verification

- `vitest`: **11 new tests** pass (publishing 5, asset 4, orchestrator 2).
- `tsc --noEmit` (api): **0 errors**.
- Biome + secretlint: clean.

This PR is tests-only — no production code changes.

---

_From the automated daily commit review (2026-06-21). Final PR in the set. Related: #718, #720, #721, #723, #724, #725, #726, #727, #728._

### DIFF
--- a/apps/server/api/src/endpoints/admin/darkroom/services/darkroom-asset.service.spec.ts
+++ b/apps/server/api/src/endpoints/admin/darkroom/services/darkroom-asset.service.spec.ts
@@ -1,0 +1,112 @@
+import { IngredientsService } from '@api/collections/ingredients/services/ingredients.service';
+import { ConfigService } from '@api/config/config.service';
+import { DarkroomAssetService } from '@api/endpoints/admin/darkroom/services/darkroom-asset.service';
+import { DarkroomTrainingService } from '@api/endpoints/admin/darkroom/services/darkroom-training.service';
+import { FilesClientService } from '@api/services/files-microservice/client/files-client.service';
+import { DarkroomReviewStatus, IngredientCategory } from '@genfeedai/enums';
+import { LoggerService } from '@libs/logger/logger.service';
+import { NotFoundException } from '@nestjs/common';
+import { Test, type TestingModule } from '@nestjs/testing';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+describe('DarkroomAssetService.reviewAsset', () => {
+  let service: DarkroomAssetService;
+  let ingredientsService: Record<string, ReturnType<typeof vi.fn>>;
+  let filesClientService: Record<string, ReturnType<typeof vi.fn>>;
+  let darkroomTrainingService: Record<string, ReturnType<typeof vi.fn>>;
+
+  const imageAsset = {
+    _id: { toString: () => 'asset-1' },
+    category: IngredientCategory.IMAGE,
+    cdnUrl: 'https://cdn/asset-1.jpg',
+    generationPrompt: 'a portrait',
+    personaSlug: 'alice',
+    reviewStatus: DarkroomReviewStatus.PENDING,
+  };
+
+  beforeEach(async () => {
+    ingredientsService = {
+      findAllByOrganization: vi.fn(),
+      findOne: vi.fn(),
+      patch: vi.fn().mockResolvedValue({ ...imageAsset }),
+    };
+    filesClientService = { uploadToS3: vi.fn().mockResolvedValue(undefined) };
+    darkroomTrainingService = {
+      syncDataset: vi.fn().mockResolvedValue(undefined),
+    };
+
+    const module: TestingModule = await Test.createTestingModule({
+      providers: [
+        DarkroomAssetService,
+        { provide: IngredientsService, useValue: ingredientsService },
+        { provide: FilesClientService, useValue: filesClientService },
+        {
+          provide: DarkroomTrainingService,
+          useValue: darkroomTrainingService,
+        },
+        { provide: ConfigService, useValue: { get: () => 'darkroom-bucket' } },
+        { provide: LoggerService, useValue: { log: vi.fn() } },
+      ],
+    }).compile();
+
+    service = module.get(DarkroomAssetService);
+  });
+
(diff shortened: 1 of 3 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**PR #729 — `test(api): cover darkroom asset/publishing/training-orchestrator guards`**

Scope: three new vitest spec files, no production code touched. The review therefore judges *test quality* — whether these tests actually pin the invariants they claim to, and whether the mocks faithfully model the real service contracts. Findings were produced by parallel dimension finders and each was independently confirmed by a 3-skeptic adversarial pass.

### Structural findings

**1. The primary success path (`startTraining` happy path) is untested, and the mock that a happy-path test would need is silently broken.** — `darkroom-training-orchestrator.service.spec.ts:265,298`
The only `startTraining` test trips the `imageCount === 0` guard (`BadRequestException`) before the service ever reaches `autoTuneHyperparameters`. That method is mocked as a bare `vi.fn()` returning `undefined`, while the real `DarkroomTrainingService.autoTuneHyperparameters` returns a `TrainingHyperparams` object the service immediately destructures (`autoTuned.steps`, `.rank`, `.learningRate`). So the moment anyone adds a happy-path test it crashes with `Cannot read properties of undefined (reading 'steps')` — and right now there is no assertion that a training record is created with the right hyperparameters, that `executeTrainingQuery` is fired-and-forgotten, or that `personasService.patch` sets `LoraStatus.TRAINING`.
*Fix:* give the mock a real return value — `autoTuneHyperparameters: vi.fn().mockReturnValue({ steps: 1000, rank: 16, learningRate: 1e-4, ... })` — and add a happy-path test (`getDatasetInfo` → `{ imageCount: 10 }`) asserting the create payload, the fire-and-forget call, and the persona status transition.

**2. The approved-IMAGE sync test under-asserts and would miss real regressions.** — `darkroom-asset.service.spec.ts:70-89`
On approval of an IMAGE asset the service calls `ingredientsService.patch` **twice** (reviewStatus, then `generationCompletedAt` inside `syncApprovedAssetToDataset`) and `filesClientService.uploadToS3` **twice** (image + caption `.txt`, since the fixture has `generationPrompt` set). The test asserts only `uploadToS3` `toHaveBeenCalled()` and the first `patch` shape — so dropping the caption upload, or the `generationCompletedAt` write, sails through green.
*Fix:* assert `uploadToS3` `toHaveBeenCalledTimes(2)` and add `patch` `toHaveBeenNthCalledWith(2, 'asset-1', expect.objectContaining({ generationCompletedAt: expect.any(Date) }))`.

**3. The rejection test verifies an absence but not the write that must happen.** — `darkroom-asset.service.spec.ts:107-117`
`does not sync on rejection` only checks that `uploadToS3` was *not* called. The service unconditionally calls `ingredientsService.patch({ reviewStatus: REJECTED })` before the sync branch; if that write were accidentally moved inside the APPROVED guard, this test still passes.
*Fix:* add `expect(ingredientsService.patch).toHaveBeenCalledWith('asset-1', { reviewStatus: DarkroomReviewStatus.REJECTED })`.

### Simplification / hygiene

**4. `twitterService` mock is dead weight and silently wrong.** — `darkroom-publishing.service.spec.ts:17,30,37,154`
It is declared at describe scope, re-initialised in every `beforeEach`, and injected — but no `expect()` ever reads it. It also omits a return value while the real `TwitterService.uploadMedia` returns `Promise<string>` (awaited by the service as `platformId`), so it silently contradicts the Instagram mock's correct `mockResolvedValue({ mediaId: 'ig-1' })`. Since the variable is typed `Record<string, ReturnType<typeof vi.fn>>`, TypeScript won't catch the drift.
*Fix:* drop the captured variable and inline the provider exactly like `FacebookService`/`CredentialsService` (`{ provide: TwitterService, useValue: { uploadMedia: vi.fn().mockResolvedValue('tw-1') } }`).

**5. Two guard branches the PR set out to cover are left uncovered.** — `darkroom-asset.service.spec.ts:70-89`, `darkroom-publishing.service.spec.ts:201-220`
(a) The idempotency guard — re-approving an already-`APPROVED` asset must *not* re-upload or re-sync — has no test; the fixture is always `PENDING`. (b) The per-platform error-swallow path for a *supported* platform is untested: the `myspace` case exercises the switch default, not the `try/catch` that turns an Instagram failure into `{ success: false, error }`.
*Fix:* add (a) an `APPROVED → APPROVED` test asserting `uploadToS3`/`syncDataset` were not called, and (b) an `instagramService.uploadImage.mockRejectedValue(...)` test asserting `result.success === false` and the error is captured.

### Verdict: **APPROVE**

No structural regression, no file near the 1,000-line limit, and the tests that exist are correct. But this is a *coverage* PR, and its assertions are consistently weaker than the behaviour they touch — several would stay green through real regressions, and the main `startTraining` success path plus the idempotency guard are absent. Strongly recommend tightening the assertions in findings 1–3 and closing the gaps in 4–5 before this is treated as the darkroom services' safety net. None of these block merge.

PR #821, opened 2026-06-24
fix(ci): enable API E2E gate

### LINKED ISSUE #702: Re-enable the API E2E nightly job (E2ETestModule DI + subscriptions service/alias)
## Context

The `e2e-api` job in [`.github/workflows/e2e.yml`](.github/workflows/e2e.yml) is disabled behind the `E2E_API_ENABLED` repo variable (`if: ${{ vars.E2E_API_ENABLED == 'true' }}`, also `continue-on-error: true`). It was gated off because the API E2E suite fails every run on test-infra wiring, not on real regressions. Surfaced by the merged-PR follow-up audit of #684.

## Blockers (verified)

1. **`E2ETestModule` DI wiring** — [`apps/server/api/test/e2e-test.module.ts`](apps/server/api/test/e2e-test.module.ts) does not provide `PrismaService` and `FilesClientService`, so the suite throws on module compile.
2. **Missing subscriptions service/alias** — `@api/collections/subscriptions/services/subscriptions.service` is imported (e.g. `test/integration/payment-processing.integration.spec.ts`) but no `services/` directory exists under `apps/server/api/src/collections/subscriptions/` (only `controllers/`, `dto/`, `subscriptions.module.ts`). Either the service was moved/renamed or the import path is stale.

## Scope

- Provide `PrismaService` + `FilesClientService` (and any other missing providers) in `E2ETestModule`.
- Resolve the `subscriptions/services/subscriptions.service` import — point it at the real location or restore the missing module.
- Get the suite green against the workflow's `postgres:17-alpine` + `redis:7` service containers (run prisma migrate in setup as needed).
- Once green, set `E2E_API_ENABLED=true` and drop `continue-on-error`.

## Acceptance

- `e2e-api` runs and passes in nightly CI without `continue-on-error`.
- No stale `@api/collections/subscriptions/services/*` import paths remain.

Not covered by #623 (frontend Playwright reliability) or #439 (e2e coverage measurement).

VincentShipsIt:
## Summary
- enable the API E2E job as a required gate in `.github/workflows/e2e.yml`
- provide `PrismaService` and `FilesClientService` through the API E2E module without test-only service shims
- move the payment integration spec to the canonical `SUBSCRIPTIONS_SERVICE` token and mirror required workspace aliases in the E2E Vitest config

## Validation
- `bun install --frozen-lockfile --ignore-scripts`
- `npx biome check --write .github/workflows/e2e.yml apps/server/api/test/e2e-test.module.ts apps/server/api/test/integration/payment-processing.integration.spec.ts apps/server/api/test/mocks/external-services.mocks.ts apps/server/api/vitest.config.e2e.ts`
- `bunx prisma generate` for local ignored Prisma client generation
- local Postgres `test` DB created and migrated with `bunx prisma migrate deploy`
- `bunx vitest run --config vitest.config.e2e.ts test/integration/payment-processing.integration.spec.ts` (17 tests)
- `bun run --cwd apps/server/api test:e2e:ci` (3 files, 34 tests)
- `bunx turbo run type-check --filter=@genfeedai/api`
- `actionlint .github/workflows/e2e.yml`
- `git diff --cached --check` and stale import / E2E_API guard scan
- staged filename/diff secret scans and pre-commit secretlint/Biome hooks

Closes #702

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -39,19 +39,14 @@ jobs:
       - name: Check route coverage
         run: node scripts/e2e-route-coverage.mjs
 
-  # API E2E Tests — DISABLED by default. The test infrastructure is incomplete:
-  # the suite fails every run on E2ETestModule DI wiring (PrismaService/
-  # FilesClientService not provided) and a missing @api/collections/subscriptions
-  # path alias. It was already continue-on-error (never gated), so it only burned
-  # ~2 min/run and added noise. Gated behind the E2E_API_ENABLED repo variable
-  # (mirrors e2e-frontend-authed): set it to 'true' to re-enable once the DI /
-  # path-alias fixes land.
+  # API E2E Tests — boot/smoke gate for the real API test graph.
+  # The #702 blockers are resolved in-source: E2ETestModule provides the
+  # files client mock, API E2E resolves the OSS billing provider alias, and the
+  # payment integration spec uses the canonical subscription DI token.
   e2e-api:
     name: API E2E Tests
-    if: ${{ vars.E2E_API_ENABLED == 'true' }}
     runs-on: ubuntu-latest
     timeout-minutes: 20
-    continue-on-error: true
     services:
       postgres:
         image: postgres:17-alpine
@@ -203,19 +198,24 @@ jobs:
     name: E2E Gate (all shards)
     runs-on: ubuntu-latest
     timeout-minutes: 5
-    needs: [e2e-route-coverage, e2e-frontend]
+    needs: [e2e-route-coverage, e2e-frontend, e2e-api]
     if: always()
     steps:
       - name: Check shard results
         run: |
           echo "Route coverage: ${{ needs.e2e-route-coverage.result }}"
           echo "Frontend matrix: ${{ needs.e2e-frontend.result }}"
+          echo "API E2E: ${{ needs.e2e-api.result }}"
           if [[ "${{ needs.e2e-route-coverage.result }}" == "failure" ]]; then
             echo "Route coverage gate failed" && exit 1
           fi
           if [[ "${{ needs.e2e-frontend.result }}" == "failure" || \
                 "${{ needs.e2e-frontend.result }}" == "cancelled" ]]; then
             echo "One or more E2E shards failed or were cancelled" && exit 1
+          fi
+          if [[ "${{ needs.e2e-api.result }}" == "failure" || \
+                "${{ needs.e2e-api.result }}" == "cancelled" ]]; then
+            echo "API E2E boot gate failed or was cancelled" && exit 1
           fi
           echo "All required E2E checks passed."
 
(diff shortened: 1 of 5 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**Verdict:** APPROVE — no structural blockers. The CI gate enablement is safe to merge.

### Structural findings

- **[MEDIUM] `apps/server/api/test/integration/payment-processing.integration.spec.ts:143-150, 176`** — `PaymentSubscriptionsServiceMock` declares `cancel`, `create`, `findByCustomer`, `update`, `updateStatus` and is registered under the `SUBSCRIPTIONS_SERVICE` DI token, but the canonical `ISubscriptionsService` contract (at `packages/interfaces/src/billing/subscriptions-service.contract.ts`) exposes none of those names — it exposes `findOne`, `findByOrganizationId`, `findAll`, `patch`, `findByStripeCustomerId`, `syncWithStripe`, `createForOrganization`, `syncSubscriptionState`. Only `findOne` overlaps. The practical blast radius is contained: `StripeService` itself does not inject `SUBSCRIPTIONS_SERVICE`, the `cancel`/`create` stubs at lines 324/355 are never asserted, and the `updateStatus` assertion at line 713 is tautological (a `mockImplementation` calling another mock). However, the token change from `SubscriptionsService` → `SUBSCRIPTIONS_SERVICE` was the right direction and should have been paired with aligning the mock surface to the contract. As-is, the spec provides zero type-level protection against `ISubscriptionsService` drift on the paths that actually matter (`patch`, `findByStripeCustomerId`, `syncWithStripe`, `syncSubscriptionState`). → Replace `PaymentSubscriptionsServiceMock` with `vi.Mocked<ISubscriptionsService>`, update the `useValue` mock to expose `{ findOne, patch, findByStripeCustomerId, syncWithStripe, syncSubscriptionState }` as `vi.fn()`, and update the retrieval cast at lines 187-189 to `moduleRef.get<vi.Mocked<ISubscriptionsService>>(SUBSCRIPTIONS_SERVICE)`. Drop or re-scope the dead `cancel`/`create` stub setup and the tautological `updateStatus` assertion.

- **[LOW] `apps/server/api/test/integration/payment-processing.integration.spec.ts:143-151`** — The local `PaymentSubscriptionsServiceMock` type is a named alias for a shape that could simply be expressed as `Partial<ISubscriptionsService>` with the genuinely-exercised methods. This makes the type self-documenting and structurally bound to the contract so future interface changes fail at compile time in this file rather than silently compiling against a stale shadow type. The `findOne` method is the only overlapping name between the current mock type and the real contract, which means five of the six declared methods are dead declarations relative to the interface. This is a hygiene cleanup with no behavioral consequence today but meaningful maintenance value going forward.

### Notes

- All three findings collapse to the same underlying issue — the `PaymentSubscriptionsServiceMock` type is off-contract — reported at slightly different granularities. The MEDIUM is the authoritative framing; the two LOW entries are the same observation with narrower scope.
- The pre-existing divergence in mock method names was present before this PR; the PR introduced the named type alias that formalizes (and makes visible) the mismatch. The token switch from `SubscriptionsService` → `SUBSCRIPTIONS_SERVICE` is the correct fix and should be kept.
- This was a read-only diff review. No tests, builds, or typechecks were executed. CI results should be the final gate.

PR #822, opened 2026-06-25
fix(api): address Sentry production feed errors

VincentShipsIt:
## Sentry coverage

Covers these unresolved production issues:

- API-GENFEED-AI-65, GET /v1/public/articles/slug/*: Prisma selected users.authProviderId, but the live DB can still have the pre-rename clerkId column. Added a conditional Prisma migration to restore authProvider* column/index names.
- API-GENFEED-AI-62, GET /v1/public/posts: public posts copied an asset scope filter onto Post queries; Post has no scope field. Removed the invalid filter.
- API-GENFEED-AI-61 and API-GENFEED-AI-5X, GET /v1/images and GET /v1/musics: malformed scope query objects could reach Prisma as scope: { not: null }. Scope filters now only accept real AssetScope string values.
- API-GENFEED-AI-63, POST /v1/skills-pro/checkout: placeholder Stripe secrets reached the Stripe API. Checkout now returns a service-unavailable response before creating a Stripe session when the secret is missing or placeholder-shaped.

Also found existing master fixes for API-GENFEED-AI-64 and GENFEED-AI-43; those are not changed in this PR.

## Validation

- bunx biome check --write apps/server/api/src/endpoints/public/controllers/posts/public.posts.controller.ts apps/server/api/src/endpoints/public/controllers/posts/public.posts.controller.spec.ts apps/server/api/src/helpers/utils/collection-filter/collection-filter.util.ts apps/server/api/src/helpers/utils/collection-filter/collection-filter.util.spec.ts apps/server/api/src/skills-pro/services/skill-checkout.service.ts apps/server/api/src/skills-pro/services/skill-checkout.service.spec.ts
- bun run --cwd apps/server/api test:file src/endpoints/public/controllers/posts/public.posts.controller.spec.ts src/helpers/utils/collection-filter/collection-filter.util.spec.ts src/skills-pro/services/skill-checkout.service.spec.ts
- bunx turbo run type-check --filter=@genfeedai/api
- git diff --cached --check plus staged credential-pattern scan

No production deploys or production migrations were run.

### DIFF
--- a/apps/server/api/src/endpoints/public/controllers/posts/public.posts.controller.spec.ts
+++ b/apps/server/api/src/endpoints/public/controllers/posts/public.posts.controller.spec.ts
@@ -5,7 +5,7 @@ import { PublicPostsController } from '@api/endpoints/public/controllers/posts/p
 import { BaseQueryDto } from '@api/helpers/dto/base-query.dto';
 import { RolesGuard } from '@api/helpers/guards/roles/roles.guard';
 import type { AggregatePaginateResult } from '@api/types/aggregate-paginate-result';
-import { AssetScope, PostStatus } from '@genfeedai/enums';
+import { PostStatus } from '@genfeedai/enums';
 import { LoggerService } from '@libs/logger/logger.service';
 import { Test, TestingModule } from '@nestjs/testing';
 import type { Request } from 'express';
@@ -181,9 +181,9 @@ describe('PublicPostsController', () => {
       };
       expect(callArgs.where).toEqual({
         isDeleted: false,
-        scope: AssetScope.PUBLIC,
         status: PostStatus.PUBLIC,
       });
+      expect(callArgs.where.scope).toBeUndefined();
     });
 
     it('should handle invalid account id gracefully', async () => {
--- a/apps/server/api/src/endpoints/public/controllers/posts/public.posts.controller.ts
+++ b/apps/server/api/src/endpoints/public/controllers/posts/public.posts.controller.ts
@@ -64,7 +64,6 @@ export class PublicPostsController {
 
     const matchQuery: PrismaWhereQuery = {
       isDeleted: false,
-      scope: AssetScope.PUBLIC,
       status: PostStatus.PUBLIC,
     };
 
--- a/apps/server/api/src/helpers/utils/collection-filter/collection-filter.util.spec.ts
+++ b/apps/server/api/src/helpers/utils/collection-filter/collection-filter.util.spec.ts
@@ -55,6 +55,16 @@ describe('CollectionFilterUtil', () => {
       const result = CollectionFilterUtil.buildScopeFilter(undefined);
       expect(result).toBeUndefined();
     });
+
+    it('omits malformed scope filter objects', () => {
+      const result = CollectionFilterUtil.buildScopeFilter({ not: null });
+      expect(result).toBeUndefined();
+    });
+
+    it('omits unknown scope strings', () => {
+      const result = CollectionFilterUtil.buildScopeFilter('unknown');
+      expect(result).toBeUndefined();
+    });
   });
 
   describe('buildSearchFilter', () => {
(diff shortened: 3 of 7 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**Verdict: BLOCK** — one confirmed blocker in the migration layer that will cause runtime auth failures on deployment.

---

### Structural findings

- **[BLOCKER] `packages/prisma/prisma/migrations/20260625085045_restore_auth_provider_column_names/migration.sql:215-217`** — The migration comment asserts "Current Prisma schema expects authProvider* column names," but this is factually wrong. `schema.prisma` lines 490, 598, and 874 define `clerkId`, `clerkOrganizationId`, and `clerkMembershipId` respectively, with no `@map` directives. Prisma derives DB column names directly from field names when `@map` is absent. On any database that currently carries the `clerk*` column names (i.e., all production and dev DBs that applied migrations up through this PR's base), the conditional `IF EXISTS (clerkId) AND NOT EXISTS (authProviderId)` branch will execute and rename those columns. After the rename, the Prisma client continues to issue queries against `clerkId`, `clerkOrganizationId`, and `clerkMembershipId` — columns that no longer exist. Every user authentication lookup, organization resolution, and member query will fail at runtime with a Prisma column-not-found error or silent null return. This PR ships neither a `schema.prisma` field rename nor `@map` directives to bridge the gap. The two halves of this rename must land atomically. → Either (a) update `schema.prisma` to rename the three fields (adding `@map("clerkId")` etc. if backward-compat aliases are needed) in this same PR, or (b) invert the migration direction and keep the DB columns as `clerk*` until the schema rename is ready. Do not merge with the schema and migration out of sync.

---

- **[LOW] `apps/server/api/src/helpers/utils/collection-filter/collection-filter.util.ts` — `static buildScopeFilter(scope?: unknown)`** — The parameter was previously `scope?: AssetScope`, which statically rejected non-enum values at every call-site. All four active callers (`images.controller.ts`, `gifs.controller.ts`, `videos.controller.ts`, `musics.controller.ts`) pass `query.scope` from their respective DTOs, and every one of those DTOs declares `scope?: AssetScope` with `@IsEnum(AssetScope)` validation. The root Sentry error was in `public.posts.controller.ts` and was correctly fixed by removing the erroneous `scope` field from that controller's query — `buildScopeFilter` was never called there at all. Widening the shared utility to `unknown` was not required to address the Sentry errors, trades a compile-time guarantee for a runtime guard, and forces two `as AssetScope` casts that are only necessary because the parameter is now `unknown`. There is no behavioral regression (the guard is correct), but the compile-time contract is now weaker for all callers. → Revert the signature to `scope?: AssetScope`. If a future caller needs to handle raw HTTP strings, introduce a separate `parseScope(raw: string): AssetScope | undefined` helper rather than degrading the shared utility.

- **[LOW] `apps/server/api/src/skills-pro/services/skill-checkout.service.ts` — `if (!this.stripeService.stripe || ...)`** — `StripeService.stripe` is declared `public readonly stripe: StripeClient` (non-nullable), but the self-hosted code path in `stripe.service.ts` assigns `null as unknown as StripeClient`. The new `assertStripeCheckoutConfigured` guard's `!this.stripeService.stripe` check depends entirely on this unsafe cast persisting. The guard functions correctly at runtime today. However, the repo's established pattern for self-hosted guarding is an explicit `IS_SELF_HOSTED` static check (confirmed in `managed-stripe-checkout.service.ts:41` and `webhooks.stripe.controller.ts:43`). The new guard departs from that pattern and relies on runtime truthiness against a non-nullable declared type — if `StripeService` is ever refactored to initialise `stripe` with a non-null sentinel on unconfigured paths, the guard fails silently without a type error. This is pre-existing debt made newly visible by the PR; the guard itself is a net improvement over having no guard. → As a follow-up (not a blocker here), change `StripeService.stripe` to `StripeClient | null`, remove the `null as unknown as StripeClient` cast, and let TypeScript enforce null checks at all call-sites.

---

### Notes

- This was a read-only diff review. No tests, builds, or type-checks were executed locally; runtime and CI behavior was inferred from the diff and the repository files at HEAD.
- The LOW findings are non-blocking cleanups; the BLOCKER is a hard stop. The migration must not reach production with `schema.prisma` in its current state.

PR #824, opened 2026-06-25
ci: move heavy gates to release cut

### LINKED ISSUE #744: ci: fast PRs, heavy gates at release-cut, self-hosted E2E nightly-only
Parent: Deployment-modes epic.

Rework CI so **PR checks are fast and all heavy gates move to the release-cut** before deploy. Four directives:

1. Remove the ~20-min `api-boot-check` from the PR gate (`ci.yml`); make it the first step of the release-cut `build-verify.yml`.
2. Keep the full API integration E2E off PRs (release-cut only).
3. Self-hosted release E2E (`e2e-selfhosted-release.yml`) → **nightly cron only**; drop the post-publish `workflow_run` trigger; it must never block a release.
4. **Split the community image build/publish from the SaaS deploy** — remove `selfhosted-build-verify` from the E2E gate's `needs` in `full-suite.yml`.

Result: PR gate ≈ format/lint/typecheck/guards/build (`--affected`); release-cut carries boot-check → image → E2E before `deploy-ecs`.

## Related
#732 (self-hosted E2E failing as a release blocker), #702 (API E2E re-enable).

## Done when
PRs are fast; community flakiness never blocks a SaaS release; gates run at release-cut.

VincentShipsIt:
## Summary
- remove the compiled API boot check from PR CI and run it first in release `build-verify.yml`
- make self-hosted release E2E nightly/manual only by dropping the Docker Publish `workflow_run` trigger
- decouple self-hosted build verification from the SaaS full-suite E2E gate

## Validation
- `actionlint .github/workflows/ci.yml .github/workflows/build-verify.yml .github/workflows/e2e-selfhosted-release.yml .github/workflows/full-suite.yml`
- static workflow contract assertions for #744
- `git diff --check`
- staged filename/diff secret scans
- `./node_modules/.bin/secretlint .github/workflows/ci.yml .github/workflows/build-verify.yml .github/workflows/e2e-selfhosted-release.yml .github/workflows/full-suite.yml`
- pre-commit `bun scripts/run-secretlint.ts --no-glob`

Closes #744
Related #732

### DIFF
--- a/.github/workflows/build-verify.yml
+++ b/.github/workflows/build-verify.yml
@@ -20,10 +20,35 @@ env:
   IMAGE_NAME: genfeed-server-verify
 
 jobs:
+  # Release-cut boot gate — builds the COMPILED api bundle and loads its full
+  # provider graph with BOOT_CHECK=1, which exits 0 BEFORE NestFactory/config
+  # (no env, DB, or Redis needed). Kept out of normal PR CI so pull requests stay
+  # fast, but still runs before Docker release verification.
+  api-boot-check:
+    name: API Boot Check
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v5
+      - name: Setup Bun environment
+        uses: ./.github/actions/setup-bun-env
+        with:
+          bun-version: '1.3.14'
+      - name: Build api (compiled bundle)
+        # The api package emits to apps/server/dist, outside apps/server/api.
+        # Force a real build so a Turbo cache-hit replay cannot skip restoring
+        # the compiled bundle this boot check loads.
+        run: bunx turbo run build --filter=@genfeedai/api --force
+      - name: Boot check — compiled graph must load
+        run: BOOT_CHECK=1 timeout 120 node apps/server/dist/apps/api/main.js
+
   build-and-boot:
     name: Build & Boot Check
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    needs: api-boot-check
     permissions:
       contents: read
       packages: write
@@ -120,6 +145,8 @@ jobs:
       - name: Build summary
         if: success()
         run: |
-          echo "## ✅ Build Verification Passed" >> $GITHUB_STEP_SUMMARY
-          echo "" >> $GITHUB_STEP_SUMMARY
-          echo "Docker image built + all 12 bundles verified; EE api bundle booted past module init (#574 guard)." >> $GITHUB_STEP_SUMMARY
+          {
+            echo "## ✅ Build Verification Passed"
+            echo ""
+            echo "Docker image built + all 12 bundles verified; EE api bundle booted past module init (#574 guard)."
+          } >> "$GITHUB_STEP_SUMMARY"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -534,30 +534,3 @@ jobs:
           else
             bun run build
           fi
-
-  # Boot gate — builds the COMPILED api bundle and loads its full provider graph
-  # with BOOT_CHECK=1, which exits 0 BEFORE NestFactory/config (no env, DB, or
-  # Redis needed). A circular-import TDZ — e.g. #711's "Cannot access 'X' before
-  # initialization", which crashed the API on boot and shipped because nothing in
-  # PR CI booted the build — crashes the process here and fails the PR. vitest
-  # cannot reproduce this class (its SWC/ESM resolution differs); only the
-  # compiled webpack bundle does, which is exactly what this runs.
-  api-boot-check:
-    name: API Boot Check
-    runs-on: ubuntu-latest
-    timeout-minutes: 20
-    needs: trust
-    if: needs.trust.outputs.is-trusted == 'true'
-    steps:
-      - uses: actions/checkout@v5
-      - name: Setup Bun environment
-        uses: ./.github/actions/setup-bun-env
-        with:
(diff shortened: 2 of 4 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**Verdict: BLOCK** — one HIGH regression (api-boot-check removed from PR CI) warrants an explicit sign-off before merge, given the documented production history.

### Structural findings

- **[HIGH] `.github/workflows/ci.yml:74-99`** — `api-boot-check` is removed from PR CI entirely. The deleted comment documented the precise reason this job must live here: *"vitest cannot reproduce this class [of TDZ crash] (its SWC/ESM resolution differs); only the compiled webpack bundle does."* This is the exact failure mode that caused production incident #711 ("Cannot access X before initialization") — it passed lint, typecheck, and vitest, then crashed the API on boot. The job still exists in `build-verify.yml`, but that file is `workflow_dispatch` / `workflow_call` only; it runs only when someone manually dispatches the Full Suite or cuts a release. Every PR that merges between now and a release cut is unguarded. The release-cut catch is real, but it means a broken master build can persist across multiple merges. → Restore `api-boot-check` in `ci.yml` with the same `needs: trust` + `if: needs.trust.outputs.is-trusted == 'true'` guards as the other CI jobs. The `build-verify.yml` copy can remain as the Docker-boot variant for the release gate — the two purposes are distinct and both are warranted. If the PR-speed argument is the blocker, at minimum document explicit sign-off from Vincent that the #711 risk class is acceptable during the window between PR merge and release cut.

- **[MEDIUM] `.github/workflows/full-suite.yml:179-187`** — `selfhosted-build-verify` job removed from Full Suite and dropped from `needs: [ci, build-verify, selfhosted-build-verify]` → `needs: [ci, build-verify]`. The removed job's own comment read: *"Release-time self-hosted image validation. Kept out of normal PR CI so routine pull requests stay fast; run through Full Suite before release."* That was exactly the right design. After this PR, `build-verify-selfhosted.yml` has only `workflow_dispatch` and `workflow_call` triggers with zero active callers — the PR comment that it "runs independently" is aspirational, not structural. No cron, no `workflow_run`, no new caller exists. `Dockerfile.selfhosted` can now break (bad apt package, entrypoint drift, ee/ coupling regression) and ship in the community `:latest` image without any automated gate catching it pre-publish. → Either (a) add `selfhosted-build-verify` back to `full-suite.yml` as a non-blocking parallel job (no entry in `e2e` needs, just visibility), or (b) add a `schedule` trigger directly to `build-verify-selfhosted.yml` so it runs on its own cadence. Leaving `workflow_call` with no callers is dead-code structure.

- **[MEDIUM] `.github/workflows/e2e-selfhosted-release.yml:119-123` and `41-43`** — `workflow_run: workflows: ['Docker Publish (Self-Hosted)'] types: [completed]` trigger removed along with its `if: github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success'` guard. The old design ran release E2E immediately after a successful publish, proving the just-shipped container worked before the release was announced. The replacement is a nightly schedule at 05:23 UTC — meaning a broken image published at 14:00 UTC is live and undetected for up to ~15 hours. The `if` guard was also doing meaningful work: without it, a failed publish (GHCR auth error, etc.) would trigger E2E against stale `:latest` and produce a misleading green. The PR rationale ("community image flakiness never blocks a SaaS release") is valid as a decoupling goal, but removing the `workflow_run` trigger achieves that goal by eliminating the correctness gate entirely rather than by isolating it. → Restore `workflow_run` alongside the nightly schedule. These serve orthogonal purposes: `workflow_run` = per-publish correctness gate; `schedule` = drift/dependency regression canary. If flakiness is the concern, add `continue-on-error: true` or move the self-hosted E2E to a separate non-blocking status, rather than removing the trigger.

- **[MEDIUM] `.github/workflows/full-suite.yml:33-36` (duplicate angle on selfhosted removal)** — Note: findings #2 and #4 in the verified set both cover the same `selfhosted-build-verify` removal from `full-suite.yml`. The core gap is the same: community Dockerfile.selfhosted is now unguarded in any automated pre-release path. The suggestion in finding #4 is the tightest fix: add a `workflow_call` invocation of `build-verify-selfhosted.yml` from `docker-publish.yml` before the push step, gating publication on a successful build-verify. This closes the gap at the artifact's actual publication point rather than at the manual Full Suite dispatch.

- **[LOW] `.github/workflows/e2e-selfhosted-release.yml:127`** — Schedule changed from `23 5 */3 * *` (~10 runs/month) to `23 5 * * *` (~30 runs/month) with no documented rationale for the 3x frequency increase. Between releases, `:latest` is static — nightly runs against an unchanged image produce no incremental signal versus the tri-daily canary and run 20 extra jobs per month. If the goal was to improve drift detection cadence after removing `workflow_run`, weekly is a more defensible choice for a post-release canary. Document the intent or revert to a lower cadence.

### Notes

- This was a read-only diff review. No tests, builds, or typechecks were run.
- The HIGH finding (api-boot-check removal) is the merge blocker. The MEDIUM findings together represent a meaningful regression in the community self-hosted release gate, but each has a compensating control (nightly canary, manual dispatch, docker-publish's own build step) — they warrant fixes before the next release cut but could technically be addressed in a follow-up if Vincent explicitly accepts the risk window.
- The two MEDIUM selfhosted findings (#2 and #4) overlap substantially; they are both grounded but describe the same structural gap from slightly different angles. One consolidated fix (restoring `selfhosted-build-verify` as a parallel non-blocking job in `full-suite.yml`, or gating `docker-publish.yml` on `build-verify-selfhosted.yml`) would close both.

PR #847, opened 2026-06-26
refactor(autopilot): decompose executeOpportunity + typed ID accessors (#779)

### LINKED ISSUE #779: Refactor autopilot opportunity execution
_Filed by the automated code-quality sweep (read-only review). Findings below were verified against current source._

## Summary

`apps/server/api/src/collections/agent-strategies/services/agent-strategy-autopilot.service.ts` (1,431 lines) concentrates the autopilot execution flow in a single 286-line method, `executeOpportunity()` (lines 705–990), and leans on a repeated unsafe `as Record<string, unknown>` ID-extraction idiom to read document IDs. The casts defeat the type system (41 occurrences) and violate the repo's strict-types rule, while the giant method makes the per-branch publish/revise/handoff logic untestable.

## Confirmed evidence

`apps/server/api/src/collections/agent-strategies/services/agent-strategy-autopilot.service.ts` (1,431 lines)

- `private async executeOpportunity(...)` spans **lines 705–990 (286 lines)**; the next method `evaluateDraft` begins at line 992. It interleaves status transitions, generation, a draft-evaluation gate, a `revise` re-evaluation sub-path, an auto-publish path, and a publishing-inbox handoff path under deep nesting.
- The unsafe ID idiom `((opportunity as Record<string, unknown>).id as string) ?? String(opportunity._id)` (and the `draft` equivalent) is repeated throughout, confirmed at lines 719/720, 729/730, 763/764, 788/789, 817/818, 844/845, 865/866, 889/890, and more.
- `opportunitiesService.updateStatus(...)` is called **8×**, each with that repeated ID-extraction expression as its first argument.
- The file contains **41** `as Record<string, unknown>` casts and **8** `as never` casts, both violating the repo strict-types rule (no `any`, no inline casting).
- The file already has the right pattern to follow: typed accessors `strategyId()` (~398), `strategyOrganizationId()` (~402), `strategyBrandId()` (~410), `asOptionalString()` (~452).

The ID idiom is copy-pasted ~18 times; if document-ID normalization changes (`id` vs `_id`), every site must change and a missed one silently passes a wrong ID into `updateStatus`/approve/reject. The casts blind the compiler to wrong field names. At 286 lines with four nesting levels, `executeOpportunity` cannot be tested per-branch, and the approve/handoff logic is duplicated between the auto-publish-fallback path and the non-publish path.

1. Add private typed accessors `opportunityId(o)`, `draftId(d)`, `strategyDocId(s)` returning `string`, following the existing `strategyId()`/`asOptionalString()` pattern. Replace every inline ID extraction with these.
2. Decompose `executeOpportunity` by extracting `handleVideoHold(...)`, `reviseAndReEvaluate(...)` (the `gate.decision === 'revise'` block), `handleGateRejection(...)`, and a single parameterized `finalizeApprovalAndHandoff(...)` shared by the auto-publish and non-publish paths. Leave `executeOpportunity` as an orchestrator.
3. Consolidate brand/org/id resolution in `getPerformanceSnapshot` / `generateStrategyReport` onto the existing accessors rather than re-deriving inline.

## Acceptance criteria

- `executeOpportunity()` <= 100 lines; no extracted helper exceeds 80 lines.
- The `((x as Record<string, unknown>).id as string) ?? String(x._id)` idiom appears 0 times inline (replaced by one typed accessor per entity).
- `as Record<string, unknown>` count materially reduced (target <= 10, down from 41); no new `any`.
- Identical autopilot behavior: same status transitions (`generating`/`held`/`discarded`/`approved`/`published`), same gate decisions, same handoff batch creation; existing tests pass.

VincentShipsIt:
## What

Resolves the #779 code-quality finding: `AgentStrategyAutopilotService.executeOpportunity()` was a single 286-line method, and the file leaned on an unsafe `((x as Record<string,unknown>).id as string) ?? String(x._id)` ID idiom (41 `as Record<string,unknown>` casts) that defeated the type system.

## Changes

**Decomposition** — `executeOpportunity()` is now a 97-line orchestrator delegating to focused helpers, each ≤80 lines:
- `generateAutopilotDraft` — provider request + draft extraction
- `persistAutopilotMetadata` — build + patch autopilot metadata
- `handleVideoHold` — video draft-only hold
- `reviseAndReEvaluate` — the `revise` optimize→re-evaluate path (discriminated-union return: terminal discard vs fall-through)
- `handleGateRejection` — discard/hold/non-approved gate handling
- `finalizeApprovalAndHandoff` — thin dispatcher → `finalizeAutoPublish` / `finalizeManualReviewHandoff`

**Type safety** — added typed ID accessors `documentId`/`opportunityId`/`draftId` (and `strategyId` now delegates to `documentId`); replaced every inline ID-extraction site. `as Record<string,unknown>` casts: **41 → 4**; no new `any`.

**Brand/org consolidation** — `getPerformanceSnapshot`, `generateStrategyReport`, and `publishTextDraft` now use the existing `strategyBrandId`/`strategyOrganizationId` accessors. Missing-brand query semantics are preserved with `?? String(strategy.brand)` so a brandless strategy still filters by a non-matching value instead of dropping the brand filter (which would have broadened a credential lookup across brands — caught in adversarial review and fixed before merge).

## Behavior

Behavior-preserving: identical status transitions (`generating`/`held`/`discarded`/`approved`/`published`), gate decisions, and publishing-inbox handoff batch creation. The successful-`revise` fall-through still uses the original gate object and the original (non-optimized) draft content, matching the prior implementation exactly.

## Acceptance criteria

- [x] `executeOpportunity()` ≤ 100 lines (now 97); no extracted helper > 80 lines
- [x] Inline `((x as Record<string,unknown>).id as string) ?? String(x._id)` idiom appears 0 times
- [x] `as Record<string,unknown>` ≤ 10 (now 4, down from 41); no new `any`
- [x] Identical autopilot behavior; existing spec passes

## Verification

- `vitest run agent-strategy-autopilot.service.spec.ts` → **6/6 pass** (run on Mac Studio)
- `tsc --noEmit` (api package) → **no errors in the changed file** (only a pre-existing `tsconfig baseUrl` deprecation, unrelated to this diff)
- `biome check` → clean; secretlint (pre-commit) → clean
- Adversarial multi-agent diff review (3 lenses → verify): 2 confirmed findings (brandless-strategy query broadening) fixed; every branch otherwise verified behavior-equivalent against `origin/master`

Closes #779

### DIFF
--- a/apps/server/api/src/collections/agent-strategies/services/agent-strategy-autopilot.service.ts
+++ b/apps/server/api/src/collections/agent-strategies/services/agent-strategy-autopilot.service.ts
@@ -6,6 +6,7 @@ import { AgentStrategiesService } from '@api/collections/agent-strategies/servic
 import { AgentStrategyOpportunitiesService } from '@api/collections/agent-strategies/services/agent-strategy-opportunities.service';
 import { AgentStrategyReportsService } from '@api/collections/agent-strategies/services/agent-strategy-reports.service';
 import { BrandsService } from '@api/collections/brands/services/brands.service';
+import { type ContentDraftDocument } from '@api/collections/content-drafts/schemas/content-draft.schema';
 import { ContentDraftsService } from '@api/collections/content-drafts/services/content-drafts.service';
 import { ContentPerformanceService } from '@api/collections/content-performance/services/content-performance.service';
 import { PerformanceSummaryService } from '@api/collections/content-performance/services/performance-summary.service';
@@ -86,6 +87,18 @@ interface OptimizerAnalysisResult {
   overallScore?: number;
 }
 
+interface FinalizeOpportunityInput {
+  draft: ContentDraftDocument;
+  draftContent: string;
+  format: string;
+  gate: PublishGateResult;
+  opportunity: AgentStrategyOpportunityDocument;
+  organizationId: string;
+  platform: string;
+  strategy: AgentStrategyDocument;
+  userId: string;
+}
+
 @Injectable()
 export class AgentStrategyAutopilotService {
   private readonly defaultEventOpportunityCost = 12;
@@ -130,29 +143,21 @@ export class AgentStrategyAutopilotService {
     const strategy = await this.requireStrategy(strategyId, organizationId);
     const { periodEnd, periodStart } = this.resolveReportWindow('weekly');
 
+    const strategyBrandId = this.strategyBrandId(strategy);
+    const strategyOrganizationId = this.strategyOrganizationId(strategy);
+
     const [drafts, opportunities, performance, summary] = await Promise.all([
       this.contentDraftsService.find({
-        brandId:
-          (strategy as Record<string, unknown>).brandId ??
-          String(
-            (strategy as Record<string, unknown>).brandId ?? strategy.brand,
-          ),
+        brandId: strategyBrandId ?? String(strategy.brand),
         createdAt: { gte: periodStart, lte: periodEnd },
         isDeleted: false,
-        organizationId:
-          (strategy as Record<string, unknown>).organizationId ??
-          String(
-            (strategy as Record<string, unknown>).organizationId ??
-              strategy.organization,
-          ),
+        organizationId: strategyOrganizationId,
       }),
       this.opportunitiesService.listByStrategy(strategyId, organizationId),
       strategy.brand
         ? this.contentPerformanceService.queryPerformance(
             {
-              brand: String(
-                (strategy as Record<string, unknown>).brandId ?? strategy.brand,
-              ),
+              brand: String(strategyBrandId),
               endDate: periodEnd.toISOString(),
               limit: 250,
               startDate: periodStart.toISOString(),
@@ -162,16 +167,10 @@ export class AgentStrategyAutopilotService {
         : [],
       strategy.brand
         ? this.performanceSummaryService
-            .getWeeklySummary(
-              organizationId,
-              String(
-                (strategy as Record<string, unknown>).brandId ?? strategy.brand,
-              ),
-              {
-                endDate: periodEnd,
-                startDate: periodStart,
-              },
-            )
+            .getWeeklySummary(organizationId, String(strategyBrandId), {
+              endDate: periodEnd,
+              startDate: periodStart,
+            })
             .catch(() => null)
         : null,
     ]);
@@ -277,33 +276,23 @@ export class AgentStrategyAutopilotService {
           `Bias next runs toward ${pair.platform}/${pair.format} based on current performance.`,
       );
 
-    const strategyRecord = strategy as Record<string, unknown>;
     return this.reportsService.createReport({
       allocationChanges,
       bestPlatformFormatPairs: snapshot.bestPlatformFormatPairs,
       bestPostingWindows: snapshot.bestPostingWindows,
-      brandId:
-        strategyRecord.brandId ??
-        String((strategy as Record<string, unknown>).brandId ?? strategy.brand),
+      brandId: this.strategyBrandId(strategy) ?? String(strategy.brand),
       clicks: snapshot.clicks,
       costPerVisit: snapshot.costPerVisit,
       creditsSpent: snapshot.creditsSpent,
       ctr: snapshot.ctr,
       generatedCount: snapshot.generatedCount,
       impressions: snapshot.impressions,
-      organizationId:
-        strategyRecord.organizationId ??
-        String(
-          (strategy as Record<string, unknown>).organizationId ??
-            strategy.organization,
-        ),
+      organizationId: this.strategyOrganizationId(strategy),
       periodEnd,
       periodStart,
       publishedCount: snapshot.publishedCount,
       reportType,
-      strategyId:
-        strategyRecord.id ??
-        String((strategy as Record<string, unknown>).id ?? strategy._id),
+      strategyId: this.strategyId(strategy),
       topHooks: snapshot.topHooks,
       topTopics: snapshot.topTopics,
       visits: snapshot.visits,
@@ -395,8 +384,21 @@ export class AgentStrategyAutopilotService {
     return strategy;
   }
 
+  private documentId(entity: unknown): string {
+    const record = entity as Record<string, unknown>;
+    return String(record.id ?? record._id);
+  }
+
+  private opportunityId(opportunity: AgentStrategyOpportunityDocument): string {
+    return this.documentId(opportunity);
+  }
+
+  private draftId(draft: ContentDraftDocument): string {
+    return this.documentId(draft);
+  }
+
   private strategyId(strategy: AgentStrategyDocument): string {
-    return String((strategy as Record<string, unknown>).id ?? strategy._id);
+    return this.documentId(strategy);
   }
 
   private strategyOrganizationId(strategy: AgentStrategyDocument): string {
@@ -709,15 +711,13 @@ export class AgentStrategyAutopilotService {
     defaultModel?: string,
   ): Promise<{ contentGenerated: number; creditsUsed: number }> {
     const strategyOrganizationId = this.strategyOrganizationId(strategy);
-    const strategyBrandId = this.strategyBrandId(strategy);
     const targetPlatform = this.resolveOpportunityPlatform(
       strategy,
       opportunity,
     );
 
     await this.opportunitiesService.updateStatus(
-      ((opportunity as Record<string, unknown>).id as string) ??
-        String(opportunity._id),
+      this.opportunityId(opportunity),
       strategyOrganizationId,
       'generating',
     );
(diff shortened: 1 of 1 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

This PR decomposes `executeOpportunity` by extracting eight private methods and introducing typed ID accessors — genuine wins that shrink the hot path from ~180 inlined lines to a clean dispatch sequence. However, every extracted method stays in the same class, leaving the file 551 lines above the 1,000-line hard ceiling and 120 lines *longer* than before the PR. A decomposition PR that makes its target file larger has not met its stated goal.

---

### 🚫 Blockers

- **`agent-strategy-autopilot.service.ts` (entire file)** — File grew from 1,431 to 1,551 lines in a PR explicitly titled "decompose." All eight new private methods (`generateAutopilotDraft`, `persistAutopilotMetadata`, `handleVideoHold`, `reviseAndReEvaluate`, `handleGateRejection`, `finalizeApprovalAndHandoff`, `finalizeAutoPublish`, `finalizeManualReviewHandoff`) were added to the same class in the same file. Net change is +385 / -265 = +120 lines. The file was already 431 lines above the 1,000-line ceiling before the PR; it is now 551 over. The diff's own `FinalizeOpportunityInput` interface is the natural seam: the three `finalize*` methods accept it and share no cross-cutting state with the generation or gate-evaluation path, making them a clean extraction target.

  _Fix:_ Extract the opportunity-execution sub-pipeline into a new `AgentStrategyOpportunityExecutorService` (or `agent-strategy-execution.service.ts`). Minimum scope: `generateAutopilotDraft`, `persistAutopilotMetadata`, `handleVideoHold`, `reviseAndReEvaluate`, `handleGateRejection`, `finalizeApprovalAndHandoff`, `finalizeAutoPublish`, `finalizeManualReviewHandoff` — approximately 250 lines. Inject the new service into `AgentStrategyAutopilotService`. The parent file should fall below 1,000 lines post-split. `FinalizeOpportunityInput` moves with the new service.

---

### ⚠️ Warnings

- **`agent-strategy-autopilot.service.ts` diff lines 384–405 (`persistAutopilotMetadata`)** — The method builds a metadata object AND issues a `contentDraftsService.patch()` DB write, then returns the metadata for use as input to `reviseAndReEvaluate`. The name implies a void persistence operation; callers read `const autopilotMetadata = await this.persistAutopilotMetadata(...)` and see a data-extraction, not a mutation. The write sequence is now split across two methods: `persistAutopilotMetadata` issues the first patch, and `reviseAndReEvaluate` (diff line 453) issues a second patch on the same draft — with no single site in `executeOpportunity` where both writes are visible together.

  _Fix:_ Split into a pure synchronous `buildAutopilotMetadata(...)` (returns `Record<string, unknown>`, no `await`) and an explicit `contentDraftsService.patch` call at the call site in `executeOpportunity`, alongside the other writes. This makes the write sequence visible at the orchestration layer and makes the builder testable without a DB mock. The metadata object is a 6-key spread that does not warrant a separate helper; inlining the construction and making the write explicit is the cleaner outcome.

- **`agent-strategy-autopilot.service.ts` diff lines 423–497 (`reviseAndReEvaluate` return type)** — The method returns `Promise<{ result: { contentGenerated: number; creditsUsed: number }; terminal: true } | { terminal: false }>`. The `{ terminal: false }` branch carries no payload — it is a unit type. The caller pattern at diff line 290 (`if (revision.terminal) { return revision.result; }`) is semantically equivalent to a null check. The `terminal` discriminant adds ceremony without expressiveness.

  _Fix:_ Change return type to `Promise<{ contentGenerated: number; creditsUsed: number } | null>`. Return `null` when the revised gate passes (continue to finalization). Rename to `tryReviseOrDiscard` to make the nullable-return contract legible. Call site becomes `const revisionResult = await this.tryReviseOrDiscard(...); if (revisionResult) return revisionResult;` — one field removed, same semantics.

- **`agent-strategy-autopilot.service.ts` diff lines 526–534 (`finalizeApprovalAndHandoff`)** — This method contains exactly three lines of logic: a single `if/else` routing to `finalizeAutoPublish` or `finalizeManualReviewHandoff`. It has one call site, performs no setup, no teardown, no logging, and enforces no invariant. The intermediate dispatcher adds a call-stack hop for a reader who must now trace three hops (`executeOpportunity` → `finalizeApprovalAndHandoff` → one of two leaf methods) to follow what was previously a flat branch.

  _Fix:_ Inline the two-branch `if/else` directly into `executeOpportunity`, calling `finalizeAutoPublish` or `finalizeManualReviewHandoff` directly. `FinalizeOpportunityInput` remains useful as a parameter bag and does not require a dispatcher to justify it.

- **`agent-strategy-autopilot.service.ts` diff lines 130–141 (`documentId`, `opportunityId`, `draftId`)** — `documentId(entity: unknown)` is a new private copy of the `String(record.id ?? record._id)` pattern that already exists as private methods in `auth-identity-resolver.service.ts:22`, `better-auth-identity-resolver.service.ts:18`, and `onboarding.service.ts:146`, and appears 40+ times inline across 20+ files in `apps/server/api/src/`. The fix scopes the abstraction too narrowly: a private method is invisible to every other call site. Additionally, `opportunityId` and `draftId` are typed wrappers that immediately widen their parameter to `unknown` before passing to `documentId` — providing zero type narrowing over calling `documentId` directly.

  _Fix:_ Add `export function extractDocumentId(entity: unknown): string { const r = entity as Record<string, unknown>; return String(r.id ?? r._id); }` to `packages/helpers/src/` (e.g. `document-id.helper.ts`). Import from `@genfeedai/helpers` in this service (which already depends on it). Remove the private `documentId` method and the `opportunityId`/`draftId` typed wrappers — callers can use `extractDocumentId(draft)` directly. This makes the fix available to the 20+ other files with the same inline pattern. For Prisma documents where `id: string` is typed directly (confirmed in schema.prisma for all three document types), prefer `entity.id` over the helper where the index signature permits it.

- **`agent-strategy-autopilot.service.ts` diff lines 17–27 (`FinalizeOpportunityInput`)** — The PR introduces a fourth module-level private `interface` directly in the service file (`FinalizeOpportunityInput` joins the pre-existing `ExecuteRunResult` and `OptimizerAnalysisResult`). The project rule is explicit: "No inline interfaces in component/service files — place in dedicated types packages or colocated `*.types.ts` files." This PR reorganized the entire service and was the natural moment to move all three to a colocated types file. A `agent-strategy-autopilot.types.ts` in the same directory is the acceptable minimum; `packages/interfaces/src/ai/` is the correct target if the type ever crosses module boundaries (it currently does not).

  _Fix:_ Create `apps/server/api/src/collections/agent-strategies/services/agent-strategy-autopilot.types.ts` and move `FinalizeOpportunityInput`, `ExecuteRunResult`, and `OptimizerAnalysisResult` into it. Import them back in the service file. This PR is the right time.

- **`agent-strategy-autopilot.service.ts` lines 402–416 (`strategyOrganizationId`, `strategyBrandId`)** — The PR's stated goal is "typed ID accessors" to eliminate `as Record<string, unknown>` scatter. Yet the two most important accessors — `strategyOrganizationId` and `strategyBrandId` — still use `(strategy as Record<string, unknown>).organizationId` and `(strategy as Record<string, unknown>).brandId` internally. Both `organizationId` (`String @id`) and `brandId` (`String?`) are declared in the Prisma schema and are present on `PrismaAgentStrategy`; neither is omitted by `AgentStrategyDocument`'s `Omit<PrismaAgentStrategy, 'config' | 'policies' | 'agentType' | 'platforms'>`. The abstraction the PR created is half-finished: it hides repeated casts behind named accessors but the casts inside those accessors are still unnecessary.

  _Fix:_ Re-declare `organizationId: string` and `brandId: string | null | undefined` explicitly in `AgentStrategyDocument` (or in the type extension). Once the index-signature ambiguity is resolved by explicit field declarations, the `as Record<string, unknown>` casts inside `strategyOrganizationId` and `strategyBrandId` can be removed, leaving direct property access. The `?? strategy.organization` fallback for legacy records remains warranted as a runtime guard regardless of the cast change.

---

### Verdict

**BLOCK**

The PR delivers real, measurable wins: `executeOpportunity` is now a clean 80-line dispatch sequence rather than a 180-line deeply-nested function, repeated `as Record<string, unknown>` casts are consolidated behind named accessors, and each execution path has a named method. These are genuine complexity reductions.

However, the PR cannot merge as-is because a "decompose" PR that leaves the file 551 lines above the 1,000-line threshold — and 120 lines *longer* than before — has not completed its stated job. The natural extraction boundary already exists in the PR's own work (`FinalizeOpportunityInput` crosses no shared-state boundary with the generation/evaluation path). The blocker is one structural split away from resolution.

**Must resolve before merge:**
- Extract the opportunity-execution sub-pipeline into a dedicated service; bring the parent file below 1,000 lines post-split.

PR #842, opened 2026-06-26
refactor(videos): extract create() into VideoGenerationService, unify provider routing (#835)

VincentShipsIt:
## Summary

`videos.controller.ts` was a 1,370-line god-controller whose business logic was an **835-line `create()`** (~61% of the file). This extracts the full generation workflow into a new **`VideoGenerationService`**; the controller's `create()` is now a thin handler that keeps all decorators/guards/interceptors and delegates. ~25 of the 27 deps move off the controller.

Closes #835.

## What changed

- **New `services/video-generation.service.ts`** — owns model resolution, deferred credit math, prompt building, document persistence, provider fan-out, credit deduction, activity/websocket events, bookmark linking, music orchestration, and optional polling.
- **Provider routing fix (latent bug).** The two divergent `switch` statements are collapsed into a single `dispatchVideoGeneration()` that routes via `isFalDestination()` (`model-key.util`), matching the `images-operations` precedent. The second (multi-output) switch was missing `FAL_KLING_VIDEO_V3_PRO`, `FAL_VEO_3_1` and `FAL_PIXVERSE_V6`, so **additional outputs of those models silently fell through to Replicate**. Routing now lives in exactly one place, structurally eliminating the missing-key bug class.
- **Deleted the no-op `Promise.all(... { where: {} })` block** (it did nothing).
- **Extracted `createVideoPlaceholderActivity()`** used by the primary, batch and non-batch paths (was triplicated activity-create + websocket-publish).
- **Extracted `resolveVideoModel()` and `ensureDeferredCredits()`** as private methods (each <80 lines).

## Behavior preservation

Routes, decorators, guards, status codes and JSON:API response shape are unchanged. `videos.controller.spec.ts` is rewired to build the real `VideoGenerationService` from the same mocked providers (no expectation changes), plus a **new regression test** asserting every output of a multi-output `FAL_VEO_3_1` / `FAL_KLING_VIDEO_V3_PRO` / `FAL_PIXVERSE_V6` request routes to FAL, never Replicate.

## Acceptance criteria

- [x] No file in the videos controllers/services module exceeds ~500 lines; no method exceeds 80 lines.
- [x] Provider routing exists in exactly one place; test asserts multi-output FAL requests route all outputs to `falService`.
- [x] The no-op `Promise.all` block is gone.
- [x] Activity-create + websocket-publish lives in one helper, invoked from all three paths.
- [x] `videos.controller.spec.ts` passes (call-site/mock relocations only, no expectation changes).
[...]

Issue #835, opened 2026-06-26
Move the 835-line create() out of videos.controller.ts (1,370 lines) into a service and fix the divergent provider-routing switch


VincentShipsIt:
_Filed by the automated code-quality sweep (read-only review). Findings below were verified against current source by an adversarial second pass (no corrections needed)._

## Summary

`apps/server/api/src/collections/videos/controllers/videos.controller.ts` is a 1,370-line god-controller whose business logic is dominated by a single **835-line `create` method (lines 493-1327, ~61% of the file)** that resolves models, validates org access, performs deferred credit math, builds prompts, persists documents, fans out to external providers (KlingAI/Fal/Replicate), deducts credits, creates activities, emits websocket events, links bookmarks, kicks off background music orchestration, and optionally polls for completion, all inline. The provider-routing logic is duplicated as two divergent switch statements, and the duplication has already produced a **latent routing bug**. The sibling `images` collection already establishes the correct fix (split into `operations/images-operations.controller.ts` and route providers via `model-key.util` helpers), and an existing 43KB `videos.controller.spec.ts` makes the refactor test-guarded.

## Confirmed findings

**1. Oversized file driven by one monster method.** Method map: `findLatest` (173-206), `findAll` (216-297), `findOne` (307-366), `getThumbnail` (370-438), `remove` (442-474), `create` (493-1327), `calculateDynamicVideoCost` (1333-1369). `create` alone is 835 lines, ~10x the 80-line threshold.

**2. Duplicated, divergent provider-routing switch (confirmed correctness bug).** The primary switch (827-864) routes 8 FAL model keys plus `KLINGAI_V2` and a Replicate default. The second copy inside the multi-output loop (1063-1098) lists only 5 FAL keys (836-843 vs 1072-1076), omitting `FAL_KLING_VIDEO_V3_PRO`, `FAL_VEO_3_1`, and `FAL_PIXVERSE_V6`. The two `falService.generateVideo` bodies (844-855, 1077-1088) are otherwise identical.

*Impact:* for a multi-output (`outputs > 1`) request using one of those three FAL models, the **first** output goes to `falService.generateVideo` (correct) but every **additional** output falls through the second switch's default to `replicateService.generateTextToVideo`, a different provider. Outputs in one request silently route to two backends, purely from copy-pasting the switch and not keeping the copies in sync.

[...]

PR #848, opened 2026-06-26
refactor(images): extract create() into ImageGenerationService, dedup provider scaffolding (#778)

### LINKED ISSUE #778: Decompose image generation create
_Filed by the automated code-quality sweep (read-only review). Findings below were verified against current source._

## Summary

`apps/server/api/src/collections/images/controllers/operations/images-operations.controller.ts` is 1,494 lines for just **two endpoints**, and its `create()` handler is a single **~1,090-line method** (lines 158–1248; the next endpoint `@Post(':id/split')` begins at 1253). The method is a flat chain of per-provider branches with copy-pasted failure-handling, polling, and timeout-recovery scaffolding, and it carries provider-routing and credit logic that belongs in a service. This mirrors the already-filed posts-operations.controller refactor (#691), applied to images.

## Confirmed evidence

`apps/server/api/src/collections/images/controllers/operations/images-operations.controller.ts` (1,494 lines)

- `async create(...)` spans **lines 158–1248 (~1,090 lines)**; the next endpoint is at line 1253. The controller exposes only two endpoints total.
- The constructor injects **27 dependencies** (lines ~110–139), a direct symptom of one method doing too much.
- The body is a flat sequence of provider branches (not `else if`), each ~100 lines, mutating shared `url`/promise state: `isGenfeedAi` (~501), `isKlingAI` (~598), `isFal` (~695), `isLeonardoAI` (~810), `isReplicate` (~904).
- Duplicated scaffolding, confirmed by grep:
  - `failedGenerationService.handleFailedImageGeneration(...)` appears **5×**.
  - The `if (waitForCompletion) { ... }` poll-and-serialize block appears **4×**.
  - The `PollingTimeoutError` recovery block (re-fetch ingredient, throw `GATEWAY_TIMEOUT`) is repeated ~3× with near-identical bodies.
- Controller-inappropriate business logic inlined: model auto-routing/resolution (~208–261), deferred-credit recompute + balance check (~272–324), provider capability classification (~326–343).

Adding a provider means pasting another ~100-line block plus its own copies of the failure-handler call, the `waitForCompletion` poll block, and the timeout-recovery block — four places to keep in sync. A fix in timeout handling must be applied in three spots and is easy to miss in one. The method is far past any reviewable size and untestable without driving the entire 27-dependency pipeline, while a shared mutable `url` threaded through every block obscures control flow.

1. Define an `ImageProviderHandler` interface (`supports(model)`, `enqueue(ctx)`) and extract each provider block (GenfeedAi/Kling/Fal/Leonardo/Replicate/SDXL) into its own handler owning only provider-specific input-building and enqueue.
2. Extract the duplicated tail into one private helper `awaitOrBackground(...)` holding the single canonical copy of the `waitForCompletion` poll -> serialize -> `PollingTimeoutError` recovery logic.
3. Wrap `failedGenerationService.handleFailedImageGeneration` in one `handleProviderFailure(...)` so the 5 call sites collapse to one.
4. Move model resolution (~208–261) and deferred-credit handling (~272–324) into a service (e.g. an orchestrator/router/credits service).
5. Reduce `create()` to: validate -> resolve model -> ensure credits -> build prompt -> dispatch to handler -> `awaitOrBackground(...)`.

## Acceptance criteria

- `create()` is <= 150 lines; no method in the file exceeds 150 lines.
- The failure handler, the `waitForCompletion` block, and the `PollingTimeoutError` recovery each appear exactly once on the controller's path.
- Provider selection is a dispatch map/array, not a flat `if` chain; adding a provider touches only a new handler plus registration.
- No change to request/response contract, status codes (`BAD_REQUEST`/`FORBIDDEN`/`PAYMENT_REQUIRED`/`GATEWAY_TIMEOUT`), or background-vs-wait behavior; existing tests pass unchanged.

VincentShipsIt:
## Summary

`images-operations.controller.ts` was **1,494 lines** dominated by a **~1,090-line `create()`** with **27 injected deps**. This extracts the full image-generation workflow into a new `ImageGenerationService`; the controller's `create()` is now a thin handler that keeps all decorators/guards/interceptors and delegates. The controller constructor drops from 29 deps to **8** (split-endpoint deps + the service). Mirrors the videos #835 `VideoGenerationService` precedent (PR #842).

## Decomposition

- **Provider selection is a dispatch map** (`classifyProvider` → `dispatchGeneration` switch), not a flat `if`-chain. Adding a provider touches only a new handler + the switch.
- The **5** copy-pasted `failedGenerationService.handleFailedImageGeneration` calls collapse to one `handleProviderFailure()`.
- The **3** duplicated `PollingTimeoutError` recovery blocks collapse to one `throwGatewayTimeoutIfPending()`.
- The poll/serialize tails collapse into `finishGeneration()` + `resolveCompletedIngredient()`, keyed by a `CompletionPlan` kind that preserves each provider's exact wait semantics: GenfeedAi `inline` (no timeout recovery, as before) · Kling+Leonardo `poll-single` · Replicate `poll-multiple` · Fal `background-only` · SDXL `none`.
- The placeholder activity+websocket triplication collapses to one `createImagePlaceholderActivity()`.
- The shared mutable `url` label is gone — each handler passes its own provider label.

## Acceptance criteria (issue #778)

- [x] `create()` ≤ 150 lines (now a ~6-line delegate).
- [x] No method in the new service exceeds 150 lines (largest ~112). _Pre-existing `splitContactSheet` (~180 lines) is left untouched — it is unrelated to the `create()` god-method and refactoring it would be out of scope._
- [x] Provider selection is a dispatch map, not a flat `if` chain.
- [x] Failure handler / `waitForCompletion` tail / `PollingTimeoutError` recovery each on the path exactly once.
- [x] No change to request/response contract, status codes (`BAD_REQUEST`/`FORBIDDEN`/`PAYMENT_REQUIRED`/`GATEWAY_TIMEOUT`), or background-vs-wait behavior.

## Behavior preservation

Verified by a multi-agent adversarial review (8 findings surfaced, **0 confirmed** after adversarial verification — the one real ordering change is the invalid-model `BAD_REQUEST` now firing *before* the read-only reference fetch, which changes no status code and no asserted side effect). Notable invariants preserved: Fal pushes additional ids to the shared `pendingIngredientIds`; Replicate tracks its own local poll-id array; the `waitForCompletion`-for-unsupported-provider warn fires only for SDXL.

## Verification

- `bunx tsc --noEmit -p tsconfig.typecheck.json` (apps/server/api) → **0 errors**.
- `bunx biome check` on all touched files → clean.
- `images-operations.controller.spec.ts` → **38/38 pass**. Spec rewired to construct the real `ImageGenerationService` from the same mocks; all assertions unchanged.
- Full workspace gates (lint/build/e2e) deferred to CI per laptop resource limits.

Closes #778


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

This PR extracts the 1,091-line monolithic `create()` method out of `ImagesOperationsController` into a new `ImageGenerationService`, achieving its primary decomposition goal — the controller is now a thin 9-line delegate — but lands all of that complexity in a single 1,396-line service file and misses several consolidation opportunities that were squarely within scope.

**Genuine wins:** The controller drops from 1,494 to ~379 lines. The scattered inline logic is now structured into 22 focused methods with named types (`GenerationContext`, `CompletionPlan`), a single `handleProviderFailure` path, a single `finishGeneration` tail, and the `IMAGE_POPULATE` constant deduplicated. The test refactor correctly splits the DI graph between the controller and the extracted service. These are real improvements.

---

### ⚠️ Warnings

- **`image-generation.service.ts:1–1396`** — The extracted service is 1,396 lines — 40% over the 1,000-line decomposition ceiling — while eight existing services in this codebase already exceed that mark (largest at 2,649 lines). The structural direction is correct, but the file is not finished. The five provider dispatch methods (`dispatchGenfeedAi` ~75 lines, `dispatchKlingAI` ~40, `dispatchLeonardo` ~35, `dispatchFal` ~50, `dispatchReplicate` ~95) plus three Replicate sub-helpers (~239 lines combined) account for ~534 provider-specific lines that share no data except `GenerationContext`. The `dispatchGeneration` switch already draws the correct split boundary. A follow-up PR extracting a `ReplicateImageGenerationService` (the most complex branch at ~334 lines with its sub-helpers) would bring the orchestrator under 1,100 lines and make the Replicate multi-output paths independently testable.
  _Fix:_ File a follow-up issue targeting `ReplicateImageGenerationService` extraction as the minimum viable split. The `KlingAI`/`Leonardo` pair are near-identical and are a second natural extraction.

- **`image-generation.service.ts:1484–1488` + `line 2096`** — `CompletionPlan.pollIds` is typed `string[] | undefined` across all four `kind` values. Only `dispatchReplicate` ever produces `kind: 'poll-multiple'`, and it always initialises `pollIds`. The `plan.pollIds ?? [context.ingredientData._id.toString()]` fallback at line 2096 is dead code that exists only to satisfy a type contract the PR itself made too loose. The `kind` discriminant was introduced specifically to give callers a compile-time contract; this defeats it at the one field that actually varies by kind.
  _Fix:_ `type CompletionPlan = { kind: 'inline' | 'poll-single' | 'background-only'; generationPromise: Promise<unknown> } | { kind: 'poll-multiple'; generationPromise: Promise<unknown>; pollIds: string[] }`. Remove the `??` fallback at the call site — the compiler now enforces the invariant.

- **`image-generation.service.ts:1899–1958`** — `ensureDeferredCredits` reads and writes `request.creditsConfig` via an `as unknown as` cast, mutating HTTP interceptor state from the service layer. `CreditsInterceptor.intercept` reads `request.creditsConfig.deferred` back after the handler returns. This was acceptable in the controller (same HTTP layer as `@UseInterceptors`) but is a layer violation in a service. The parallel deferred-credits callers in `insights.controller.ts:303` and `contexts.controller.ts:313` both keep this mutation inline in the controller, confirming the established convention.
  _Fix:_ Move `ensureDeferredCredits` back into the controller's `create()` method and pass the resolved `requiredCredits` as a typed parameter to `imageGenerationService.generateImage(...)`. The service stays pure of HTTP lifecycle knowledge.

- **`image-generation.service.ts:2714–2743`** — `calculateDynamicImageCost` carries a JSDoc that reads _"Mirrors the CreditsGuard.calculateDynamicCost logic for image models."_ The canonical implementation is at `credits.guard.ts:648` (a three-branch switch covering `PER_MEGAPIXEL`, `PER_SECOND`, and `per-token`). The image copy handles only `PER_MEGAPIXEL`, silently dropping the other pricing types. A second variant already exists at `videos.controller.ts:1333` with an identical comment. This PR relocated the method from the controller into the new service — the natural moment to extract a shared utility — and did not take it. No canonical helper exists in `packages/helpers`, `packages/utils`, or `packages/core` (grep confirms).
  _Fix:_ Extract to `apps/server/api/src/helpers/utils/credits/dynamic-cost.util.ts`. Signature: `calculateDynamicModelCost(model: { cost?: number; pricingType?: PricingType; costPerUnit?: number; minCost?: number }, width?: number, height?: number, duration?: number): number`. Replace the body of all three private copies with delegate calls. Add the `PER_SECOND` branch to the image-specific path — image models can be billed per second in the future and silently returning zero cost is a latent billing bug.

- **`image-generation.service.ts:1843`** — `resolveImageModel` types its `organizationSettings` parameter as `{ defaultImageModel?: unknown } | null` rather than `OrganizationSettingDocument | null`, which is what `organizationSettingsService.findOne` actually returns. `PrismaOrganizationSetting.defaultImageModel` is `String?` — the real type is `string | null`, not `unknown`. The loose typing forces the `as string` cast on line 1881 (`modelKeys.includes(organizationSettings.defaultImageModel as string)`); with the correct type, the truthy guard would narrow to `string` without a cast. The no-inline-interfaces rule is also violated.
  _Fix:_ Change the parameter type to `Pick<OrganizationSettingDocument, 'defaultImageModel'> | null`. The `as string` cast is then unnecessary and should be removed.

- **`image-generation.service.ts:1455–1461`** — `GenerationContext.promptBuilderBrand` re-declares `{ description?: string; label: string; primaryColor?: string; secondaryColor?: string; text?: string }` inline. This shape is field-for-field identical to `PromptBuilderParams['brand']` at `prompt-builder-params.interface.ts:50–56`. The codebase already uses the indexed-access pattern for this exact type in `evaluations-operations.service.ts:20`.
  _Fix:_ `promptBuilderBrand: NonNullable<PromptBuilderParams['brand']>`. Import `PromptBuilderParams` from `@api/services/prompt-builder/interfaces/prompt-builder-params.interface`.

- **`image-generation.service.ts:1760–1805`** — In `persistImageDocuments`, `promptsService.create(...)` and `promptBuilderService.buildPrompt(...)` are awaited in series. `buildPrompt` receives `prompt: promptData.original`, which equals `promptOriginalText` invariantly (the `PromptEntity` constructor is `Object.assign(this, partial)` with no transformation of `original`). `promptData._id` is not consumed until `saveDocuments` downstream. The two I/O round-trips are fully independent and this is the hot path for every generation request.
  _Fix:_ `const [promptData, { templateUsed, templateVersion }] = await Promise.all([this.promptsService.create(...), this.promptBuilderService.buildPrompt(..., { prompt: promptOriginalText, ... }, ...)])`. Replace `prompt: promptData.original` with `prompt: promptOriginalText` in the `buildPrompt` call.

---

### ✂️ Simplifications

- **`image-generation.service.ts:2082`** — `resolveCompletedIngredient` is declared `Promise<unknown>`. All three return paths produce `IngredientDocument`-derived values. The PR already establishes `NonNullable<Awaited<ReturnType<...>>>` aliases for `ResolvedBrand`, `ResolvedPrompt`, etc. — using `Promise<unknown>` here is an inconsistency. Replace with `Promise<IngredientDocument | null>` (the inline branch genuinely can return `null`).

- **`image-generation.service.ts:1865, 1886`** — `recommendation.selectedModel as string` and `(await this.routerService.getDefaultModel(...)) as string` are no-op casts. `ModelRecommendation.selectedModel` is declared `string` (non-optional) at `router.interfaces.ts:27`; `getDefaultModel` returns `Promise<string>` at `router.service.ts:598`. Both were carried forward from the controller unchanged. Remove both `as string` casts.

- **`image-generation.service.ts:2592–2616`** — `createReplicateBatchOutputs` runs two sequential `Promise.all` blocks whose outputs are not interdependent: block 1 patches `metadataService` and `imagesService`; block 2 calls `createImagePlaceholderActivity`. Neither block reads the other's output. Merge into a single `flatMap`-fed `Promise.all` to eliminate an unnecessary serialization point.

---

### Verdict

**APPROVE** — No finding rises to a rejection-level defect given codebase norms and the confirmed runtime correctness of the extracted service. The refactor achieves its stated goal: the controller's 1,091-line monolith is gone. The warnings above are real structural debts — the `CompletionPlan` discriminated union, the `ensureDeferredCredits` layer violation, the `calculateDynamicImageCost` duplication, and the hot-path sequential awaits in `persistImageDocuments` are all worth addressing promptly in a follow-up. The file size issue should be tracked as a decomposition obligation before the images collection grows further.

PR #849, opened 2026-06-27
refactor(articles): dedup virality/credit logic, decompose generators (#716)

### LINKED ISSUE #716: Deduplicate articles services
## Summary

The `apps/server/api/src/collections/articles/services/` cluster carries two oversized files — `articles.service.ts` (1,548 lines) and `articles-content.service.ts` (1,527 lines) — with a genuine **cross-file duplication**: an entire virality/credit pipeline exists in `articles.service.ts` *and* in its sibling `articles-analytics.service.ts`. There are also three oversized AI-generation methods. Not covered by any existing open quality issue.

Directory context (for reference): `articles.service.ts` (1,548, CRUD/orchestration facade), `articles-content.service.ts` (1,527, AI generation/edit/review), `articles-analytics.service.ts` (586, virality + performance orchestration), `article-analytics.service.ts` (277, Prisma CRUD over `articleAnalytics`). Note the singular/plural name collision between the latter two.

### Cross-file duplication (highest value)
The following members of `ArticlesService` are functional duplicates of members in `ArticlesAnalyticsService`:
- `analyzeVirality` — `articles.service.ts:1177` vs `articles-analytics.service.ts:148` (same `ARTICLE_VIRALITY` template, same `buildPrompt`/`generateTextCompletionSync`/`JSON.parse`/`buildViralityAnalysisResponse` pipeline).
- `updatePerformanceMetrics` — `articles.service.ts:1277` vs `articles-analytics.service.ts:245` (both delegate to `ArticleAnalyticsService.updatePerformanceMetrics(normalizePerformanceMetrics(...))` with the same warn-and-skip guard).
- `assertDefaultTextCreditsAvailable` — `articles.service.ts:228` vs `articles-analytics.service.ts:434`.
- `settleDefaultTextCredits` — `articles.service.ts:258` vs `articles-analytics.service.ts:467`.
- `getDefaultTextModel` — `articles.service.ts:287` vs `articles-analytics.service.ts:523` (essentially byte-identical).
- `TEXT_MAX_OVERDRAFT_CREDITS = 5` static duplicated (`articles.service.ts:81` and `articles-analytics.service.ts:49`).

There are now two parallel virality/credit implementations in the same directory; only one should exist.

### Oversized methods
- `articles-content.service.ts`: `generateLongFormArticle` (386-611, **~225 lines**), `generateArticles` (192-380, **~188 lines**), `enhance` (619-744, **~125 lines**), `convertToTwitterThread` (749-875, ~126 lines). The three generate/enhance methods share a large common prologue (template fetch → harness build → prompt build → Replicate call → billing → JSON parse) that differs mainly in template key and response shape.
- `articles.service.ts`: `update` (479-639, **~160 lines**) mixes id validation, publish-state logic, ownership re-check, cache invalidation, and an embedded Discord-notification block (579-628).

### Repeated patterns within `articles.service.ts`
- **Id-validation triad** (`if (!userId) / if (!organizationId) / if (!brandId)`) duplicated 4×: `createArticle` (370-378), `findOneArticle` (425-435), `update` (493-503), `removeArticle` (651-659).
- **Cache-invalidation tag block** copy-pasted 3-4×: `createArticle` (390-407), `update` (550-573), `removeArticle` (677-696), `generateArticles` (857-874).
- **Config inconsistency / latent bug**: `getTranscriptById` (309) builds its URL from `configService.get('GENFEEDAI_API_URL')`, but `linkTranscriptToArticle` (331) hardcodes `http://localhost:3010` — a latent prod bug, not just style.
- `getPopulationForContext` (354-358) always returns `[]` with a "kept for signature compatibility" comment — a deletion candidate under the repo's green-field "delete dead code" rule.

### Honest non-findings (avoid over-reach)
- The lazy `ModuleRef` getter for `articlesService` (`articles-content.service.ts:139-145`) is justified in its comment (breaks an ESM TDZ circular-dep crash) — leave it.
- Parsers, token estimators, and prompt builders are appropriately sized — leave them.

Two divergent virality/credit pipelines in one directory mean a pricing or prompt-template change must be made in two places or silently diverge, directly affecting billing correctness. The ~225-line generator and its near-identical siblings force readers to re-parse the same prologue three times, and the hardcoded `localhost:3010` will break transcript linking in production.

## Proposed behavior-preserving restructuring
1. **Kill the virality/credit duplication.** Remove `analyzeVirality`, `updatePerformanceMetrics`, `assertDefaultTextCreditsAvailable`, `settleDefaultTextCredits`, `getDefaultTextModel`, and `TEXT_MAX_OVERDRAFT_CREDITS` from `articles.service.ts`; make `ArticlesService.analyzeVirality`/`updatePerformanceMetrics` thin delegators to `ArticlesAnalyticsService` (mirroring how it already delegates to `ArticlesContentService`). **Caveat:** the `ArticlesService` variant persists `viralityAnalysis` via `patch` (1256) while the analytics one does not — preserve the persist step in the delegator. Removes ~190 lines.
2. **Extract helpers in `articles.service.ts`:** `assertArticleOwnershipIds(...)` (replaces the 4 validation triads), `invalidateArticleCaches(opts?)` (replaces the 3-4 cache blocks), and split `update` by extracting `applyPublishStateTransition(...)` and `sendArticlePublishedNotification(...)`. Fix `linkTranscriptToArticle` to use `configService`. Result: ~1,548 → ~1,150 lines.
3. **Decompose the AI generators in `articles-content.service.ts`:** extract the shared prologue into `runTextGenerationStep({...})` consumed by `generateArticles`/`generateLongFormArticle`/`enhance`; extract `buildXArticleContentAndMetadata(sections)` from `generateLongFormArticle` (526-556); move `convertToTwitterThread`'s text-splitting (793-836) into a pure `articles/utils/article-thread.util.ts`. Result: ~1,527 → ~1,150 lines; each generate/enhance method under 80 lines.
4. **Follow-up (optional):** rename to resolve the `articles-analytics` vs `article-analytics` collision.

## Acceptance criteria
- The virality analysis, performance-metrics update, and credit helpers exist in exactly one service; `ArticlesService` delegates and still persists `viralityAnalysis`.
- No method in either file exceeds ~120 lines; the three generate/enhance methods share one prologue helper.
- Id validation and cache invalidation each exist in one helper; `linkTranscriptToArticle` no longer hardcodes localhost.
- Public article CRUD, generation, virality, and analytics behavior is unchanged.
- Existing tests still pass; new unit tests cover the shared generation step and the thread-splitting util.

VincentShipsIt:
Closes #716

## Summary

Deduplicates the `apps/server/api/src/collections/articles/services/` cluster and decomposes its oversized AI-generation methods. **Behavior-preserving; net −276 lines.**

## The dedup — corrected direction (key decision)

The issue proposed making `ArticlesService` delegate **into** `ArticlesAnalyticsService`. Research showed that's backwards and unsafe:

- `ArticlesAnalyticsService` had **zero consumers** anywhere in the repo — it was registered in `articles.module.ts` but never injected or called. `ArticlesService` is the live virality/credit path wired to the controller (`articles.controller.ts:414`).
- The two `analyzeVirality`/credit implementations had **diverged**: `ArticlesService` uses `DEFAULT_MINI_TEXT_MODEL` + a `maxTokens` cap, **persists** `viralityAnalysis`, and does **not** charge credits; the dead one used the full `DEFAULT_TEXT_MODEL`, **charged** credits, and didn't persist. Delegating would have silently changed the model and billing.

So the genuinely behavior-preserving move (and the repo's "delete dead code aggressively" rule) is to **delete the dead `ArticlesAnalyticsService`** and its 3 module registrations. The live `article-analytics.service.ts` (singular Prisma CRUD) is untouched.

## Other changes (live files, pure refactors + one bug fix)

`articles.service.ts`
- Extract `assertArticleOwnershipIds` (replaces 4 identical id-validation triads).
- Extract `invalidateArticleListCaches(context, includePublic)` (replaces the create/update/remove cache-tag blocks; `update` still conditionally busts `public`). `generateArticles`' distinct `AggregationCacheUtil` invalidation is intentionally left as-is to preserve behavior.
- Split `update()` into `applyPublishStateTransition` + `sendArticlePublishedNotification`.
- **Fix:** `linkTranscriptToArticle` used a hardcoded `http://localhost:3010` (latent prod bug) — now uses `configService.get('GENFEEDAI_API_URL')`, matching `getTranscriptById`.
- Delete the dead `getPopulationForContext` (+ its unused `PopulateOption` import).

`articles-content.service.ts`
- Extract the shared `runTextGenerationStep` prologue used by `generateArticles` / `generateLongFormArticle` / `enhance` (model, prompt-builder options, failure message, and billing are parameterized so each method's behavior is unchanged — `enhance` remains unbilled).
- Extract `buildXArticleContentAndMetadata` from `generateLongFormArticle`.
- Move the Twitter/X thread text-splitting into a pure `utils/article-thread.util.ts` (`buildTwitterThreadTweets`), preserving the historical newline-collapse quirk exactly.

## Acceptance criteria
- [x] Virality/credit logic exists in exactly one service; live behavior unchanged.
- [x] The three generate/enhance methods share one prologue helper; no method exceeds ~120 lines.
- [x] Id-validation and list/aggregation cache invalidation each live in one helper.
- [x] `linkTranscriptToArticle` no longer hardcodes localhost.
- [x] Existing tests pass; new unit tests cover the shared generation step and the thread-splitting util.

## Verification
- `bun run vitest` (scoped): new `article-thread.util.spec.ts` + `articles-content.service.run-step.spec.ts` — **11 passing**; existing `src/collections/articles/**` suite — **47 passing** (incl. `articles.controller.spec.ts`). `articles.module.spec.ts` errors only on an unbuilt `@genfeedai/tools` dist via an unrelated import chain — environmental, green on CI.
- `tsc -p apps/server/api --noEmit` on Mac Studio: **zero errors in `collections/articles`** (only a pre-existing `baseUrl` TS5101 deprecation in `tsconfig.json`, untouched by this PR).
- `biome check`: clean.
- 18-agent adversarial diff review (find → verify): **0 confirmed defects**.


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

This PR deletes a 586-line duplicate analytics service (`articles-analytics.service.ts`), consolidates three copy-pasted generation pipelines into a shared `runTextGenerationStep` prologue, and extracts four repeated private helper blocks in `articles.service.ts` — genuine complexity reduction, and the net line delta across the articles domain is roughly -635 lines. The refactor goal is achieved. No surviving blockers; several structural loose ends worth cleaning up before this area is touched again.

---

### ⚠️ Warnings

- **`apps/server/api/src/collections/articles/services/articles.service.ts:1–1548`** — File is 1,548 lines post-PR (threshold: 1,000). The PR reduced it by only 5 net lines while adding four new private methods. Three distinct concern clusters coexist: CRUD, transcript generation, and the publish-state pipeline. The PR was the natural opportunity to start splitting; it did not.
  _Fix:_ Track a follow-on task to extract virality/analytics methods into `ArticlesViralityService` and transcript methods into `ArticlesTranscriptService`, both as providers in `ArticlesModule`. Each resulting file should land under 600 lines.

- **`apps/server/api/src/collections/articles/services/articles-content.service.ts:1–1527`** — File is 1,527 lines post-PR. Three generation pipelines (standard article, X Article long-form, enhancement), a review cycle, thread conversion, harness assembly, and account-publishing resolution all live here. The PR touched all three pipelines — it was the right moment to split — but added the new private helpers inside the existing file instead.
  _Fix:_ A dedicated decomposition pass should promote the generation pipelines into an `ArticleGenerationPipeline` service and the thread conversion into `ArticlesThreadService`. Track alongside the service above.

- **`apps/server/api/src/collections/articles/services/articles-content.service.ts:1167–1219`** — `buildXArticleContentAndMetadata` is a pure data-shaping function (zero `this.` references in the body, no I/O, no injected dependencies) extracted as a private class method rather than a standalone util. This is internally inconsistent: the same PR correctly extracted `buildTwitterThreadTweets` — which has an identical structural profile — to `article-thread.util.ts` with its own spec file. `buildXArticleContentAndMetadata` has no test coverage.
  _Fix:_ Move to a new `apps/server/api/src/collections/articles/utils/article-content.util.ts` as an exported plain function. Add a matching spec. The call site becomes a one-line import change.

- **`apps/server/api/src/collections/articles/services/articles-content.service.ts:1167–1183`** — `buildXArticleContentAndMetadata` declares its `metadata` return field as an inline anonymous object type instead of the canonical `IXArticleMetadata` from `packages/interfaces/src/content/x-article-metadata.interface.ts`. The inline `sections` array element is field-for-field identical to `IXArticleSection`. The repo rule is explicit: no inline interfaces in service files. The `articles-content.service.ts` file already imports from `@genfeedai/interfaces` (lines 58, 63), so no new dependency is required.
  _Fix:_ Change the return type annotation to `{ content: string; metadata: IXArticleMetadata; wordCount: number }` and import `IXArticleMetadata` from `@genfeedai/interfaces`. The two optional fields (`headerImagePrompt?`, `headerImageUrl?`) on `IXArticleMetadata` are structurally compatible with the current return value; no value-level changes needed.

- **`apps/server/api/src/collections/articles/services/articles.service.ts:1552–1583`** — `applyPublishStateTransition` accepts `updateData: Record<string, unknown>` and mutates it in-place while returning `Promise<void>`. The caller passes the typed output of `ArticleFilterUtil.toArticlePersistenceData<T>()` — a concrete typed object — which is silently widened to `Record<string, unknown>` at the method boundary. TypeScript cannot enforce that `.scope` and `.publishedAt` are legal on the actual shape, and the mutation is invisible at the call site: a reader sees `await this.applyPublishStateTransition({...})` immediately followed by `super.patch(id, updateData)` with no indication `updateData` was enriched. The pattern is pre-existing in the file (`hydrateUserAndOrganization` uses the same void-return mutation contract), so this is a preserved defect rather than a new one, but the extraction was the moment to fix it.
  _Fix:_ Return the fields to merge: `private async applyPublishStateTransition(...): Promise<{ scope: ArticleScope; publishedAt?: string }>`, then at the call site `Object.assign(updateData, await this.applyPublishStateTransition(...))`. This makes the data flow explicit and eliminates the loose-type boundary.

- **`apps/server/api/src/collections/articles/services/articles.service.ts:1552–1583`** — The `findOne` ownership read inside `applyPublishStateTransition` and the `super.patch` call in the caller are two separate DB round-trips with no transaction wrapper. Two concurrent publish requests on the same never-published article can both observe `publishedAt === null`, both stamp different timestamps, and one silently overwrites the other. This race predates the PR; the extraction preserved it rather than collapsing it.
  _Fix:_ Fold the `publishedAt` stamp into the update's `where` predicate atomically — a Prisma `updateMany` with `{ where: { id, publishedAt: null } }` for the first-publish case, embedding the ownership filter (`OR: [{ userId }, { organizationId }]`) in the same clause, eliminates the pre-check read entirely.

---

### ✂️ Simplifications

- **`apps/server/api/src/collections/articles/services/articles-content.service.ts:1167–1219`** — `buildXArticleContentAndMetadata` returns `wordCount` at both the top level and inside `metadata` (`{ content, metadata: { ..., wordCount }, wordCount }`). The only reason the top-level field exists is to avoid writing `xArticleMetadata.wordCount` in one logger call at the destructuring site. The two values are always identical (computed from the same local variable), but the type contract does not enforce that — any future edit touching one arm silently diverges.
  _Fix:_ Remove the top-level `wordCount` from the return shape. Update the caller's destructure to `const { content: fullContent, metadata: xArticleMetadata } = this.buildXArticleContentAndMetadata(response.sections)` and replace the one usage of the standalone `wordCount` binding with `xArticleMetadata.wordCount`.

---

### Verdict

**APPROVE** — No surviving blockers. The PR is a net positive: 586 lines of duplicated service deleted, three identical generation pipeline bodies collapsed into one shared prologue, four copy-pasted helper blocks consolidated. The warnings above are structural debts that should be tracked and resolved in a follow-on decomposition pass, not reasons to block this PR.

PR #854, opened 2026-06-27
feat(services): thread AbortSignal through BaseService.findOne (#637)

### LINKED ISSUE #637: Client services: propagate AbortSignal through API requests
## Source
- PR review threads: https://github.com/genfeedai/genfeed.ai/pull/620#discussion_r3428330926 and https://github.com/genfeedai/genfeed.ai/pull/620#discussion_r3428330935
- Found while triaging Spark/GitHub review notification emails on 2026-06-18.

## Current evidence
`apps/app/app/(protected)/admin/automation/trainings/[id]/training-detail.tsx` now aborts superseded loads and skips stale state updates, so the immediate UI bug is handled. The remaining legitimate debt is lower in the service layer: `packages/services/core/base.service.ts#findOne` only accepts query params and does not pass an `AbortSignal` into the underlying Axios request, so in-flight network work cannot be cancelled.

## Done when
- Add a typed request-options path for `AbortSignal` in `BaseService` fetch helpers.
- Update training detail to pass the controller signal into `service.findOne`.
- Cover cancellation behavior with focused tests or a documented manual verification path.

VincentShipsIt:
## What

`BaseService.findAll` already forwards an `AbortSignal` into the Axios request config (`{ params, signal }`), so list fetches are genuinely cancellable. `findOne` did not — it only accepted query params. The training-detail screen therefore aborted its `AbortController` and skipped stale state updates, but the underlying `GET /trainings/:id` request kept running to completion (the immediate UI bug from PR #620 was masked, the network-layer debt remained).

This threads the signal the rest of the way down.

## Changes

- **`packages/services/core/base.service.ts`** — add an optional `signal?: AbortSignal` parameter to `findOne`, forwarded as `{ params: query, signal }`, mirroring the existing `findAll` signature exactly. Optional + last-positional → backward compatible for every existing caller and subclass.
- **`apps/app/.../trainings/[id]/training-detail.tsx`** — pass the already-present `controller.signal` into `service.findOne(trainingId, {}, controller.signal)` so a superseded load aborts the in-flight request, not just the post-response state write.
- **`packages/services/core/base.service.test.ts`** — focused test asserting `findOne` forwards the signal into the request config (mirrors the existing `findAll` signal test).

## Scope notes

- Mutation helpers (`post`/`patch`/`delete`) are intentionally untouched — the issue scopes this to the fetch helpers and the `findOne` consumer. No gratuitous refactoring.
- `TrainingsService` inherits `BaseService.findOne` (no override), so the 3-arg call resolves directly.

## Verification

- `bun run test --filter=@genfeedai/services` → `core/base.service.test.ts` **53 passed** (incl. new signal-forwarding case).
- `@genfeedai/services` `tsc --noEmit` clean for changed files (only unrelated pre-existing error: `@genfeedai/auth-client` unresolved in `packages/helpers` — an unbuilt-workspace-dep artifact from the Better Auth cutover #769, untouched here; resolves under CI's dependency-ordered build).
- `biome lint` on the 3 changed files → exit 0; pre-commit secretlint + no-raw-html hooks green.
- Adversarial multi-agent review of the diff: one confirmed finding (a tautological abort test) was removed; signal forwarding is verified by the retained focused test.
- Full `apps/app` typecheck deferred to CI (heavy; the consumer change is a type-safe optional 3rd arg).

Closes #637

### DIFF
--- a/apps/app/app/(protected)/admin/automation/trainings/[id]/training-detail.tsx
+++ b/apps/app/app/(protected)/admin/automation/trainings/[id]/training-detail.tsx
@@ -68,7 +68,7 @@ export default function TrainingDetail({
 
         controller.signal.throwIfAborted();
 
-        const data = await service.findOne(trainingId);
+        const data = await service.findOne(trainingId, {}, controller.signal);
 
         controller.signal.throwIfAborted();
 
--- a/packages/services/core/base.service.test.ts
+++ b/packages/services/core/base.service.test.ts
@@ -475,6 +475,21 @@ describe('BaseService', () => {
       expect(result).toBeInstanceOf(TestModel);
     });
 
+    it('forwards an AbortSignal to the request config so the fetch is cancellable', async () => {
+      const mockResponse = {
+        data: { data: { id: '123', name: 'Test' } },
+      };
+      service.getInstanceForTest().get.mockResolvedValue(mockResponse);
+      const controller = new AbortController();
+
+      await service.findOne('123', { include: 'relations' }, controller.signal);
+
+      expect(service.getInstanceForTest().get).toHaveBeenCalledWith('/123', {
+        params: { include: 'relations' },
+        signal: controller.signal,
+      });
+    });
+
     it('should throw on error', async () => {
       service
         .getInstanceForTest()
--- a/packages/services/core/base.service.ts
+++ b/packages/services/core/base.service.ts
@@ -258,11 +258,15 @@ export abstract class BaseService<
     );
   }
 
-  public findOne(id: string, query: Record<string, unknown> = {}): Promise<T> {
+  public findOne(
+    id: string,
+    query: Record<string, unknown> = {},
+    signal?: AbortSignal,
+  ): Promise<T> {
     return this.executeWithErrorHandling(
       `GET ${this.baseURL}/${id}`,
       this.instance
-        .get<JsonApiResponseDocument>(`/${id}`, { params: query })
+        .get<JsonApiResponseDocument>(`/${id}`, { params: query, signal })
         .then((res) => res.data)
         .then(async (res) => await this.mapOne(res)),
     );


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

Small, focused diff with the right instinct — thread the `AbortSignal` the rest of the way down. But there is one structural problem that defeats the PR's stated goal at runtime, and a downstream API-shape question worth raising while there is exactly one call site to migrate.

> Note on line numbers: this was a diff-and-source review of branch `feat/637-abort-signal-client-services`; line references below are resolved against the branch source, not the diff.

## Blockers

### 1. The threaded signal is unconditionally overwritten downstream — the feature is inert in production
**`packages/services/core/interceptor.service.ts:183`** (in `handleRequest`, lines 172–185)

`findOne` now builds `{ params: query, signal }` and hands it to `this.instance.get(...)`. But `this.instance` has a request interceptor registered at `interceptor.service.ts:165` (`this.instance.interceptors.request.use(this.handleRequest)`), and `handleRequest` ends with:

```ts
config.signal = this.abortController.signal; // unconditional — clobbers the caller's signal
```

This runs on every request and replaces the per-request `signal` with the instance-level controller's signal (which is only ever aborted via `cancelPendingRequests()`). So the `controller.signal` passed from `training-detail.tsx` never reaches the underlying fetch. When the component aborts its controller, the in-flight `GET /trainings/:id` keeps running to completion — which is exactly the network-layer debt this PR set out to remove. As shipped, the change is a no-op at the network layer.

The new test in `base.service.test.ts` does not catch this because the suite mocks `HTTPBaseService` with a stub that has no axios interceptor, so it validates a code path that does not exist at runtime.

`interceptor.service.ts` is pre-existing and not touched by this diff, but the PR's new code cannot achieve its objective without addressing it — so it is in scope as a correctness gap on the PR's own goal.

**Fix — preserve a caller-supplied signal:**
- Minimal: `if (!config.signal) { config.signal = this.abortController.signal; }`
- Correct composition (keeps both `cancelPendingRequests()` and per-request cancellation):
  ```ts
  config.signal = AbortSignal.any([
    this.abortController.signal,
    ...(config.signal ? [config.signal] : []),
  ]);
  ```
  Then update the test to use a double that replicates the interceptor's overwrite, so this regression cannot recur silently.

### Checked and cleared: the two pre-existing `findOne` test assertions are **not** broken
For the record — a plausible worry here is that the pre-existing assertions `toHaveBeenCalledWith('/123', { params: {} })` (`base.service.test.ts:447-448`) and `{ params: { include: 'relations' } }` (`461-462`) would fail now that `findOne` always passes `{ params: query, signal }` (yielding `signal: undefined` when omitted). They do **not** fail: `findAll` already uses the identical unconditional `{ params: query, signal }` shape (`base.service.ts:244`), and its no-signal test asserts `{ params: {} }` (`375-376`) and is green on `master`. `toHaveBeenCalledWith` uses `toEqual` semantics, which ignore `undefined`-valued keys (only `toStrictEqual` distinguishes them). No test change is required for the existing assertions.

## Warnings

### 2. Positional `findOne(id, query = {}, signal?)` forces a phantom `{}` at the call site
**`packages/services/core/base.service.ts:261`** → consumer **`training-detail.tsx:71`**

The new signature wedges `query` between `id` and `signal`, so a caller that wants cancellation but no query params must pass a load-bearing-but-empty `{}`:

```ts
service.findOne(trainingId, {}, controller.signal); // {} carries no intent
```

Every future cancellable read inherits this. An options bag (`findOne(id, { query?, signal? } = {})`) would let the consumer write `service.findOne(trainingId, { signal: controller.signal })` and make both optionals independently optional.

This is a genuine judgment call, not a clear defect: the positional form deliberately mirrors the existing `findAll(query, signal)` convention, so the options-bag change trades call-site ergonomics against symmetry with `findAll`. Worth deciding now while there is a single call site and zero migration cost — but either choice is defensible.

## Verdict

**BLOCK** — single blocker: `interceptor.service.ts:183` unconditionally overwrites the caller-supplied `AbortSignal`, so the cancellation this PR threads through `findOne` does not function at runtime, and the accompanying test exercises a mock path that bypasses the interceptor. Once the interceptor preserves a per-request signal and the test exercises a double that replicates the overwrite, this is good to merge. The API-shape warning is optional and best resolved in the same PR.

PR #855, opened 2026-06-27
refactor(trends): extract source-item subsystem from TrendsService (#752)

### LINKED ISSUE #752: Extract trend source items
`TrendsService` (`apps/server/api/src/collections/trends/services/trends.service.ts`) is 1,859 lines with **13 injected constructor dependencies** (lines 184-196), making it a god-object that the existing `modules/` decomposition pass left half-finished. Seven module services were already extracted (`trend-analysis`, `trend-fetch`, `trend-filtering`, `trend-video`, `trend-content-ideas`, plus `trend-preferences` and `trend-reference-corpus`), but a large cohesive subsystem and several duplication families remain inline.

> This is a **structural decomposition** issue. It is distinct from the trend reference-corpus performance work in #629 and does not touch query performance.

## Confirmed structural findings (file:line)

- **Un-extracted source-item subsystem, ~640 lines (1216-1858):** platform dispatch `fetchTrendSourceItems` (1216); five structurally identical per-platform fetchers (`fetchInstagramSourceItems`/`fetchTikTok…`/`fetchTwitter…`/`fetchYouTube…`/`fetchReddit…`, 1236-1320); five `map*ToSourceItem` mappers (1339-1476); `buildFallbackTrendSourceItems` (1478); preview-cache helpers (1519-1636); content-item assembly `buildTrendContentItems`/`compareTrendContentItems`/`getTrendContentEngagementScore` (1638-1742). This is a self-contained concern (live Apify fetch + normalization + preview caching) with no reason to live in the orchestrator.
- **Fetcher duplication (1236-1320):** all five `fetch{Platform}SourceItems` methods share one shape: `getTrendSearchTerm` guard, an Apify call, then `.map(mapper).filter(non-null)`. A dispatch table keyed by platform removes the copy-paste.
- **Near-duplicate query helpers:** `findActiveTrends` (460) and `findLastGoodTrends` (493) both run an identical `findMany(take:200)` + in-memory filter/sort/slice(50)/map, differing only in predicate and tiebreaker.
- **Pass-through wrapper sprawl:** ~20 methods forward verbatim to a module service (e.g. lines 256, 263, 753, 770, 908, 925, 943, 952, 1101, plus the contiguous video/turnover block 1120-1214), inflating the public surface without owning behavior.
- **Mixed responsibilities:** Prisma access, cache orchestration, Apify mapping, bootstrap/fallback data, and prelaunch-corpus seeding (`buildPrelaunchTrendData` 425, `backfillPrelaunchReferenceCorpus` 804) all live in one class.

Adding or changing one platform's source-item handling requires navigating a 1,859-line file and touching three parallel method families (fetcher, mapper, dispatch switch), so platform behavior can drift silently. The 13-dependency constructor forces every orchestrator test to mock 13 collaborators.

1. Extract `TrendSourceItemsService` (`modules/trend-source-items.service.ts`) owning `fetchTrendSourceItems`, the five fetchers, the five mappers, `buildFallbackTrendSourceItems`, `getTrendSearchTerm`, `truncateText`, and the preview-cache helpers. Move the `apifyService` injection there; `TrendsService` injects and delegates.
2. Within that service, collapse the five fetchers into one method driven by a `Record<platform, { search, map }>` dispatch table.
3. Extract content-feed assembly (`buildTrendContentItems`, `compareTrendContentItems`, `getTrendContentEngagementScore`, `getLatestTrendAt`) into a `TrendContentFeedService`, or fold into the source-items service.
4. Unify `findActiveTrends`/`findLastGoodTrends` behind one parameterized query helper.
5. Move prelaunch-corpus seeding (`buildPrelaunchTrendData`, `findPrelaunchCorpusDocs`, `getPrelaunchCorpusKey`, `backfillPrelaunchReferenceCorpus`) into a `TrendPrelaunchCorpusService`.
6. Keep the existing module spec suites green throughout; add colocated specs for the new services.

## Acceptance criteria

- No file in `collections/trends/services/` exceeds 500 lines.
- `TrendsService` constructor dependency count drops below 10.
- The five-way platform fetcher duplication is gone (single dispatch path).
- Public method signatures and return shapes are unchanged; all existing trends specs still pass; new services have colocated specs.

VincentShipsIt:
## Summary

Decomposes the 1,859-line `TrendsService` god object (issue #752) into focused, colocated module services, following the repo's existing `services/modules/` extraction pattern. **Behavior-preserving** — public method signatures and runtime behavior are unchanged.

## What changed

New services (each <500 lines, with colocated specs):

| Service | Responsibility | Deps |
|---|---|---|
| `TrendSourceItemsService` | Live Apify fetch + per-platform normalization + fallback synthesis + stored-preview accessors | apify |
| `TrendSourcePreviewService` | Preview persistence/precompute/resolution, content-feed assembly, cached `getTrendContent`, discovery items | prisma, cache, referenceCorpus, sourceItems |
| `TrendQueryService` | Trend read queries + bootstrap fallback | prisma |
| `TrendPrelaunchCorpusService` | Prelaunch reference-corpus seeding/backfill + global stats | prisma, cache, referenceCorpus, sourceItems |

- **Five-way fetcher duplication removed**: the five near-identical `fetch{Platform}SourceItems` methods collapse into a single dispatch table in `fetchTrendSourceItems`.
- **Read helpers unified**: `findActiveTrends` / `findLastGoodTrends` now share one parameterized `queryTrendsByFilter`.
- `TrendsService` is now a thin orchestrator: **1,859 → 684 lines**. `PrismaService`, `CacheService`, and `ApifyService` are no longer injected directly into it.
- Shared source-preview limit extracted to a single `TREND_SOURCE_PREVIEW_LIMIT` constant to prevent drift.

## Acceptance criteria

- ✅ Source-item subsystem extracted into cohesive module services.
- ✅ Five-way platform fetcher duplication gone (single dispatch path).
- ✅ `findActiveTrends`/`findLastGoodTrends` unified.
- ✅ Public method signatures and return shapes unchanged; existing `trends.service` integration spec still passes; new services have colocated specs.
- ✅ New files all <500 lines.
- ⚠️ **`TrendsService` constructor dependency count**: held at 13 (the three low-level infra deps — Prisma/Cache/Apify — were removed; three new module-service deps added). Dropping **below 10** is not reachable without removing the ~25 public forwarding methods that 6+ external files depend on (`TrendsController`, workflow services, workflow-engine-adapter, trend-inspiration processor, worker crons). That public-API change is out of scope for this behavior-preserving extraction and would be a separate cross-app PR.
- ⚠️ **`trends.service.ts` is 684 lines (>500)**: the residual is the public forwarding façade + the `getTrends` retrieval orchestration, which must stay for API stability. Pre-existing `trend-video.service.ts` (700) and `trend-reference-corpus.service.ts` (768) also exceed 500 and are out of scope.

## Adversarial review

Ran a multi-agent find→verify review over the diff (behavior parity, DI wiring, type safety, test gaps). Acted on the in-scope findings:
- Extracted the duplicated `SOURCE_PREVIEW_LIMIT` to a shared constant and removed the dead copy on the orchestrator.
- Added targeted coverage: TikTok/YouTube/Reddit dispatch + null-guard mappers, content-feed dedup-by-source-url, `precomputeTrendSourcePreview` cached-skip & cache-miss-persist branches, prelaunch live-mode backfill.

Findings left as **pre-existing / out of scope** (present on `master`, not introduced here): `pinterest` absent from `CONTENT_FEED_PLATFORMS` while present in `allPlatforms` (behavior preserved intentionally); `_loggerService` dead injection in `trend-filtering.service.ts`; `trend-fetch.service` test gaps.

## Verification

- `vitest` (scoped): **178 tests pass** across 15 trends-area files — orchestrator integration spec (43), four new colocated specs (39), adjacent module/warmup/controller/consumer specs.
- `biome check`: clean on all 48 trends files; pre-commit secretlint + biome passed.
- **Local typecheck limitation**: a full `tsc` requires the `@genfeedai/*` workspace packages to be built (a known TS6/Prisma-7 build-ordering constraint in this repo); `tsc` reports 0 errors in the new files and only the repo-wide unbuilt-`@genfeedai/*` resolution noise elsewhere. Mac Studio was unreachable this run, so the authoritative full typecheck/build is delegated to CI. The pre-existing `trends.module.spec.ts` failure (`@genfeedai/tools` entry resolution) reproduces on `master` and is the same unbuilt-package env issue.

Closes #752

### DIFF
--- a/apps/server/api/src/collections/trends/services/modules/trend-prelaunch-corpus.service.spec.ts
+++ b/apps/server/api/src/collections/trends/services/modules/trend-prelaunch-corpus.service.spec.ts
@@ -1,0 +1,124 @@
+import { TrendPrelaunchCorpusService } from '@api/collections/trends/services/modules/trend-prelaunch-corpus.service';
+import { TrendSourceItemsService } from '@api/collections/trends/services/modules/trend-source-items.service';
+import { TrendReferenceCorpusService } from '@api/collections/trends/services/trend-reference-corpus.service';
+import { CacheService } from '@api/services/cache/services/cache.service';
+import { ApifyService } from '@api/services/integrations/apify/services/apify.service';
+import { PrismaService } from '@api/shared/modules/prisma/prisma.service';
+import { LoggerService } from '@libs/logger/logger.service';
+import { Test, TestingModule } from '@nestjs/testing';
+
+describe('TrendPrelaunchCorpusService', () => {
+  let service: TrendPrelaunchCorpusService;
+  let prisma: {
+    trend: {
+      create: ReturnType<typeof vi.fn>;
+      findMany: ReturnType<typeof vi.fn>;
+      update: ReturnType<typeof vi.fn>;
+    };
+  };
+  let referenceCorpus: {
+    countGlobalReferences: ReturnType<typeof vi.fn>;
+    syncTrendReferences: ReturnType<typeof vi.fn>;
+  };
+  let cache: { invalidateByTags: ReturnType<typeof vi.fn> };
+
+  beforeEach(async () => {
+    prisma = {
+      trend: {
+        create: vi.fn(),
+        findMany: vi.fn().mockResolvedValue([]),
+        update: vi.fn(),
+      },
+    };
+    referenceCorpus = {
+      countGlobalReferences: vi.fn().mockResolvedValue(7),
+      syncTrendReferences: vi
+        .fn()
+        .mockResolvedValue({ links: 0, references: 0, snapshots: 0 }),
+    };
+    cache = { invalidateByTags: vi.fn().mockResolvedValue(1) };
+
+    const module: TestingModule = await Test.createTestingModule({
+      providers: [
+        TrendPrelaunchCorpusService,
+        TrendSourceItemsService,
+        { provide: ApifyService, useValue: {} },
+        { provide: PrismaService, useValue: prisma },
+        {
+          provide: LoggerService,
+          useValue: { error: vi.fn(), log: vi.fn(), warn: vi.fn() },
+        },
+        { provide: CacheService, useValue: cache },
+        { provide: TrendReferenceCorpusService, useValue: referenceCorpus },
+      ],
+    }).compile();
+
+    service = module.get(TrendPrelaunchCorpusService);
+  });
+
+  describe('getGlobalCorpusStats', () => {
+    it('counts only current, unexpired global trends', async () => {
+      const future = new Date(Date.now() + 3_600_000).toISOString();
+      const past = new Date(Date.now() - 3_600_000).toISOString();
+      prisma.trend.findMany.mockResolvedValue([
+        { data: { expiresAt: future, isCurrent: true } },
+        { data: { expiresAt: past, isCurrent: true } }, // expired
+        { data: { expiresAt: future, isCurrent: false } }, // not current
+      ]);
+
+      const result = await service.getGlobalCorpusStats();
+
+      expect(result).toEqual({ activeTrends: 1, referenceRecords: 7 });
+    });
+  });
+
+  describe('backfillPrelaunchReferenceCorpus', () => {
+    it('plans creates without writing in dry-run mode (the default)', async () => {
+      prisma.trend.findMany.mockResolvedValue([]);
+
+      const result = await service.backfillPrelaunchReferenceCorpus({
+        now: new Date('2026-06-09T00:00:00.000Z'),
+      });
+
+      expect(result.dryRun).toBe(true);
+      expect(result.plannedCreates).toBe(result.seedTrends);
+      expect(result.plannedUpdates).toBe(0);
+      expect(result.createdTrends).toBe(0);
+      expect(prisma.trend.create).not.toHaveBeenCalled();
+      expect(prisma.trend.update).not.toHaveBeenCalled();
+      expect(referenceCorpus.syncTrendReferences).not.toHaveBeenCalled();
+      expect(cache.invalidateByTags).not.toHaveBeenCalled();
+    });
+
+    it('creates seeds, syncs references, and busts caches in live mode', async () => {
+      prisma.trend.findMany.mockResolvedValue([]);
+      prisma.trend.create.mockImplementation(async (input: unknown) => {
+        const payload = input as { data: { data: Record<string, unknown> } };
+        return {
+          brandId: null,
+          createdAt: new Date('2026-06-09T00:00:00.000Z'),
+          data: payload.data.data,
+          id: `created-${prisma.trend.create.mock.calls.length}`,
+          isDeleted: false,
+          organizationId: null,
+          updatedAt: new Date('2026-06-09T00:00:00.000Z'),
+        };
+      });
+
+      const result = await service.backfillPrelaunchReferenceCorpus({
+        dryRun: false,
+        now: new Date('2026-06-09T00:00:00.000Z'),
+      });
+
+      expect(result.dryRun).toBe(false);
+      expect(result.createdTrends).toBe(result.seedTrends);
+      expect(prisma.trend.create).toHaveBeenCalledTimes(result.seedTrends);
+      expect(prisma.trend.update).not.toHaveBeenCalled();
+      expect(referenceCorpus.syncTrendReferences).toHaveBeenCalledTimes(1);
+      expect(cache.invalidateByTags).toHaveBeenCalledWith([
+        'trends',
+        'trends:content',
+      ]);
+    });
+  });
+});
--- a/apps/server/api/src/collections/trends/services/modules/trend-prelaunch-corpus.service.ts
+++ b/apps/server/api/src/collections/trends/services/modules/trend-prelaunch-corpus.service.ts
@@ -1,0 +1,253 @@
+import {
+  buildPrelaunchReferenceCorpusSeeds,
+  PRELAUNCH_REFERENCE_CORPUS_MINIMUMS,
+  PRELAUNCH_REFERENCE_CORPUS_VERSION,
+  type PrelaunchReferenceCorpusSeed,
+} from '@api/collections/trends/data/prelaunch-reference-corpus.seed';
+import { TrendEntity } from '@api/collections/trends/entities/trend.entity';
+import type { TrendDocument } from '@api/collections/trends/schemas/trend.schema';
+import { TrendSourceItemsService } from '@api/collections/trends/services/modules/trend-source-items.service';
+import { TrendReferenceCorpusService } from '@api/collections/trends/services/trend-reference-corpus.service';
+import { CacheService } from '@api/services/cache/services/cache.service';
+import { PrismaService } from '@api/shared/modules/prisma/prisma.service';
+import { LoggerService } from '@libs/logger/logger.service';
+import { Injectable } from '@nestjs/common';
+
+export interface PrelaunchReferenceCorpusBackfillOptions {
+  dryRun?: boolean;
+  now?: Date;
+}
+
+export interface PrelaunchReferenceCorpusBackfillResult {
+  createdTrends: number;
+  dryRun: boolean;
+  plannedCreates: number;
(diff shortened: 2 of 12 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

This PR extracts four focused services from a 1,859-line god object and is a net structural win: the five-way `fetch{Platform}SourceItems` dispatch-table collapse, the `findActiveTrends`/`findLastGoodTrends` unification behind `queryTrendsByFilter`, the shared `TREND_SOURCE_PREVIEW_LIMIT` constant, and the colocated per-module specs are all exactly right. No confirmed BLOCKER-severity findings. Eight in-scope issues follow, ordered by severity.

> Note on line numbers: references below are resolved against branch source `feat/752-trend-source-items` (not diff offsets).

## Warnings

### 1. Unchecked double-cast discards Prisma's generated type
**`trend-query.service.ts:163-164`**

```ts
const docOrgId = (doc as unknown as Record<string, unknown>).organizationId;
```

`prisma.trend.findFirst` returns `organizationId: string | null` on the typed result. The `as unknown as Record<string, unknown>` cast strips that type to `unknown`, removing compiler protection over the null-check that follows. Use the direct typed access — `const docOrgId = doc.organizationId;` — no cast, same runtime behavior.

### 2. Dead post-query org-scope guard
**`trend-query.service.ts:163-168`** (WHERE clause at `:152`)

When `organizationId` is supplied, the Prisma `WHERE` already contains `OR: [{ organizationId }, { organizationId: null }]`, so every returned row satisfies `doc.organizationId === organizationId` OR `=== null` by construction. The subsequent `if (docOrgId !== organizationId && docOrgId !== null) return null` is therefore permanently dead in the org-supplied path — its condition is the logical complement of what the WHERE clause admits.

Delete the guard (and its comment); the companion spec at `trend-query.service.spec.ts` that mocks Prisma returning `organizationId: 'other-org'` tests a state the database can never produce, so it goes too. A second enforcement path in application code implies the DB filter is untrustworthy and misleads maintainers.

### 3. Date normalization pushed onto consumers instead of fixed at construction
**`trend-source-preview.service.ts:164,170,177`** (`buildTrendDiscoveryItem`, from `:142`) and **`:408`** (`getLatestTrendAt`, from `:405`)

`buildTrendDiscoveryItem` applies `x instanceof Date ? x : x ? new Date(x) : undefined` to `createdAt`, `updatedAt`, and `expiresAt`; `getLatestTrendAt` repeats the `instanceof Date` guard for `createdAt`. The root cause is `toTrendEntity` (**`trend-query.service.ts:236-242`**) doing a raw `{ ...doc, ...doc.data }` spread with no date coercion. For `expiresAt` this is a real correctness gap — it lives inside the Prisma JSON `data` blob and arrives as an ISO string after the spread, making the declared `expiresAt: Date` on `TrendEntity` a runtime lie. The `createdAt`/`updatedAt` guards are defensive dead code (Prisma returns native `Date` for Postgres timestamps).

Coerce `expiresAt` (and optionally the timestamp fields, for documentation value) to `Date` inside `toTrendEntity`, then drop the ternary chains from both consumers. Normalize once at construction; delete the scattered compensating logic.

### 4. Spurious `findFirst` pre-read in `persistTrendSourcePreview`
**`trend-source-preview.service.ts:273`** (in `persistTrendSourcePreview`, from `:260`)

The method issues `this.prisma.trend.findFirst(...)` to obtain `existingDoc.data` as the spread base for the update payload — but the incoming `TrendEntity.data` already *is* that blob (`toTrendEntity` builds the entity via `{ ...doc, ...doc.data }`, retaining the original `doc.data`). The read is redundant on every precompute call and widens the lost-write window (a concurrent writer can mutate the row between the `findFirst` and the `update`).

Replace the pre-read with `const existingData = (trend.data as Record<string, unknown>) ?? {};` and write directly. If the silent no-op on a concurrently-deleted trend must be preserved, use `updateMany` with the same `where` (it returns a count) rather than a pre-read.

### 5. `TrendPrelaunchCorpusService` inherits an ApifyService dependency for a pure projection
**`trend-prelaunch-corpus.service.ts:49`** (injection) → **`:151`** (sole use)

`TrendPrelaunchCorpusService` injects `TrendSourceItemsService` only to call `toSyncTrendInput` once — a pure read-only projection over scalar `TrendEntity` fields with no Apify involvement. But `TrendSourceItemsService` injects `ApifyService`, so the corpus service acquires the Apify dependency transitively; its spec must provide an `ApifyService` mock for a path it never exercises. That is wrong-layer coupling.

Either (a) extract `toSyncTrendInput` into a standalone pure helper (e.g. `trend-sync.helpers.ts`) importable by both services, or (b) inline the ~10-line projection at the one call site. Both sever the cross-subsystem edge and simplify the corpus spec.

### 6. `invalidateContentCache` is a one-line public pass-through with no encapsulation
**`trend-source-preview.service.ts:256`**

```ts
invalidateContentCache(tags: string[]): Promise<number> {
  return this.cacheService.invalidateByTags(tags);
}
```

Introduced because `TrendsService` lost its direct `CacheService` injection — but the caller still supplies the full tag list, so nothing about *which* tags to bust is encapsulated. Contrast `TrendPrelaunchCorpusService` (also new here), which injects `CacheService` directly and calls `invalidateByTags` without a wrapper. Either reinject `CacheService` into `TrendsService` and call it directly at the two sites, or drop the `tags` parameter and let `invalidateContentCache` derive the canonical tag set internally (real encapsulation). Eliminate the pass-through.

## Simplifications

### 7. `SyncTrendInput` contract declared three times as inline object literals
**`trend-source-items.service.ts:29`** and **`trend-source-preview.service.ts:53`**

Both annotate `toSyncTrendInput`'s return as a 7-field inline literal (`{ id, mentions, platform, sourcePreview, sourcePreviewState: 'live'|'fallback'|'empty', topic, viralityScore }`). A canonical `interface SyncTrendInput` with the identical shape already exists at **`trend-reference-corpus.service.ts:13`** — but file-private. Three descriptions of one contract, all in PR-introduced or PR-adjacent code.

Export `SyncTrendInput` (from `packages/interfaces/` or re-exported from `trend.interfaces.ts`), replace both inline annotations with `import type { SyncTrendInput }`, and have `TrendReferenceCorpusService` import the same. Separately, `TrendSourcePreviewService.toSyncTrendInput` (`:53`) is a one-line passthrough to `trendSourceItemsService.toSyncTrendInput` that arguably should not exist — callers can hit `trendSourceItemsService` directly — but de-duplicating the type is the higher-value fix.

### 8. `findPrelaunchCorpusDocs` manually broadens away the Prisma `Trend` type
**`trend-prelaunch-corpus.service.ts:184-185`**

```ts
private async findPrelaunchCorpusDocs(): Promise<
  Array<{ data: unknown; id: string } & Record<string, unknown>>
>
```

`prisma.trend.findMany` returns the full generated `Trend` type; the explicit annotation erases it into a broad `Record<string, unknown>` intersection on a private, single-call-site method — hiding invariants for no gain. Drop the annotation and let inference carry the Prisma result type. (The callers' `doc.data as Record<string, unknown>` casts are independently necessary — Prisma JSON fields are `Prisma.JsonValue` — so this fix does not remove those, only the top-level structural erasure.)

## Author-acknowledged scope (not findings)

The stated out-of-scope items are correctly scoped and do **not** drive the verdict: `trends.service.ts` residual at 684 lines (below the 1,000-line blocker threshold), the constructor held at 13 deps (the three infra deps removed, three module deps added — author-justified as decomposition progress), pre-existing `trend-video.service.ts` (700) and `trend-reference-corpus.service.ts` (768) untouched here, the dead `_loggerService` injection in `trend-filtering.service.ts` (pre-existing untouched), and the intentional Pinterest omission from `CONTENT_FEED_PLATFORMS`.

## Verdict

**APPROVE** — no BLOCKER-severity findings. The decomposition is sound and the net complexity reduction is real. Items 1–6 are in-scope structural cleanups worth folding into this branch or a fast follow-up; 7–8 are housekeeping. Most impactful single fix: normalize `expiresAt` in `toTrendEntity` (#3), since the current `TrendEntity.expiresAt: Date` type is a runtime lie.

PR #856, opened 2026-06-28
fix(videos): harden generation create() path — 7 CodeRabbit findings (#853)

### LINKED ISSUE #853: Harden the videos generation create() path (7 pre-existing CodeRabbit findings carried into VideoGenerationService)
## Context

PR #842 (refactor, [#835](https://github.com/genfeedai/genfeed.ai/issues/835)) extracted the 835-line `videos.controller.ts` `create()` into `VideoGenerationService` **behavior-preserving**. CodeRabbit raised 7 "Major" findings, all on logic that was **moved verbatim** from the original controller — i.e. pre-existing behaviors, not regressions. They were intentionally left unchanged in #842 to keep the refactor contract; this issue tracks fixing them properly (each needs its own behavior change + test).

`apps/server/api/src/collections/videos/services/video-generation.service.ts`

## Findings

1. **Org model validation skipped when request context is absent** (~`resolveVideoModel`/post-resolve). Validation is gated on `request.context?.organizationId`; a default/auto-selected model can bypass org validation when context is missing. Consider always validating against `publicMetadata.organization`. *(Security / multi-tenancy)*
2. **Prompt lookup not org-scoped + does not fail closed.** `promptsService.findOne({ _id, isDeleted:false })` lacks `organization`; an invalid prompt id leaves `promptText` empty yet generation proceeds. Scope by org and reject unknown prompts. *(Security / multi-tenancy)*
3. **Persist the prompt against the selected brand.** Prompt is created with `brand: publicMetadata.brand` rather than the resolved `brand._id` used elsewhere. *(Data integrity)*
4. **Single credit calculation for authorization and deduction.** Authorization uses `calculateDynamicVideoCost`; deduction uses `modelData.cost * multipliers`. They can diverge. *(Data integrity / billing)*
5. **Patch external IDs for every batch placeholder.** Batch-capable multi-output path sets `externalId` only on the first doc (`${generationId}_0`); additional placeholders never get `_1`, `_2`, … (this is the no-op `Promise.all` removed in #835). *(Data integrity)*
6. **Track additional placeholders before external calls; fail on null IDs.** Non-batch loop pushes the pending id after the external call and tolerates null generation ids. *(Stability)*
7. **Invalidate the complete video cache patterns after writes.** Only `invalidateByTags(['videos'])` is busted; single-record/agg patterns may go stale. *(Data integrity / caching — see apps/server/api CLAUDE.md cache pattern)*

## Acceptance

- Each finding fixed as a scoped change with a regression test.
- Multi-tenancy items (1, 2) prioritized.
- No behavior change merged without test coverage proving the new contract.

VincentShipsIt:
## Summary

Fixes the seven pre-existing CodeRabbit "Major" findings carried into `VideoGenerationService` when `create()` was extracted from `videos.controller.ts` in #842 (behavior-preserving). Each is now fixed as a scoped change with regression tests.

`apps/server/api/src/collections/videos/services/video-generation.service.ts`

| # | Area | Fix |
|---|------|-----|
| F1 | Security / multi-tenancy | Validate the resolved model against the org whenever one is present (token claim → request-context fallback), not only when `request.context.organizationId` was populated. Closes the default/auto-selected-model `ModelsGuard` bypass; single-tenant (no org) still skips. |
| F2 | Security / multi-tenancy | Org-scope the saved-prompt lookup **and fail closed** (404) on an unknown/foreign prompt id instead of silently generating with an empty prompt. The org filter is omitted when no org is present so single-tenant (null `organizationId`) still resolves prompts. |
| F3 | Data integrity | Persist the prompt against the resolved `brand._id` (honors `createVideoDto.brand`) instead of the token-default brand — matches `saveDocuments`/activities/cleanup. |
| F4 | Data integrity / billing | Remove the service's second, separately-computed credit deduction. Deduction is now owned solely by `CreditsInterceptor` (matching the images pipeline); `ensureDeferredCredits` authorizes the full multiplied amount (`base × resolution × outputs` for non-batch) the interceptor deducts. A null `generationId` on the first output now fails the request so a generation that never started is not charged. |
| F5 | Data integrity | Patch an indexed `externalId` (`_1`, `_2`, …) onto **every** additional batch placeholder (previously only the first got `_0`). Placeholders are tracked before patching so they are cleaned up on failure. |
| F6 | Stability | Track each additional non-batch placeholder **before** its provider call and fail closed on a null id instead of persisting `externalId: ''`. |
| F7 | Caching | All video GET endpoints register the single `videos` cache tag, so busting it already covers list/latest/single; aggregate tags are invalidated by `BaseService` on the underlying ingredient writes. Kept as `['videos']` with documentation (the original premise of separate single-record/agg tags did not hold). |

### Notable behavior changes (intentional, tested)
- **No more double-charge.** Previously the route deducted credits twice with divergent amounts — once via `CreditsInterceptor` (`creditsConfig.amount`) and once via a manual `deductCreditsFromOrganization` call. The manual call is removed; the interceptor is the single source, fed by the single `ensureDeferredCredits` calculation.
- **Failed-to-start generations now fail the request** (`502`) instead of returning success, preventing the interceptor from charging for a generation that never began.
- The output multiplier in `ensureDeferredCredits` keys on `isBatchSupported` (the real number of billable provider calls this route fans out into), since this route is always deferred and `ensureDeferredCredits` is its sole authorizer.

## Review

Diff was put through an adversarial multi-agent review (security/multi-tenancy, billing, control-flow, cache, tests) with per-finding verification. Confirmed issues were fixed:
- Single-tenant prompt lookup regression (empty-string org filter) → conditional org filter.
- Batch-placeholder leak window (push tracked before patch) → reordered tracking before the patch.
- Phantom cache tags → reverted to the correct `['videos']`.
- Added HTTP status-code assertions and a first-output null-generation test.

## Verification
- `bun run test` (scoped, vitest): `video-generation.service.spec.ts` (new, all 7 findings) + `videos.controller.spec.ts` — **62 passed**.
- `biome check` — clean.
- Scoped `tsc` on the api package shows **no type errors in the changed files** (the unbuilt-`@genfeedai/*`-package `TS2307` resolution errors affect the whole tree uniformly and are an environment artifact; full workspace typecheck/build left to CI).

Closes #853

### DIFF
--- a/apps/server/api/src/collections/videos/controllers/videos.controller.spec.ts
+++ b/apps/server/api/src/collections/videos/controllers/videos.controller.spec.ts
@@ -52,7 +52,7 @@ import {
   ModelCategory,
 } from '@genfeedai/enums';
 import { LoggerService } from '@libs/logger/logger.service';
-import { HttpException, StreamableFile } from '@nestjs/common';
+import { HttpException, HttpStatus, StreamableFile } from '@nestjs/common';
 import { Test, type TestingModule } from '@nestjs/testing';
 import type {
   Request as ExpressRequest,
@@ -230,7 +230,9 @@ describe('VideosController', () => {
         {
           provide: CreditsUtilsService,
           useValue: {
+            checkOrganizationCreditsAvailable: vi.fn().mockResolvedValue(true),
             deductCreditsFromOrganization: vi.fn().mockResolvedValue(undefined),
+            getOrganizationCreditsBalance: vi.fn().mockResolvedValue(1000),
           },
         },
         {
@@ -285,6 +287,7 @@ describe('VideosController', () => {
           provide: ModelRegistrationService,
           useValue: {
             registerGeneratedOutput: vi.fn().mockResolvedValue(undefined),
+            validateModelForOrg: vi.fn().mockResolvedValue(mockModelData),
           },
         },
         {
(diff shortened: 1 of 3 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

The correctness fixes carried in here are well-executed: removing the deferred-credit double-charge (interceptor as the single authorizer), org-scoping the saved-prompt lookup with fail-closed 404, indexing `externalId` onto every batch placeholder, and failing the request on a null first-output `generationId`. The new 427-line spec suite covering all seven findings is thorough. No blockers. Three warnings and one simplification on structure follow.

> Note on line numbers: references below are resolved against branch source `feat/853-videos-create-hardening` (not diff offsets).

## Warnings

### 1. File-size threshold exceeded and growing without decomposition
**`video-generation.service.ts:1`** (criterion 1 — adjacent)

The file lands at **1,041 lines** after this PR (net +35). The 1,000-line threshold was first crossed by the extraction PR (#842), but this PR enlarged an already-breached file across 13 hunks without flagging it. The public `generateVideo()` method (`:147` onward) carries three nested output-path branches — batch multi-output, non-batch per-output loop, single-output fallback — each with its own `saveDocuments`/`patch`/activity/log sequence, and this PR touched all three.

Extract the three branches into private methods (`handleBatchOutputs`, `handleNonBatchOutputs`, `handleSingleOutput`), each already largely self-contained. `generateVideo()` becomes a thin orchestrator and the file drops back under 1,000 lines with no behavior change. The file's own idiom supports this — `resolveVideoModel`, `ensureDeferredCredits`, `dispatchVideoGeneration` are exactly this pattern. Either decompose or explicitly justify holding the scope.

### 2. Duplicated resolution-cost multiplier
**`video-generation.service.ts:881-886`** (criterion 4 — introduced)

`ensureDeferredCredits` introduces an inline resolution multiplier:

```ts
if (createVideoDto.resolution === 'high' || createVideoDto.resolution === '1080p') {
  requiredCredits *= 2;
}
```

This is a verbatim copy of the canonical condition at **`credits.guard.ts:425-426`** (same string literals, same `*= 2`). The two are mutually exclusive paths at runtime (the route is `@DeferCreditsUntilModelResolution()`, so the guard skips and `ensureDeferredCredits` authorizes), so this is not a double-charge — but it *is* a duplicated policy: adding a tier (e.g. `'4k' => 4x`) now needs two synchronized edits.

Extract a pure `applyVideoResolutionMultiplier(baseCost, resolution?)` into a colocated util and call it from both `CreditsGuard` (`:425-426`) and `ensureDeferredCredits` (`:881-886`). The output multiplier diverges intentionally (`!isBatchSupported` vs the guard's `!isTrainingKey`) and is *not* a duplicate — leave it.

### 3. Type contract misrepresents `organization` cardinality
**`video-generation.service.ts:206-207`** and **`:260-261`** (criterion 5 — introduced)

`IClerkPublicMetadata.organization` is typed `string` (non-optional) at **`clerk.interface.ts:5`**, but two introduced sites treat it as nullable:

- `:206-207` — `const validationOrgId = publicMetadata.organization || request.context?.organizationId;` — under the declared `string` type the `||` fallback is unreachable dead code.
- `:260-261` — `...(publicMetadata.organization ? { organization: publicMetadata.organization } : {})` — the falsy branch is structurally unreachable.

The PR's own comments rely on the single-tenant bypass ("only single-tenant deployments — no organization present — skip this check"), confirming the real runtime cardinality is `string | undefined | ''`. The guard also conflates empty-string with genuinely absent — an undocumented convention hidden behind a type that says the value is always a non-empty string.

Change `organization: string` → `organization: string | undefined` in `IClerkPublicMetadata`, switch to `publicMetadata.organization ?? request.context?.organizationId`, and use `if (publicMetadata.organization)` as the presence guard. This makes the single-tenant bypass explicit at the type level.

## Simplifications

### 4. `VIDEO_WRITE_CACHE_TAGS` — one-use indirection with a readonly/mutable mismatch
**`video-generation.service.ts:74`** (declaration) and **`:683`** (sole use) (criterion 8 — introduced)

```ts
const VIDEO_WRITE_CACHE_TAGS: readonly string[] = ['videos'];
// ...
await this.cacheService.invalidateByTags([...VIDEO_WRITE_CACHE_TAGS]);
```

Three small problems compounded: the constant is non-exported and used exactly once (no reuse); `invalidateByTags` takes `string[]` (mutable), so the `readonly` annotation forces the `[...]` spread purely to satisfy its own over-restrictive type; and the call-site comment ("busts list + single-record + aggregate") overstates a constant that holds only `'videos'` (aggregate tags are handled by `BaseService`, as the constant's own JSDoc correctly states).

Pick one: (a) restore the inline `invalidateByTags(['videos'])` with the JSDoc co-located — matches pre-PR style and is simplest; or (b) keep the constant but type it `string[]` (drop `readonly`) so no spread is needed. Use (b) only if multi-tag expansion is genuinely imminent.

## Verdict

**APPROVE** — correctness is sound and the seven hardening fixes are well-tested; no blockers. Address W1–W3 before or shortly after merge (file-size decomposition, the resolution-multiplier duplicate, and the `organization` type contract); S1 is cosmetic with no behavioral impact. Per team policy these are at the author's discretion as in-branch fixups or a fast follow-up.

PR #877, opened 2026-06-29
fix(posts): failure-safe activity creation, weighted thread limits, hook prompt-builder alignment (#861)

### LINKED ISSUE #861: Posts generation: failure-safe activity creation, weighted thread limits, hook DTO/Replicate alignment (#691)
Heavier CodeRabbit findings from #857, deferred from the quick-win batch. File: `apps/server/api/src/collections/posts/services/post-generation.service.ts`.

1. **Activity creation outside failure handling (~524, also 686-699):** if `activitiesService.create()` throws, the method exits before the catch and leaves already-created processing posts stuck. Create the activity inside the try and guard `activitiesService.patch()` when no activity exists.
2. **Thread posts not validated with weighted limits (~743):** `parseTweetContent(content, additionalCount)` falls back to 560 unweighted chars, so thread replies over X's 280 weighted-char limit can be accepted. Pass the Twitter weighted-character constraint into the call.
3. **Hook generation DTO/Replicate alignment (~947):** verify `generateTextCompletionSync` signature + `GenerateHooksDto` org field usage flagged by CodeRabbit's analysis.

VincentShipsIt:
## Summary

Resolves the three deferred CodeRabbit findings from #857 in `apps/server/api/src/collections/posts/services/post-generation.service.ts`.

### 1. Failure-safe activity creation
`activitiesService.create()` ran **before** the `try` in both `generateAccountContentAsync` and `expandThreadAsync`. If it threw, the method exited before the catch, leaving the already-created `PROCESSING` posts/children stuck forever. Now the activity is created **inside** the `try`, and the catch's `activitiesService.patch()` is guarded with `if (activity)` — so every post/child is driven to `FAILED` even when activity creation itself fails.

### 2. Weighted thread-limit validation
Thread expansion called `parseTweetContent(content, additionalCount)` with no context, falling back to the **560 unweighted** char default — so thread replies exceeding X's **280 weighted-character** limit were silently accepted. It now passes a canonical `TWITTER_THREAD_CONSTRAINTS` (280 weighted). `parseTweetContent`'s context param is narrowed to the only field it reads (`constraints`).

### 3. Hook generation DTO/Replicate alignment
`generateHookVariations` passed a raw string to `replicateService.generateTextCompletionSync(version, input: Record<string, unknown>)` behind a `@ts-expect-error`, and the controller computed `publicMetadata` then **discarded it** (org never reached the service). It now routes through `promptBuilderService.buildPrompt` with the caller's organization and `SystemPromptKey.HOOK_GENERATOR`, forwarding a typed input object — matching the other generation flows (`enhanceDescription`, `buildAccountGenerationPrompt`). The `@ts-expect-error` is removed. A dedicated `hookGeneration` token limit (900) is added so the newly-introduced `max_tokens` cap does not truncate multi-hook output (the prior raw-string call had no cap).

## Verification
- `bunx biome check` — clean (+ pre-commit secretlint/biome hooks passed)
- `bunx vitest run src/collections/posts/services/post-generation.service.spec.ts` — **17/17 pass** (13 existing + 4 new covering all three paths: activity-creation-throws drives posts/children to FAILED without patching a missing activity; over-280-weighted thread replies rejected; hooks build input via the prompt builder with org + HOOK_GENERATOR)
- Adversarial multi-agent review of the diff (find → verify); the one confirmed actionable finding (a too-tight hook `max_tokens` cap) is fixed in this PR.
- Full workspace type-check deferred to CI: a standalone `tsc` can't resolve `@genfeedai/*` workspace packages without a full workspace build (excluded by the laptop-friendly resource policy); the Mac Studio heavy-validation host was unreachable. New symbols (`AccountPublishingConstraints`, `SystemPromptKey.HOOK_GENERATOR`, `TEXT_GENERATION_LIMITS.hookGeneration`) are confirmed exported/present.

Closes #861

### DIFF
--- a/apps/server/api/src/collections/posts/controllers/operations/posts-operations.controller.ts
+++ b/apps/server/api/src/collections/posts/controllers/operations/posts-operations.controller.ts
@@ -828,10 +828,13 @@ export class PostsOperationsController {
     @Body() dto: GenerateHooksDto,
     @Req() _request: Request,
   ) {
-    const _publicMetadata = getPublicMetadata(user);
+    const publicMetadata = getPublicMetadata(user);
 
     try {
-      return await this.postGenerationService.generateHookVariations(dto);
+      return await this.postGenerationService.generateHookVariations(
+        dto,
+        publicMetadata,
+      );
     } catch (error: unknown) {
       const errorMessage = (error as Error)?.message || 'Unknown error';
       this.logger.error(`Hook generation failed: ${errorMessage}`);
--- a/apps/server/api/src/collections/posts/services/post-generation.service.spec.ts
+++ b/apps/server/api/src/collections/posts/services/post-generation.service.spec.ts
@@ -9,10 +9,15 @@ import { PostGenerationService } from '@api/collections/posts/services/post-gene
 import { PostsService } from '@api/collections/posts/services/posts.service';
 import { TemplatesService } from '@api/collections/templates/services/templates.service';
 import { TrendReferenceCorpusService } from '@api/collections/trends/services/trend-reference-corpus.service';
+import { TEXT_GENERATION_LIMITS } from '@api/constants/text-generation-limits.constant';
 import { ReplicateService } from '@api/services/integrations/replicate/replicate.service';
 import { NotificationsPublisherService } from '@api/services/notifications/publisher/notifications-publisher.service';
 import { PromptBuilderService } from '@api/services/prompt-builder/prompt-builder.service';
-import { CredentialPlatform } from '@genfeedai/enums';
+import {
+  CredentialPlatform,
+  PostStatus,
+  SystemPromptKey,
+} from '@genfeedai/enums';
 import { LoggerService } from '@libs/logger/logger.service';
 import { Test, TestingModule } from '@nestjs/testing';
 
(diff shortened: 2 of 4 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

This PR introduces failure-safe activity creation, weighted thread limits, and hook prompt-builder alignment across the post generation pipeline. The behavioral intent is sound. The structural problem is that the implementation adds ~49 lines to an already over-limit file, duplicates two private helpers verbatim, duplicates a business-constraint literal that already has a canonical home, and introduces a repeated inline pattern for activity failure handling — all in code the PR directly touched. Four of these are blockers under the thermo standard's criterion 6 (preserved incidental complexity adjacent to changed code) and criterion 1 (file over 1000 lines).

### Blockers

**1. `apps/server/api/src/collections/posts/services/post-generation.service.ts:883–1036` — File crosses 1000-line threshold; two methods are fully extractable**

The file is 1036 lines post-PR (+49 net). `enhanceDescription` (lines 883–934) and `generateHookVariations` (lines 944–1036) are stateless, synchronous-result flows with zero coupling to the background pipeline state or any private helper in the class — they call only three injected services (`promptBuilderService`, `replicateService`, `templatesService`). Extract them into a new `PostEnhancementService` with those three dependencies. Add it to the same module's providers and inject it into the controller alongside the existing service. `PostGenerationService` drops to ~882 lines with a single clear domain (background generation pipelines). No behavior change.

**2. `apps/server/api/src/collections/posts/services/post-generation.service.ts:677–696 and 857–873` — `handleGeneratedPostFailure` and `handleExpandPostFailure` are character-for-character duplicates**

Both private methods carry identical signatures `(postId: string, error: unknown): Promise<void>` and identical bodies (patch to `FAILED`, websocket emit with `WebSocketPaths.post(postId)` and `Status.FAILED`, same catch/log message). The PR added new outer-catch call sites to both rather than collapsing them. Delete `handleExpandPostFailure` entirely, rename `handleGeneratedPostFailure` to `handlePostFailure`, and update all six call sites (lines 643, 648, 669, 823, 830, 849 in the post-PR file). Net reduction is ~20 lines with no behavior change.

**3. `apps/server/api/src/collections/posts/services/post-generation.service.ts:60–67` — `TWITTER_THREAD_CONSTRAINTS` duplicates the canonical value already in `AccountPublishingContextService`**

The five-field object literal at lines 60–67 is byte-for-byte identical to the object returned by `AccountPublishingContextService.getConstraints()` at `account-publishing-context.service.ts:202–209` for `platform=TWITTER, surface!='x-article'`. `AccountPublishingContextService` is already injected into `PostGenerationService`. Any future change to the Twitter character limit, weighted-characters flag, or rich-article toggle now requires edits in two files with no compiler enforcement of consistency. Export a named constant from `account-publishing-context.service.ts` — e.g. `export const TWITTER_STANDARD_POST_CONSTRAINTS: AccountPublishingConstraints = { ... }` — and reference it from both `getConstraints()` and from `post-generation.service.ts` (replacing the inline literal). The constant at line 60–67 must be deleted so there is exactly one definition.

**4. `apps/server/api/src/collections/posts/services/post-generation.service.ts:659–666 and 839–846` — Duplicate inline activity-failure blocks without a shared helper**

The two `if (activity)` guards introduced by this PR in the outer catch blocks of `generateAccountContentAsync` and `expandThreadAsync` are structurally identical: same guard check, same `activitiesService.patch` call, same `ActivityKey.POST_FAILED` key, same `JSON.stringify` value shape — differing only in the fallback error string. The PR already delegates per-post failure to private helpers but skips that discipline for activity-level failure, leaving the same conditional duplicated across two unrelated call sites. Add a private `async failActivity(activity: ActivityDocument | undefined, errorMessage: string): Promise<void>` method that owns the guard and the patch call. Replace the two inline blocks with `await this.failActivity(activity, (error as Error)?.message ?? 'Generation failed')` and `await this.failActivity(activity, (error as Error)?.message ?? 'Thread expansion failed')` respectively. Fully behavior-preserving.

### Simplifications

**`apps/server/api/src/collections/posts/services/post-generation.service.ts:533 and 715` — Opaque utility-type chain hides a concrete named type**

Both `activity` declarations introduced by this PR use `Awaited<ReturnType<ActivitiesService['create']>>`, which resolves provably to `ActivityDocument` (`BaseService<ActivityDocument, ...>.create()` returns `Promise<T>`). `ActivityDocument` is already a named exported interface from `@api/collections/activities/schemas/activity.schema.ts`. The utility-type derivation hides the domain type, makes the declaration unsearchable by type name, and couples the annotation to the method's return-type spelling. Import `ActivityDocument` and replace both declarations with `let activity: ActivityDocument | undefined`. Identical runtime behavior.

### Verdict

**BLOCK.** Four confirmed blockers must be resolved before merge:
1. Extract `enhanceDescription` + `generateHookVariations` into `PostEnhancementService` to bring the file under 1000 lines.
2. Collapse `handleGeneratedPostFailure` / `handleExpandPostFailure` into a single `handlePostFailure` method.
3. Remove `TWITTER_THREAD_CONSTRAINTS` in favor of an exported constant from `AccountPublishingContextService`.
4. Extract the duplicated `if (activity)` catch blocks into a `failActivity` helper.

The simplification on `Awaited<ReturnType<...>>` should be addressed in the same pass since it touches the same lines being restructured.

PR #892, opened 2026-06-29
fix(brands): guard default recurring workflow creation against concurrent duplicates

VincentShipsIt:
## Problem

`ensureDefaultBundle` in [default-recurring-content.service.ts](apps/server/api/src/collections/brands/services/default-recurring-content.service.ts) read existing workflows then created one per missing `contentType`. Two concurrent onboarding requests for the same brand could both observe "no existing default workflow" and both create one, producing **duplicate system-managed default recurring workflows per `contentType`**.

> **Update:** #864 merged to `master` (`dc19ec30d`) while this was in review. It added a `Serializable` transaction around the read+loop **but left the create on a non-tx client** (`createDefaultRecurringWorkflow` still calls `this.prisma.workflow.create`). The inserts therefore execute on a separate auto-commit connection outside the transaction's SSI tracking, so no read/write conflict is registered and **duplicates remain possible**. This PR is rebased on that change and replaces the incomplete guard with a complete one (its deterministic `orderBy` is preserved).

## Options evaluated

1. **Serializable transaction, create on the tx client (chosen).** Wrap the per-`contentType` check-and-create in `$transaction(..., { isolationLevel: 'Serializable' })`, threading the tx client through to `workflow.create` so the insert actually participates in SSI conflict detection. Mirrors the in-repo pattern `WorkflowsService.ensureDailyTrendsDigestWorkflow` (fast-path pre-check → in-tx re-check → create on tx → swallow `P2034`).
2. **Partial unique index + upsert (rejected).** `contentType` lives in `workflow.metadata` JSON (`metadata Json?`), so a unique column index isn't available — it would require promoting `contentType` to a real column plus a migration. Far more invasive for no benefit over option 1.

## Change

- New private `ensureRecurringWorkflowForType`: re-checks existence **inside** a Serializable transaction (the unindexed JSON-path filter forces a seq-scan → relation-level `SIReadLock` on `workflows`, so any concurrent insert forms the read/write conflict SSI needs), then creates on the **tx client**. A `P2034` serialization failure is treated as "already created by the winner".
- The `post` credential lookup is resolved on the **non-tx** client *before* the transaction. Keeping it inside would widen the tx read-set so a concurrent credential write (e.g. an OAuth connect during onboarding) could raise a `P2034` unrelated to the workflow insert — which the catch would misread as success and silently skip the workflow. `workflow.create` itself stays fully inside the tx.
- Preserves #864's deterministic `orderBy: { createdAt: 'desc' }` on the bulk read.

## Tests

New `default-recurring-content.service.spec.ts` with an in-memory Prisma fake that models snapshot isolation + SSI conflict detection (`P2034`):

- **concurrency regression** — two concurrent `ensureDefaultBundle` calls (forced into the race by a size-2 barrier on the bulk read) yield **exactly one workflow per `contentType`**. Verified by mutation that this fails on a non-transactional create path (asserts length 3, gets 6) and on an isolation downgrade to `ReadCommitted`.
- happy-path (3 workflows, each via a `Serializable` tx), `P2034`-swallow, non-`P2034`-rethrow, and disabled-workflow re-enable idempotency.

## Verification

- `default-recurring-content.service.spec.ts` — **5 passing**; loadable brands spec sweep — **31 passing**.
- Focused type-check (`tsconfig.typecheck.json`) — clean for changed files (only pre-existing global `multer`/`baseUrl` diagnostics remain).
- `biome check` + secretlint — clean (lint-staged on commit). Rebased on current `master`; conflict with #864 resolved.


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

Structural pass (cursor/plugins `thermo-nuclear-code-quality-review` standard). Method: parallel finders across canonical-helper reuse, simpler-reframing, correctness/atomicity, and test quality, then adversarial verification — 5 candidates raised, 2 confirmed (both test-only polish). The substantive challenges all dissolved on verification, which is the headline result for this PR.

### Core fix is sound (challenges refuted)

- **Canonical-helper reuse:** the open-coded `this.prisma.$transaction(fn, { isolationLevel: 'Serializable' })` + `tx as unknown as PrismaTransactionClient` cast was checked against the existing `TransactionUtil.runInTransaction` helper. It follows the dominant in-domain precedent — `workflows.service.ts` opens Serializable transactions the same raw way ~10×, and `TransactionUtil` performs the identical double-cast internally. Not a duplicated-helper defect.
- **Simpler reframing (partial unique index instead of Serializable):** verified **infeasible** with the current schema — the dedupe predicate keys off a JSON `metadata` path plus a brand many-to-many connect (no scalar `brandId` column on `workflow` to anchor a partial unique index). This validates the Serializable + P2034-swallow approach as the right tool.
- **Correctness:** `P2034` is the correct serialization-failure code; the credential lookup is correctly hoisted **outside** the transaction (with a sound comment on not joining the tx read-set); no nested-transaction or table-wide-lock regression exists (two finder BLOCKERs alleging these cited code/line numbers that aren't in the branch — dropped).

### SIMPLIFICATION

**1. `default-recurring-content.service.spec.ts:369` — the concurrency test's transaction-count assertion is too weak to guard what it claims.**

```ts
expect(transactionSpy.mock.calls.length).toBeGreaterThan(0);
```

Its own comment says it exists so that "dropping the isolation level fails the concurrency test directly." But both callers clear the size-2 barrier, then each runs the `for...of` loop over all three content types (P2034 is caught with `continue`, not `break`), guaranteeing 6 `$transaction` calls. `> 0` would still pass even if a regression fired a single transaction — it doesn't actually pin the invariant.

**Fix:** assert the real lower bound, `toBeGreaterThanOrEqual(6)` (3 content types × 2 callers), so the test fails if either caller is skipped or the loop short-circuits.

**2. `default-recurring-content.service.spec.ts:347-350` — the setup comment overstates the interleaving guarantee.**

The comment states the per-type transactions "interleave naturally under the JS microtask queue." The barrier is a one-shot gate inside the single outer `findMany` stub, so it deterministically synchronizes only the initial bulk read. After both callers pass it, their per-content-type loops race purely on JS scheduling — genuine same-type overlap is reliable for the first type (`post`) only, not all three.

**Fix:** soften the comment to say the barrier reproduces the race on the initial read (and at least the first content type), and that the count assertion catches duplicates regardless of which path — Serializable conflict or in-tx re-check — resolved them.

### Verdict: **APPROVE**

No blockers. The concurrency fix is correct, well-reasoned, and faithfully tested by a heavyweight-but-justified fake-Prisma harness (snapshot isolation + conflict detection that genuinely reproduces the 6-rows-vs-3-rows race). The two items above are test-precision refinements, not merge gates.

VincentShipsIt:

PR #900, opened 2026-06-29
feat(security): encrypt OAuth credential secrets at rest

VincentShipsIt:
## Why

OAuth credential secrets were persisted in **plaintext** via `credentialsService`. CodeRabbit flagged this on #876 (Google Search Console), but it's a **cross-integration pattern** — Google Ads, Instagram, Telegram, Twitter analytics and others all stored/used tokens unencrypted. This is the security-hardening follow-up.

## What

A single **encrypt-on-write boundary** in `CredentialsService` so every provider's secrets are encrypted at rest, regardless of whether the provider remembered to encrypt.

**Encrypted columns (5):** `accessToken`, `accessTokenSecret`, `oauthToken`, `oauthTokenSecret`, `refreshToken`.
**Deliberately excluded:** `oauthState` — it's an OAuth-callback lookup key matched verbatim in `findOne({ oauthState })`; encrypting it would break Discord/Slack/etc. state validation. (`oauthTokenHash` is a hash, `*Expiry` are timestamps.)

### Boundary
- New injectable **`CredentialCryptoService`**, keyed from `ConfigService.tokenEncryptionKey` (**never `process.env`**). AES-256-GCM, `iv:ciphertext:authTag` envelope — **byte-compatible with the legacy `EncryptionUtil`**, so existing encrypted data and existing decrypt sites interop with **zero migration**. Encrypt is **idempotent** (skips already-encrypted values → no double-encryption).
- `CredentialsService.create/patch/patchAll` encrypt secrets before persistence — and before `BaseService` logs the payload.
- Added a `ConfigService.tokenEncryptionKey` getter (throws if unset, so encryption can't silently fall back to plaintext).

### Reconciliation (decrypt-on-read stays explicit, matching the existing codebase pattern)
- **Stripped** fanvue's manual pre-encryption (the only Credential-model pre-encrypt) to avoid double-encryption.
- **Added `EncryptionUtil.decrypt`** at six consumers that read tokens raw and would otherwise have received ciphertext post-boundary: Instagram uploads + page listing, bots livestream delivery, ads-gateway `resolveAccessToken`, ads-research, telegram distribution, and the twitter analytics processor.
- These were found via an exhaustive read-site audit; a graceful no-op decrypt means legacy plaintext still works everywhere during rollout.

### Hardening
- Removed a refresh-token **preview log leak** (plaintext + ciphertext substrings) in `youtube-auth`.
- `decrypt()` now **warns** (without the secret) when an encrypted value fails authentication, instead of silently shipping ciphertext upstream.

### Migration
- `bun run encrypt:credentials:dry` / `encrypt:credentials` — one-time backfill that encrypts existing plaintext rows. Idempotent, cursor-paginated, dry-run by default. Safe to run after deploy (graceful decrypt tolerates the transition).

## Tests
- `CredentialCryptoService`: round-trip, fresh-IV, idempotency, plaintext passthrough, `oauthState`/non-secret passthrough, auth-failure warn-and-return, `EncryptionUtil` interop.
- `CredentialsService`: encrypt-on-write on create/patch/patchAll/saveCredentials, idempotency, **no plaintext in written data or logs**.
- `ads-gateway`: decrypt-on-read returns plaintext to the adapter.
- Updated fanvue specs for the stripped pre-encryption.

## Verification
- `bunx turbo run type-check --filter=@genfeedai/api` ✅ (and `@genfeedai/workers` ✅)
- 65 focused specs green (credentials, crypto, fanvue, ads-gateway, instagram).
- Reviewed via a multi-agent adversarial pass; 3 blockers found and fixed (a missed Instagram read site, a double-decrypt, silent decrypt failures).

## Notes / follow-ups
- **Local dev:** `TOKEN_ENCRYPTION_KEY` must be **identical across all `apps/server/*` services** (api + workers), or the worker can't decrypt api-written tokens. Production injects it consistently.
- Pre-existing static `EncryptionUtil`/`CryptoService` (used on other models — OrgIntegration, BYOK) are out of scope and untouched.
- Spun off separately: removing the unused `CredentialFullSerializer` (a latent footgun that exposes token fields; not wired to any controller).

### DIFF
--- a/apps/server/api/package.json
+++ b/apps/server/api/package.json
@@ -64,6 +64,8 @@
     "dev:direct": "cd .. && NODE_OPTIONS=\"--max-old-space-size=4096 --expose-gc\" nest start api --watch",
     "dev:portless": "portless run --name api.genfeed bun run dev:direct",
     "dev:webpack": "cd .. && NODE_OPTIONS=\"--max-old-space-size=4096 --expose-gc\" nest start api --watch",
+    "encrypt:credentials": "bun run scripts/backfill-credential-encryption.ts --live",
+    "encrypt:credentials:dry": "bun run scripts/backfill-credential-encryption.ts",
     "lint": "bun run scripts/check-request-context-usage.ts && biome lint --write src --no-errors-on-unmatched",
     "migrate:workflows": "node --max-old-space-size=8192 --expose-gc -r tsconfig-paths/register ../dist/apps/api-workflow-backfill/main",
     "seed:workflows": "bun run scripts/seeds/workflows.seed.ts",
--- a/apps/server/api/scripts/backfill-credential-encryption.ts
+++ b/apps/server/api/scripts/backfill-credential-encryption.ts
@@ -1,0 +1,169 @@
+/**
+ * Backfill Script: Encrypt Credential secrets at rest
+ *
+ * One-time data migration that encrypts any plaintext secret columns on the
+ * `credentials` table (access/refresh tokens and token secrets). Values that
+ * are already in the ciphertext envelope are skipped, so the script is
+ * idempotent and safe to re-run.
+ *
+ * The cipher (aes-256-gcm), key derivation (sha256(TOKEN_ENCRYPTION_KEY)) and
+ * envelope format (iv:ciphertext:authTag hex) match both `EncryptionUtil` and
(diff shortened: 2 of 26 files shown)


VincentShipsIt:

PR #914, opened 2026-06-29
ci: gate self-hosted release publish on full-suite + build-verify-selfhosted

VincentShipsIt:
## Summary

Closes the **"all tests on release"** gap. Today the community image users `docker run` is published on `release: published` with **zero** test gate — only `validate-master-ref`. The full unit matrix + e2e only ran at production **deploy** (`deploy-ecs` → `qa-before-deploy` → `full-suite`), never at the release cut. So a release could ship a broken self-hosted image.

## Change

`docker-publish.yml` `build-and-push` now `needs: [validate-master-ref, verify-suite, verify-selfhosted]`:

- **`verify-suite`** → `full-suite.yml` (ci unit matrix: sharded app + api + packages + server-services + web/desktop/mobile + extensions, **plus** `build-verify` prod-image boot **plus** `e2e`). Permissions mirror `deploy-ecs.yml`'s `qa-before-deploy` grant (a reusable workflow is capped by the caller's job-level permissions).
- **`verify-selfhosted`** → `build-verify-selfhosted.yml`, the all-in-one community-image build/boot guard that was previously wired into **no** automated path. full-suite's `build-verify` only covers the microservice/prod image; this covers `docker/Dockerfile.selfhosted`.

A verify failure leaves the GitHub release present but **blocks the image push** — the artifact users run is what's protected.

## Tradeoff

Adds ~20–30 min to a release publish. Intentional: a release now clears the same bar a production deploy does (defense-in-depth: community image vs. prod deploy).

## Validation

- `actionlint` clean on `docker-publish.yml`, `full-suite.yml`, `build-verify-selfhosted.yml`.
- To verify behavior: dispatch *Docker Publish (Self-Hosted)* (`workflow_dispatch`, `tag: staging`) from master → confirm both verify jobs run and `build-and-push` is blocked until both are green.

Part of a CI consolidation initiative prompted by #895 (more PRs follow: per-PR API tests + guard consolidation, dead-workflow cleanup, version-drift fixes).

PR #915, opened 2026-06-29
ci: focused API changed-tests + consolidate static guards

VincentShipsIt:
## Summary

Two PR-path improvements to `ci.yml`, part of the CI consolidation initiative (follows #914).

### 1. Close the per-PR API-test gap — `test-api-changed`

The full sharded `test-api` matrix gates on `workflow_call`, so it only runs at `full-suite`/deploy — **API unit regressions went un-caught on PR and master-push until deploy time.** There was a `test-app-changed` but no API counterpart, even though `test-scope` already computes an `api` output that nothing consumed.

`test-api-changed` mirrors `test-app-changed`: builds API deps (`turbo --filter=./packages/**`), then `vitest --changed` against `apps/server/api`, gated on `test-scope.outputs.api == 'true'` and `pull_request || push`. Fast (changed-only).

### 2. Consolidate four always-on guards → one `guards` job

`Architecture Guards`, `SQL Risk Audit`, `UI Guards`, `Design System` each did ~10–20s of real work but paid a full checkout + `bun install` (~2 min) on a **separate runner**. Merged into one job sharing a single setup → **~3 fewer runners (~6–8 runner-min/run)**, no critical-path impact (they sit far below the test jobs). `build.needs` updated `design` → `guards`.

- None of the four are required status checks, so collapsing their names is safe.
- **React Doctor intentionally stays separate** — it skips on draft PRs and runs its own diff-detection, so its trigger semantics differ from these always-on guards. Folding it in would change *when* it runs. (Plan listed 5; this is the zero-risk 4 + keep React Doctor.)

## Follow-up (after merge)

Add `Test App (changed)` + `Test API (changed)` to the *Passing CI on master* ruleset so tests actually block merges — today **no test is a required check** (only format/lint/typecheck/gitleaks/secretlint/socket gate a merge). Deferred to post-merge so the required checks reference jobs that already exist on master.

## Validation

- `actionlint` clean; no dangling job references; all four guard scripts confirmed present in root `package.json`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,10 +203,19 @@ jobs:
             bunx react-doctor@0.4.2 . --blocking error --project '*' --diff false
           fi
 
-  architecture-guards:
-    name: Architecture Guards
+  # Consolidated static guards — these four checks each do ~10-20s of actual
+  # work but previously ran as four separate jobs, each paying a full
+  # checkout + bun install (~2 min) on its own runner. Merged into one job they
+  # share a single setup, cutting ~3 runners (~6-8 runner-min) per run with no
+  # critical-path impact (they sit well under the test jobs). None of these are
+  # required status checks, so collapsing their individual names is safe.
+  # React Doctor is intentionally NOT folded in: it skips on draft PRs and runs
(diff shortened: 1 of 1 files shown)


VincentShipsIt:

PR #916, opened 2026-06-29
refactor(serializers): rename 14 admin Darkroom fleet serializers to Fleet*

VincentShipsIt:
## Summary

Renames the 14 admin **fleet** serializers from their legacy `Darkroom*` names to `Fleet*`, and collapses the alias indirection in `server/admin/index.ts`. Previously each was exported twice — once as `Darkroom*Serializer` (via `export *`) and once aliased as `Fleet*Serializer` — but only the `Fleet*` alias was ever consumed (by `fleet.controller.ts`). The `Darkroom*` names were dead rename artifacts from the cloud→genfeed migration.

Each of the 14 is a triplet (`server/admin/*.serializer.ts` + `configs/admin/*.config.ts` + `attributes/admin/*.attributes.ts`), renamed via `git mv`. Barrels collapsed to a single `export *` per file (no more aliases).

## ⚠️ Wire-format change (intentional)

`buildSerializer` derives the export symbol from the config's JSON:API `type` string, so renaming the symbol to `Fleet*` **also changes the `type` field** in admin fleet API responses, e.g. `darkroom-voice` → `fleet-voice`. This is safe:
- The frontend `BaseService` deserializes **generically and never validates the `type` string**; `fleet.service.ts` passes no typed serializer; there is no `api-types` coupling.
- The only consumers of these serializers are `apps/server/api/.../fleet.controller.ts` (server) — which already imports the `Fleet*` names — and the admin fleet UI (type-agnostic deserialization).
- The required E2E suite exercises the admin fleet pages and should confirm no regression.

## 🐞 Latent bug fixed

`DarkroomCloudFrontInvalidationSerializer` was **`undefined` at runtime**: its config `type` was `'darkroom-cloudfront-invalidation'` (PascalCases to `…CloudfrontInvalidation`), but the file destructured `…CloudFrontInvalidation` (capital F). So `FleetCloudFrontInvalidationSerializer` — used by `fleet.controller` for the CloudFront-invalidation endpoint — resolved to `undefined`. The new `type: 'fleet-cloud-front-invalidation'` PascalCases correctly, so the serializer is now a real object. The new regression test would fail if this regressed again.

## New regression test

Added `describe('Fleet admin serializers')` to `server-serializers.test.ts` — asserts all 14 `Fleet*` serializers are truthy, `typeof === 'object'`, and expose a `serialize` method. This is the guard against the `type`→name derivation silently producing `undefined` (which type-check cannot catch).

## Verification (re-run independently after the agent)
- `bunx turbo run type-check --filter=@genfeedai/serializers` — ✅
- `bunx turbo run type-check --filter=@genfeedai/api` — ✅ 17/17 (confirms `fleet.controller` still resolves all 14 `Fleet*` imports)
- `bun run test --filter=@genfeedai/serializers` — ✅ 781 passed (incl. the new Fleet block)
- `bunx biome check` — ✅ clean on touched files
- Zero `darkroom` references remain in the admin serializer dirs; the unrelated, actively-used `DarkroomCapabilitiesSerializer` (org settings) is untouched.

Net: 74 files, +304 / −266.


VincentShipsIt:

PR #895, opened 2026-06-29
[codex] Upgrade CI actions to Node 24 runtime

VincentShipsIt:
## Summary

Updates GitHub Actions dependencies that still declared deprecated Node 20 runtimes so the CI gate no longer emits Node 20 deprecation warnings while `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24` is enabled.

## Changes

- Bumped Gitleaks from `gitleaks/gitleaks-action@v2` to `@v3`.
- Bumped artifact upload/download actions to existing Node 24 runtime releases.
- Bumped Docker, AWS credentials, CodeQL SARIF upload, Expo, and GitHub release actions to Node 24-compatible releases.
- Preserved existing SHA-pinning style where the workflow already used exact action SHAs.

## Notes

`PlasmoHQ/bpp` still declares `node20` upstream on its latest tag and `main`, so that workflow action remains the only known Node 20 metadata source. Replacing it would require a behavioral rewrite of the browser extension publishing step or vendoring the action bundle.

## Validation

- `git diff --check`
- `actionlint` on changed workflow files
- metadata scan of external action `runs.using` declarations
- diff-scoped secret filename and token-pattern checks

PR #917, opened 2026-06-29
ci: delete orphaned detect-changes workflow + group github-actions dependabot

VincentShipsIt:
## Summary

Two safe, sustainable cleanups from the CI consolidation initiative (follows #914, #915).

### 1. Delete the dead `detect-changes.yml` (198 lines)

It declared `workflow_call` with a rich service-change output matrix, but **no workflow references it** (`grep` across `.github/` finds zero callers). `build-server-image.yml` builds the unified image directly without it. Orphaned since the cloud → genfeed.ai migration; recoverable from git history per the green-field policy.

### 2. Group github-actions version updates in Dependabot

Action-pin drift (`upload-artifact` @v4 vs @v7, `download-artifact` @v4 vs @v8, mixed SHA/float pins) is what forced the manual **#895** Node-24 sweep — and since Dependabot was `open-pull-requests-limit: 0` (security-only) for every ecosystem, that drift could never self-correct.

This enables **grouped** version updates for github-actions only: every action bump lands in **one weekly PR**, which resolves the team's documented "stream of noisy PRs" objection while keeping drift in check. **npm + docker stay security-only** (unchanged).

## Deliberately deferred (not blind-swept here)

The one-time hand bumps — `react-doctor` 0.4.2→0.5.6, `upload/download-artifact` alignment, node 22→24 in `setup-bun-env` — each carry breaking-change risk (different CLI flags / artifact semantics), exactly the class #895 was repairing. They belong in individually-validated PRs (or will arrive via the first grouped Dependabot PR, CI-tested as a normal change) rather than a blind sweep in a cleanup PR.

## Validation

- `dependabot.yml` parses; grouping syntax is valid Dependabot v2.
- Confirmed zero remaining references to `detect-changes` in `.github/`.


VincentShipsIt:

PR #931, opened 2026-06-30
fix(infra): enable Redis TLS auth

### LINKED ISSUE #635: Infra: enable Redis transit encryption and auth token
## Source
- PR review thread: https://github.com/genfeedai/genfeed.ai/pull/609#discussion_r3418957472
- Found while triaging Spark/GitHub review notification emails on 2026-06-18.

## Current evidence
`infra/terraform/genfeed-prod/elasticache.tf` still provisions Redis without transit encryption and without an auth token. That is a real production hardening gap, but enabling it requires connection-string, secret, rollout, and compatibility work rather than a drive-by Terraform edit.

## Done when
- ElastiCache transit encryption is enabled for production.
- Redis auth token handling is added through SSM/secrets, not committed values.
- App/service Redis clients use the TLS/auth URL format successfully.
- Migration/rollout steps are documented and verified.

VincentShipsIt:
Closes #635

## Summary
- enable production ElastiCache in-transit encryption and AUTH with a Terraform-generated token stored as `/genfeed/production/REDIS_PASSWORD` SecureString
- switch ECS service Redis env to `rediss://...` + `REDIS_TLS=true` while injecting `REDIS_PASSWORD` as an ECS secret
- update shared node-redis connection helpers and callers so separately injected Redis passwords authenticate without embedding tokens in `REDIS_URL`
- document the production Redis TLS/auth rollout contract in `infra/terraform/README.md`

## Validation
- `tofu -chdir=infra/terraform/genfeed-prod init -backend=false -input=false`
- `tofu -chdir=infra/terraform/genfeed-prod fmt -check -diff`
- `tofu -chdir=infra/terraform/genfeed-prod validate`
- `bun run test:workspace -- packages/libs/redis/redis-connection.utils.spec.ts packages/libs/adapters/redis-io.adapter.spec.ts packages/libs/redis/redis.service.spec.ts`
- `bunx biome check ...touched Redis/API/files/notifications files...`
- `bunx turbo run type-check --filter=@genfeedai/libs --filter=@genfeedai/api --filter=@genfeedai/files --filter=@genfeedai/notifications`
- `bunx turbo run lint --filter=@genfeedai/libs --filter=@genfeedai/api --filter=@genfeedai/files --filter=@genfeedai/notifications`
- `bunx prettier --check infra/terraform/README.md`
- static Redis TLS/auth Terraform contract assertion
- `git diff --check origin/master..HEAD`
- touched-file Secretlint and pre-commit lint-staged Secretlint/Biome

## Notes
- I did not run `tofu plan` or `tofu apply`; this automation must not perform live production infrastructure operations or depend on production AWS state.
- The generated Redis token is not committed. Terraform stores it in SSM as a SecureString and injects it into ECS task definitions as `REDIS_PASSWORD`. The S3 Terraform state remains sensitive infrastructure state.

### DIFF
--- a/apps/server/api/src/services/cache/services/cache-client.service.ts
+++ b/apps/server/api/src/services/cache/services/cache-client.service.ts
@@ -1,6 +1,9 @@
 import { ConfigService } from '@api/config/config.service';
 import { LoggerService } from '@libs/logger/logger.service';
-import { parseRedisConnection } from '@libs/redis/redis-connection.utils';
+import {
+  buildNodeRedisClientOptions,
+  parseRedisConnection,
+} from '@libs/redis/redis-connection.utils';
 import {
   Injectable,
   type OnModuleDestroy,
@@ -24,8 +27,8 @@ export class CacheClientService implements OnModuleInit, OnModuleDestroy {
   ) {
     const config = parseRedisConnection(this.configService);
 
-    this.client = createClient({
-      socket: {
+    this.client = createClient(
+      buildNodeRedisClientOptions(config, {
         connectTimeout: 3_000,
         reconnectStrategy: (retries: number) => {
           if (retries >= CacheClientService.MAX_RECONNECT_RETRIES) {
@@ -43,10 +46,8 @@ export class CacheClientService implements OnModuleInit, OnModuleDestroy {
           );
           return delay;
         },
-        ...(config.tls && { tls: true }),
-      },
-      url: config.url,
-    });
+      }),
+    );
 
     this.client.on('error', (error: Error) => {
       this.logger.error(`${this.constructorName} Redis client error`, error);
--- a/apps/server/api/src/services/microservices/microservices.service.ts
+++ b/apps/server/api/src/services/microservices/microservices.service.ts
@@ -1,7 +1,10 @@
 import process from 'node:process';
 import { ConfigService } from '@api/config/config.service';
 import { LoggerService } from '@libs/logger/logger.service';
-import { parseRedisConnection } from '@libs/redis/redis-connection.utils';
+import {
+  buildNodeRedisClientOptions,
+  parseRedisConnection,
+} from '@libs/redis/redis-connection.utils';
 import { CallerUtil } from '@libs/utils/caller/caller.util';
 import { HttpService } from '@nestjs/axios';
 import {
@@ -105,14 +108,12 @@ export class MicroservicesService implements OnModuleInit {
     );
 
     try {
-      this.redisClient = createClient({
-        socket: {
+      this.redisClient = createClient(
+        buildNodeRedisClientOptions(config, {
           connectTimeout: MicroservicesService.REDIS_CONNECT_TIMEOUT_MS,
           reconnectStrategy: () => false as const,
-          ...(config.tls ? { tls: true as const } : {}),
-        },
-        url: config.url,
-      });
+        }),
+      );
 
       this.redisClient.on('error', (err: Error) => {
         this.loggerService.error('Redis Client Error', err);
--- a/apps/server/api/src/services/notifications/notifications.service.ts
+++ b/apps/server/api/src/services/notifications/notifications.service.ts
@@ -8,7 +8,10 @@ import type {
   ITelegramMessageOptions,
 } from '@genfeedai/interfaces';
 import { LoggerService } from '@libs/logger/logger.service';
-import { parseRedisConnection } from '@libs/redis/redis-connection.utils';
+import {
+  buildNodeRedisClientOptions,
+  parseRedisConnection,
+} from '@libs/redis/redis-connection.utils';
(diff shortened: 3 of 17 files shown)


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

This PR introduces a centralized `buildNodeRedisClientOptions()` helper in `packages/libs/redis/redis-connection.utils.ts` and threads it through seven call sites (cache-client, microservices, notifications, files/websocket, redis.service, redis-io.adapter). On the Terraform side it enables ElastiCache transit encryption, generates an AUTH token via `random_password`, stores it as an SSM SecureString, and injects it into all task definitions. The structural direction is correct: consolidating TLS + password options into one factory eliminates per-site duplication and makes the cluster hardening auditable in a single diff. There are no blockers — the password flows correctly through `parseRedisConnection` → `buildNodeRedisClientOptions` → `createClient` and the Terraform changes are self-consistent. Four simplification paths were identified, all in code this PR itself introduced.

Three candidate blockers were raised during review and **dismissed under adversarial verification**, recorded here so they are not re-litigated:
- *"`RedisReconnectStrategy` signature incompatible with node-redis"* — false. A `(retries: number) => number | false` function is structurally assignable to `(retries, cause) => ... ` (fewer-parameter functions are assignable to wider function types in TS).
- *"`transit_encryption_enabled` is ForceNew → destroys the live cluster"* — false. The provider's ForceNew on that attribute is conditional (`customdiff.ForceNewIf`) and only fires below engine 7.0.5; this cluster is `7.1`, so it is an in-place modification.
- *"8,911-line file / unsafe as-casts / silent connectTimeout"* — all pre-existing on master, not introduced or worsened here.

---

## Simplifications

**`packages/libs/adapters/redis-io.adapter.ts:30–43` — 12-line if-chain reconstitutes fields already stored on the adapter**

`connectToRedis` builds an inline `{ get: (key) => ... }` duck-typed ConfigService with three explicit string-key branches solely to satisfy the `parseRedisConnection` signature, whose output is immediately consumed by `buildNodeRedisClientOptions`. The three values being proxied (`REDIS_URL`, `REDIS_TLS`, `REDIS_PASSWORD`) are already stored as class fields (`this.redisUrl`, `this.redisTls`, `this.redisPassword`). The if-chain's only job is to undo the storage that happened in the constructor. Collapse it to an object-literal lookup:

```ts
const config = parseRedisConnection({
  get: (key) => ({
    REDIS_URL: this.redisUrl,
    REDIS_TLS: this.redisTls,
    REDIS_PASSWORD: this.redisPassword,
  } as Record<string, string | boolean | undefined>)[key],
});
```

Five lines instead of twelve, behaviorally identical.

---

**`packages/libs/adapters/redis-io.adapter.ts` — positional constructor fields create a ConfigService shim that will drift**

The deeper issue: `RedisIoAdapter` stores the three config values as separate positional constructor arguments because the caller (`notifications/src/main.ts`) already has a `ConfigService` and extracts each value individually before passing them in. This forces `connectToRedis` to reconstruct a fake ConfigService to call `parseRedisConnection`, which was designed to read from a real one. If a fourth env var is added to `parseRedisConnection` (e.g. `REDIS_CONNECT_TIMEOUT`), the shim silently omits it.

The fix is to move config parsing one level up: call `parseRedisConnection(configService)` in `notifications/src/main.ts` and pass the resulting `ParsedRedisConfig` into `RedisIoAdapter` as a single constructor argument. `connectToRedis` then calls `buildNodeRedisClientOptions(this.config)` directly with no shim. This eliminates the multi-argument positional constructor and removes the parallel config-reading path from the adapter entirely.

---

**`packages/libs/redis/redis-connection.utils.ts:~82–110` — `connectTimeout` default hardcoded in two places in the same file**

`buildNodeRedisSocketOptions` already defaults `connectTimeout` to `3_000` via its parameter default. `buildNodeRedisClientOptions` (added here) redundantly applies `options.connectTimeout ?? 3_000` before delegating to it. If the default changes it must be updated in two places. Pass `options.connectTimeout` (possibly `undefined`) straight through and let the parameter default own the value from one location.

---

**`packages/libs/redis/redis-connection.utils.ts` — `buildNodeRedisClientOptions` lacks an explicit return type**

This is a public exported function consumed at seven call sites; its inferred return type is a conditional spread. Without an explicit annotation, internal restructuring (renaming `socket`, changing the TLS branch shape) surfaces at call sites only incidentally via the `redis` package's parameter types rather than at the definition. Given the repo's strict-TS rules, export a named interface and annotate the function explicitly (`packages/interfaces/` or top of this file). Matches the existing `buildBullMQConnection` gap — worth fixing both.

---

**Verdict: APPROVE.** No blockers. All four items are new code introduced by this PR and are mechanical cleanups — land as a follow-up or squash in before merge at the author's discretion.

VincentShipsIt:

PR #937, opened 2026-06-30
fix(deploy): skip 10m image wait for non-server commits

VincentShipsIt:
## Problem

The `Resolve available server image` step in `deploy-ecs.yml` **always** waited up to ~10 minutes for `server:<HEAD>` (20 × 30s polls) *before* falling back to ancestor reuse:

```bash
if wait_for_image "$REQUESTED_TAG"; then ...   # up to 10m
fi
# ... only now check ancestors
```

For a **frontend- or workflow-only commit**, `build-server-image.yml` is paths-gated and never builds a `server:<HEAD>`. So that image will *never* appear — yet every such deploy burned the full ~10m wait before reusing a perfectly valid ancestor image.

## Fix

Reorder resolution so the wait happens only when an image is actually expected:

1. **Explicit tag** (rollback/pinned) → wait, then fail closed. *(unchanged)*
2. **`server:<HEAD>` already present** → use immediately (fast path, no loop).
3. **Otherwise** find the newest first-parent ancestor with an image:
   - **server-image inputs unchanged** since it → reuse it **immediately, no wait** (the HEAD image will never be built for a non-server commit).
   - **inputs changed** → a `server:<HEAD>` build is expected/in-flight → wait up to ~10m, then fail closed.

### Impact

- Frontend / workflow-only deploys: **~10m of dead polling → a single `git diff`**.
- Server-affecting deploys: unchanged (still wait for the HEAD build).
- Fail-closed safety on explicit tags and missing server builds preserved.

## Verification

- `python3 yaml.safe_load` parse OK.
- `actionlint` clean (incl. shellcheck on the run block).

### DIFF
--- a/.github/workflows/deploy-ecs.yml
+++ b/.github/workflows/deploy-ecs.yml
@@ -157,14 +157,23 @@ jobs:
             return 1
           }
 
-          if wait_for_image "$REQUESTED_TAG"; then
-            echo "sha=$REQUESTED_TAG" >> "$GITHUB_OUTPUT"
-            exit 0
-          fi
-
+          # Explicit tag (rollback / pinned deploy): it must already exist.
+          # Wait in case a build is still in flight, then fail closed.
           if [ "$EXPLICIT_TAG" = "true" ]; then
+            if wait_for_image "$REQUESTED_TAG"; then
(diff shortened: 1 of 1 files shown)


VincentShipsIt:

PR #1076, opened 2026-07-02
refactor(api): split WorkflowsService into focused workflow services (#754)

### LINKED ISSUE #754: Split workflows service
`apps/server/api/src/collections/workflows/services/workflows.service.ts` has grown to 1,573 lines with a 9-argument constructor (`prisma`, `logger`, plus 7 `@Optional()` services, lines 64-79) and ~25 methods spanning five unrelated responsibilities: workflow CRUD/templating, a hand-rolled legacy step-execution engine, node-engine delegation, lifecycle transitions, and a full webhook subsystem including crypto helpers.

> This is distinct from the executor/adapter decomposition tracked in #521. Those services (`workflow-executor.service.ts`, `workflow-engine-adapter.service.ts`) are *injected* here via `@Optional()`; this class holds its own separate logic and a duplicate legacy execution path.

## Confirmed structural findings (file:line)

- **God object: 9 constructor dependencies, ~25 methods across 5 concerns** (constructor 64-79): (a) CRUD/templating, (b) a legacy step execution engine, (c) the new node-engine delegation path, (d) lifecycle (`publishWorkflowLifecycle`, `archiveWorkflow`, `lockNodes`, `unlockNodes`), and (e) an entire webhook subsystem.
- **`executeWorkflow` (256-434, ~178 lines):** a single method doing status patching, event emission, a hand-rolled topological scheduler with ready-step detection (292-328), per-step `Promise.all` execution with inline try/catch progress accounting (330-377), failure-cascade marking, and two terminal-status/error branches. Largest and most tangled method in the file.
- **`executePartialAsync` (955-1077, ~122 lines):** nests a 44-line `onNodeStatusChange` callback (980-1023) with stacked ternaries for `completedAt`/`progress`/`startedAt` derivation (988-1009).
- **Two parallel execution engines.** `executeWorkflowCompat` (437-480) branches on `shouldUseNodeExecutor` (1480-1488): if the workflow has `nodes` it delegates to `workflowExecutorService`, otherwise it falls back to the legacy `executeWorkflow` + `executeWorkflowStep` switch (482-526) and eight near-identical queue wrappers (`executeTransform`/`executeUpscale`/`executeCaptioning`/`executeResize`/`executeClip`, ~10-line wrappers around `executeQueuedJob`, 569-669) — a second full implementation of step execution living in the service.
- **Webhook subsystem (1336-1572):** `generateWebhook`, `regenerateWebhookSecret`, `deleteWebhook`, `findByWebhookId`, `triggerViaWebhook`, `patchWorkflowConfig`, `getWorkflowConfigRecord`, plus crypto helpers `generateWebhookId`/`generateWebhookSecret` — a cohesive concern that does not belong in the core workflow service.
- **`findByWebhookId` (1397-1416)** loads every non-deleted workflow (`findMany({ where: { isDeleted: false } })`) and scans in JS for a matching `config.webhookId` — an O(n) full-collection scan.

Five unrelated concerns share one class and one 9-dependency constructor, so a change to webhook crypto sits beside the topological scheduler and the legacy queue wrappers. Two execution engines for the same concept invite divergent bug fixes, and the webhook lookup degrades linearly with workflow count.

1. Extract a `WorkflowWebhookService` (the 7 webhook methods + `patchWorkflowConfig`/`getWorkflowConfigRecord` + crypto helpers); `WorkflowsService` delegates. Removes ~240 lines and one cohesive concern with no behavior change.
2. Extract the legacy step engine (`executeWorkflow`, `executeWorkflowStep`, `executeQueuedJob`, the eight `execute*` helpers) into a `LegacyWorkflowStepRunner`. Confirm via call-graph whether any production path still hits the non-node branch of `shouldUseNodeExecutor`; if not, delete it per the repo's green-field "prefer deletion over deprecation" rule.
3. Break `executeWorkflow`'s inline scheduler into named private methods (`selectReadySteps`, `markBlockedByFailure`, `runStep`) to get the method under ~80 lines.
4. Index `webhookId` (promote to a top-level column, or add a Prisma index on the JSON path) so `findByWebhookId` is a direct query instead of a table scan.

## Acceptance criteria

- `WorkflowsService` has ≤6 injected dependencies and the file is under ~600 lines.
- No method exceeds ~80 lines; `executeWorkflow` and `executePartialAsync` are decomposed.
- Webhook logic lives in its own service with its own spec; existing webhook + execution tests pass unchanged.
- The legacy step engine is either extracted behind a clear interface or deleted if unreachable, with a PR note confirming which.
- `findByWebhookId` no longer performs a full-collection in-memory scan.

VincentShipsIt:
## Summary

`workflows.service.ts` had grown to **2,653 lines / 9 constructor deps / ~50 methods** spanning five unrelated concerns (the issue was filed at 1,573 lines; the seeding subsystem added since then made it worse). This splits it behavior-preserving into four focused services, per the restructuring plan in #754:

| Service | Concern | Notes |
|---|---|---|
| `WorkflowTemplateSeederService` | 12 idempotent per-org `ensure*` seeders + content-schedule mirroring | The 8 near-identical template-loop seeders (~700 duplicated lines) are deduplicated into one generic race-safe `ensureSeededWorkflow` helper; Daily-Trends (schedule repair) and content-schedule (upsert) keep their bespoke semantics |
| `LegacyWorkflowStepRunner` | Step-based (pre-node) execution engine | `executeWorkflow`'s ~180-line inline scheduler decomposed into `selectReadySteps` / `markStepsBlockedByFailure` / `runStep` / `finalizeRun`. Extends `BaseService` over the same collection so `patch()` keeps identical cache-invalidation semantics without a circular dep |
| `WorkflowWebhookService` | Webhook credential generate/rotate/delete, lookup, public trigger path | Own spec added (10 tests) |
| `WorkflowRunControlService` | Partial runs, resume-from-failed, credit validation, execution logs | `executePartialAsync`'s 44-line inline callback with stacked ternaries decomposed into `recordNodeStatusChange` / `deriveNodeProgress` / `recordRunResult` |

`WorkflowsService` keeps CRUD/templating, lifecycle transitions, node locks, ownership guards, and the node-vs-legacy compat dispatch.

## Acceptance criteria (from #754)

- ✅ `WorkflowsService` ≤6 injected deps → **4** (`prisma`, `logger`, optional legacy runner + node executor); file **529 lines** (<600)
- ✅ No method exceeds ~80 lines; `executeWorkflow` and `executePartialAsync` decomposed
- ✅ Webhook logic lives in its own service with its own spec; existing webhook + execution controller tests pass (updated only for DI tokens)
- ✅ Legacy step engine: **extracted, not deleted**. Call-graph check confirmed the non-node branch is still reachable in production — `WorkflowSchedulerService` (scheduled step-only workflows), migrated `workflow_execution` cron rows in `CronJobsService`, and the manual/webhook compat fallbacks all route step-only workflows through it
- ✅ `findByWebhookId` no longer loads every workflow and scans in JS — it filters `config->>'webhookId'` in SQL, backed by the new partial expression index `workflows_config_webhook_id_idx` (migration uses `CREATE INDEX CONCURRENTLY`, applied by deploy; **not run locally**)

## Behavior notes

- One deliberate consistency fix: `generateWebhook` now builds the trigger URL from `ConfigService.apiUrl` (`GENFEEDAI_API_URL`) — the same source `getWebhookInfo` already used — instead of a divergent raw `process.env.API_URL` read.
- Seeding callers (`organization-settings`, `self-hosted-seed`, `workflow-deployment-backfill`, `content-schedules`, 7 standalone seed scripts) now resolve `WorkflowTemplateSeederService`; the lazy dynamic-import pattern that breaks the module cycle is preserved.
- `cron-jobs` and `workflow-scheduler` invoke `LegacyWorkflowStepRunner.executeWorkflow` directly (no compat wrapper left behind).

## Verification

- `tsc --noEmit` (api `tsconfig.typecheck.json`): no errors in any touched file (pre-existing errors in unrelated files untouched)
- Biome clean on all changed paths; `bun run lint` (api) passes; secretlint pre-commit passed
- Scoped tests, all green:
  - `workflows.service.spec.ts` (4), `workflow-template-seeder.service.spec.ts` (3, moved from `workflows.service.seed.spec.ts`), `workflow-webhook.service.spec.ts` (10, new)
  - `webhooks.controller.spec.ts` (10), `workflow-execution.controller.spec.ts` (3), `workflow-crud/builder/marketplace/batch.controller.spec.ts` (12), `workflows.module.spec.ts` (1)
  - `cron-jobs.service.spec.ts` (13), `organization-settings.service.spec.ts` (3), `content-schedules.module.spec.ts` (1), `internal-workflow-executions.controller.spec.ts` (3), `agent-workflows.controller.spec.ts` (2), `analytics.controller.spec.ts` (18), `agent-tool-executor.service.spec.ts` (83)
- Full workspace gates left to CI per resource policy

Closes #754


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

Automated structural + behavior-preservation review against the thermo-nuclear code-quality standard. Method: parallel dimension-scoped finder agents (size/structure, branching, duplication, types, layering/DI, behavior preservation) with three-lens adversarial verification per finding; behavior preservation was additionally checked by chunked old-vs-new slice diffs of the 2,653-line monolith against all four extraction targets. Only findings that survived independent re-derivation are listed.

**Credit where due — a lot of this split is verified clean.** All new files are well under the 1,000-line threshold (max 694). A full method inventory of the monolith (61 methods) maps every one to a home in the five services — zero orphans. The step-runner extraction is byte-for-byte for all per-step executors, and the `executeWorkflow` decomposition into `selectReadySteps`/`markStepsBlockedByFailure`/`runStep`/`finalizeRun` is De-Morgan-equivalent with identical throw locations, patch()/cache-invalidation semantics, and terminal-state events. The webhook subsystem preserves every ownership guard, the timing-safe secret comparison, and ID/secret generation; the new SQL `config->>'webhookId'` lookup is predicate-equivalent to the old JS scan, and the `CREATE INDEX CONCURRENTLY` migration matches the query expression exactly and follows the proven Prisma 7.8.0 no-transaction precedent (`20260618130000`, `20260630142600`). Run-control decomposition is field-by-field equivalent. All callers (cron-jobs, scheduler, content-schedules, org-settings, both seed services, all 7 seed scripts, all 3 controllers) were audited 1:1 — correct new service, same args, lazy dynamic-import cycle-breakers preserved; cron-jobs and the scheduler called the plain legacy `executeWorkflow` on master too, so no dispatch regression there. Declared changes (a) and (b) are implemented as described.

Two things block, however.

### BLOCKER

**1. `ensureContentProductionWorkflows` silently drops the content-schedule mirroring loop — undeclared behavior regression** — `workflow-template-seeder.service.ts:375`

On master, `ensureContentProductionWorkflows` did two things: seed the 8 content-production templates, then query all the org's existing non-deleted `contentSchedule` rows and call `ensureContentScheduleWorkflow` for each (old `workflows.service.ts:709-733`). The new method body contains only the template seeding — the `contentSchedule.findMany` + per-schedule mirroring loop is gone, and a branch-wide search confirms no other call site picked it up (`ensureContentScheduleWorkflow` is now only invoked for single schedules from `content-schedules.service.ts:224`). Three production callers depend on the bulk semantics: `organization-settings.service.ts:123` (org onboarding), `self-hosted-seed.service.ts:127`, and `workflow-deployment-backfill.service.ts:310` — the last being a backfill job whose entire purpose is retroactively mirroring existing data. Pre-existing content schedules will no longer get workflow rows during onboarding/seed/backfill. The method's own new docstring (lines 371-374) still claims "Existing content schedules are mirrored into workflow rows," and the PR description lists "content-schedule mirroring" as a preserved responsibility — both now false.

*Restructuring:* re-add the `contentSchedule.findMany({ where: { isDeleted: false, organizationId }, ... })` + per-schedule `ensureContentScheduleWorkflow` loop at the end of the method (or as an explicit `mirrorExistingContentSchedules` step invoked alongside it), and add a spec asserting the mirroring happens.

**2. The SERIALIZABLE-transaction + P2034 idempotent-seed idiom is hand-written three times in the same new file that exists to canonicalize it** — `workflow-template-seeder.service.ts:202,569`

`ensureSeededWorkflow` (line 72) is a clean, table-driven primitive — fast-path `findFirst`, SERIALIZABLE `$transaction` re-check + create, P2034 caught as benign — and 9 of the seeders correctly delegate to it with no per-template flags. But `ensureDailyTrendsDigestWorkflow` (line 202) and `upsertContentScheduleWorkflow` (line 569) each re-implement the identical skeleton by hand because each needs one extra post-conflict branch (schedule repair; update-instead-of-skip). Three hand-maintained copies of the same transactional race-safety idiom in one brand-new file is exactly the duplication this refactor exists to eliminate, and the copies already differ subtly in post-conflict behavior — the failure mode this pattern invites. Confirmed unanimously by three independent verification passes.

*Restructuring:* extend `ensureSeededWorkflow` with an optional `onExisting` hook (or `mode: 'skip' | 'repair' | 'sync'`) receiving the existing row + tx client; default no-op preserves the 9 current call sites, the digest seeder passes the `isScheduleEnabled` repair, and the content-schedule path passes the `tx.workflow.update`. All three copies collapse into the one primitive.

### WARNING

**3. `shouldUseNodeExecutor` copied verbatim into `WorkflowWebhookService`** — `workflow-webhook.service.ts:191`. On master both call sites shared one private method; the split severed that by copy-paste, and `WorkflowSchedulerService` carries a third independently-written variant of the same `nodes.length` check. `WorkflowWebhookService` already injects `WorkflowsService`, so exposing the predicate (or extracting `isNodeBasedWorkflow(workflow)` into a shared helper) removes both extra copies with no new wiring.

**4. The "Systemic workflow templates cannot be executed directly" guard is duplicated across two of the new services** — `workflow-webhook.service.ts:145` and `legacy-workflow-step-runner.service.ts:82`, identical condition and literal string. The duplication technically existed inside the monolith too, but both homes are files this PR created — the natural moment to extract one `assertWorkflowExecutable(workflow)` guard next to the existing `assertWorkflowMutable`.

**5. No typed shape for the webhook config sub-object** — `workflow-webhook.service.ts`. The five webhook fields (`webhookId`, `webhookSecret`, `webhookAuthType`, `webhookLastTriggeredAt`, `webhookTriggerCount`) round-trip through `unknown → Record<string, unknown> → as never` at every read/write site; a field-name typo compiles clean and fails silently at runtime. Define a `WorkflowWebhookConfig` interface (repo convention: `packages/interfaces/`), type `patchWorkflowConfig`'s `updates` as `Partial<WorkflowWebhookConfig>`, and keep the existing runtime guard as the JSON-column boundary check — zero runtime change.

**6. The new class-level JSDoc on `WorkflowTemplateSeederService` overclaims uniform race-safety** — `workflow-template-seeder.service.ts:29`. It asserts every seeder follows the fast-path-read → SERIALIZABLE-recheck shape, but `ensureContentScheduleWorkflow`'s update branch does a plain unguarded `prisma.workflow.update` after a `findFirst` — a carried-over gap, but the doc asserting uniformity is new in this PR. Either bring the update branch into the pattern (which finding 2's `onExisting` hook does for free) or scope the doc to name the exception.

### Simplifications (non-blocking)

- The dedup moment was the natural point to replace the pervasive `(error as { code?: string }).code === 'P2034'` inline casts with a proper `Prisma.PrismaClientKnownRequestError` type guard, and the `as never` create-payload casts with a typed input — the patterns survive at fewer but still multiple sites in the new seeder.
- `executeQueuedJob`'s `keyof TaskQueueClientService` + fixed-signature cast, and `executePublish`'s field-by-field-cast payload ending in `as never` (both verbatim moves in `legacy-workflow-step-runner.service.ts`) — a `QueuedJobMethod` union of the five method names, and a typed create DTO, would let the compiler catch drift these casts currently mask.
- `resumeFromFailed` narrows `executePartial`'s untagged union via `'runId' in x` and re-wraps already-string fields in `String(...)` (verbatim move, `workflow-run-control.service.ts`) — a discriminated union would make the narrowing honest.
- `executeWorkflowCompat` gained an undeclared (harmless) throw path when the optional `legacyWorkflowStepRunner` is absent — a no-op under real module wiring, but worth a line in the PR description since it mirrors deliberate hardening.

### Verdict: **BLOCK**

Explicit blockers: finding 1 (an undeclared behavior regression in a PR whose core claim is behavior preservation — onboarding/seed/backfill lose content-schedule mirroring) and finding 2 (triplicated transactional scaffolding inside the file that exists to canonicalize it). Finding 1 is the serious one; finding 2's fix also resolves finding 6. The rest of the split is genuinely strong — verified behavior-preserving across the step-runner, webhook, run-control, migration, and caller surfaces — so with the mirroring loop restored and the seeder primitive parameterized, this approves.

PR #1099, opened 2026-07-02
fix: 7-day master audit — security, correctness & hardening

VincentShipsIt:
## Summary

Fixes for the security, correctness, and hardening issues surfaced by the 7-day `master` audit (commits `ffb0252ed…ba8b446d1`). Each finding was **re-verified against the code before changing it** — three "findings" turned out to be false positives (noted below) and were left alone.

Every change ships with focused tests; all touched specs are green locally.

### Critical
- **Better Auth privilege escalation** ([`7920e5a7a`](.)) — `beforeAddMember`/`beforeUpdateMemberRole` mapped a BA org `owner`/`admin` role onto the Genfeed `Member` row with no Genfeed-side authz, while BA's native `/organization/*` endpoints are mounted. A BA-org-admin could self-promote to Genfeed `admin`, bypassing RolesGuard. Genfeed owns all membership via its own Prisma services (verified: org creation + creator-member seeding never route through these hooks), so both hooks now `throw` like `beforeCreateInvitation`. Dead role-mapping helpers removed.

### High
- **API-key scope wildcard** ([`40bf3ddea`](.)) — scopes were unvalidated and `hasScope` honored `*` as full access; `POST /api-keys {"scopes":["*"]}` minted an everything-key. Now: DTO restricts to the self-service preset union; the service rejects `*`/unknown at create + rotate (allowing privileged managed scopes); `hasScope` drops the wildcard.
- **Trends corpus cross-tenant read** ([`e7207f458`](.)) — `getCorpusFreshnessHealth` aggregated `Trend` rows across **all** orgs behind only the global auth guard. Now branches: platform admins get the global view, tenants are scoped to their own + global rows. Renamed `corpus/freshness` → `corpus/health` (hard cut). Added `hasMore` so `totalReferences` no longer implies a corpus total. Added the missing `Trend` compound index (CONCURRENTLY migration).
- **ComfyUI path traversal** ([`0262d1fa7`](.)) — untrusted `/history` filename/subfolder built local FS paths + S3 keys unsanitized. Now sanitized (basename, reject `..`/`:`), with a resolved-path containment check that fails closed.
- **Brand-crawler SSRF** ([`ebfacb36b`](.)) — only the initial URL was SSRF-checked; redirects were auto-followed. Now follows redirects manually and re-validates every hop against the private/loopback/metadata blocklist. (DNS-rebind IP pinning noted as a follow-up.)
- **Reviewer notes → generation memory** ([`ea724864f`](.)) — reviewer free-text was persisted verbatim into brand context memory (injected into future prompts). Now stripped of control/zero-width chars, single-lined, fence-broken chars removed, length-bounded, and fenced as untrusted at interpolation.
- **Replicate credit under-billing** ([`0262d1fa7`](.)) — `isReplicateDestination` regex excluded `.`, so dot-versioned keys (`bytedance/seedream-4.5`) classified as `provider=null` and were under-charged. Regex widened (owner segment kept dot-free to avoid over-matching).

### Medium
- **Generation S3-failure status** ([`0262d1fa7`](.)) — S3 upload failure marked the job `completed` with a bogus filename; now throws → job `failed`.
- **Brand slug drift** ([`d93ebd2c1`](.)) — `updateBrandName` synced the org slug but never `brand.slug`; now syncs both (matching sibling paths).
- **Prisma drift guard in CI** ([`dd8a2c7a7`](.)) — `check:model-metadata` now runs in the `guards` job so a drifted enum-field-map fails the build.
- **Non-blocking backfill alerting** ([`dd8a2c7a7`](.)) — the continue-on-error credential backfill now emits a loud annotation + job summary on failure (deploy still not gated).
- **Article link tweet dropped** ([`07ebb076d`](.)) — the trailing link tweet was silently dropped when the URL pushed it over 280; now falls back to the bare URL.

### Low
- Org-scope the direct-column brand-interview field write ([`d93ebd2c1`](.)); remove `console.info` in the backfill service ([`07ebb076d`](.)); add the missing activity-cleanup-failure test for posts ([`07e944917`](.)).

## Verified false positives (left unchanged)
- **MCP brand-interview tools "ungated"** — already listed in `APPROVAL_REQUIRED_TOOLS` in the MCP tool-registry.
- **BrandInterview.start() non-transactional** — already compensates (soft-abandons the session) if credit deduction fails.
- **Article cache invalidation "no-op"** — harmless (busts non-existent keys); removing the SCAN risks stale entries, left as documented minor debt.

## Follow-ups (not in this PR)
- SSRF: connection-level IP pinning to defeat DNS rebinding of a public hostname.
- Feedback-memory: a per-org/brand opt-out toggle (needs a new setting: schema + UI).
- Audit coverage for 4 areas the review workflow never reached (deploy-infra-ci incl. Redis AUTH deferral, service-refactors, workflows-seo, mega-merge `67467c8da`) — pending a follow-up pass.

## Test plan
Focused vitest per touched package (all green): generation, model-key, api-keys, better-auth factory, trends (service + controller), brand-scraper, brand-interview, onboarding, agent-memories, article-thread, post-generation. Prisma drift guard verified locally (in sync). Both edited workflows validated as parseable YAML.


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

**7-day master security/correctness audit — every finding re-verified before landing, 3 audit false-positives deliberately left alone.** This is a net-hardening pass: a genuine privilege-escalation path is closed (Better-Auth org owner/admin hooks were silently mapping onto the Genfeed `Member` row with no Genfeed-side authz check — now throws), a self-service API-key scope allowlist replaces trust-the-caller wildcard scoping, tenant data isolation is fixed in the trends corpus-freshness read path, and path-traversal/SSRF surfaces (ComfyUI history lookup, brand-crawler redirect following) are closed. The diff is 1,669 lines but well-contained; the three files that sit above 1,000 lines (`trend-reference-corpus.service.ts`, `onboarding.service.ts`, `brand-scraper.service.ts`) were all already oversized pre-PR, so their incremental growth is a pre-existing god-object concern rather than a fresh defect introduced here. Spec coverage grew alongside the behavior changes (new `api-keys.service.spec.ts`, expanded scraper/corpus specs), and `better-auth.factory.ts` is a net deletion (dead role-map helpers removed).

### ♻️ Simplifications

- **Loose type on the new security allowlist** — `packages/constants/src/api-key-presets.constant.ts:70-72`. `SELF_SERVICE_API_KEY_SCOPES` is typed `readonly string[]` even though it is derived from the same `API_KEY_SCOPE_PRESETS` source as the `ApiKeyScopePresetValue` literal union declared two lines above it. Retype it as `readonly ApiKeyScopePresetValue[]`:
  ```ts
  export const SELF_SERVICE_API_KEY_SCOPES: readonly ApiKeyScopePresetValue[] = Array.from(
    new Set(Object.values(API_KEY_SCOPE_PRESETS).flat()),
  );
  ```
  so `CreateApiKeyDto.scopes` and any future consumer get compile-time enforcement of the exact allowed-scope set this PR introduces, instead of relying solely on the runtime `@IsIn` check. Widening a value back into the narrow union it was built from is a pure type-level tightening with no runtime change.

### Verdict
**APPROVE** — the security and correctness fixes are the point of the PR and are sound; the single simplification above is a nice-to-have type tightening, not a gate.

<sub>🤖 Automated static read-only thermo-nuclear review of `07e944917` · parallel finders × 3-lens adversarial verification · no tests/builds/typechecks were run.</sub>

PR #1105, opened 2026-07-02
fix(api): resolve 12 TS errors surfaced by ee-billing type-check + close the CI blind spot

VincentShipsIt:
## Problem

On master, `bunx turbo run type-check --filter=@genfeedai/ee-billing --force` fails with 12 TS errors — all in **api** sources, pre-existing, while CI's required `Typecheck` check stayed green.

ee-billing's `tsc --noEmit` aliases `@api/*` into `apps/server/api/src` and is the **only** tsc surface over api code (`nest build` = swc/webpack with `typeCheck: false`; no server app has a `type-check` script).

## Root causes (12 = 3 + 9)

1. **3× TS2345/TS2322** in `system-workflow-provenance.service.ts` — `createExecution`/`completeExecution` took `SystemWorkflowActionInput<unknown>`, but `failureMessage?: (result: T) => string | undefined` makes `T` contravariant under `strictFunctionTypes`, so `SystemWorkflowActionInput<T>` isn't assignable to `<unknown>`. **Fix:** helpers are generic over `<T>`.
2. **9× TS2322** in `workflows.service.ts` — `buildSeededSystemWorkflowMetadata` returned `Record<string, unknown>`, not assignable to Prisma `InputJsonValue` at 9 `workflow.create` metadata sites. **Fix:** typed `Prisma.InputJsonObject` at the builder (no `as never` at call sites).
3. **(fresh checkouts) 2× TS2307 `@genfeedai/integrations/ads`** — package exports map subpaths to `dist/`, which ee-billing's dep graph never builds. **Fix:** `@genfeedai/integrations/*` → `src/*` alias in ee-billing tsconfig, matching `apps/server/tsconfig.json`.

All fixes are type-level/config-only — zero runtime change.

## Why CI never caught it (investigated via rulesets + run logs)

`Typecheck` **is** a required check (ruleset "Passing CI on master") — it just never actually re-ran ee-billing's tsc:

- **Gap A — PR runs use `turbo run type-check --affected`**, which follows the *package.json dependency graph*. ee-billing declares no dependency on `@genfeedai/api` (the coupling is a TS path alias), so PR #1066 — which introduced the errors and touched only `apps/server/api` — ran Typecheck with scope `//, @genfeedai/api, @genfeedai/workers`. ee-billing was never selected, and api has no type-check task.
- **Gap B — master pushes run the full `bun run type-check`**, but api files weren't in ee-billing's task hash, so the Turbo **remote cache replayed a stale green result** (`cache hit, replaying logs` — last real run 2026-06-26, before the regression).

## CI hardening in this PR

- **New `ee/packages/billing/turbo.json`**: package-level `type-check.inputs` with `$TURBO_ROOT$` globs over the alias-reached sources (`apps/server/api/src`, `apps/server/files/src`, `packages/**`, `ee/packages/harness/src`). Closes **Gap B**: api changes now bust the cache.
- **ci.yml Typecheck (PR branch)**: after the `--affected` pass, run `bunx turbo run type-check --filter=@genfeedai/ee-billing` explicitly. Closes **Gap A**. With the inputs above this is a remote-cache hit unless aliased backend sources actually changed — ~free on frontend-only PRs. (Note: `--affected` + `--filter` in one command is an *intersection*, so a single-command union isn't possible.)
- **Tried and reverted**: `futureFlags.affectedUsingTaskInputs` — worked perfectly in local testing (api-only change selected `ee-billing#type-check`; dependent propagation intact), but in CI it made `turbo run build --filter='./packages/*' --affected` execute tasks **outside** the filter (app/website/docs/desktop `next build`s, which that job never runs), and the parallel turbopack builds collided on Next's build lock. Deterministic across a rerun; reverted in favor of the explicit command.

## Validation of the gate — it caught a real bug on its first run

This PR's first Typecheck run failed on the **merge commit**: #1083 (merged to master after this branch's base) annotated `let caption: CaptionDocument | { _id: string } | undefined` in `videos-captions.controller.ts` while `findOne` returns `| null` — a fresh type error #1083's own CI missed via the exact stale-cache gap this PR closes. Fixed here by adding `| null` (the existing `!caption` guard already narrows it). Master's Typecheck is still green on that broken commit — live proof of Gap B.

## Verification

- `bunx turbo run type-check --filter=@genfeedai/ee-billing --force` → **green** on the merged branch (was 14 errors incl. the 2 env-dependent TS2307)
- Cache-sensitivity: api source content change → ee-billing task `cache miss` (inputs hash works)
- Focused vitest (provenance, workflows.service, twitter-pipeline, reply-bot-orchestrator, campaign/dm executors): **6 files, 67 tests pass**
- Biome + secretlint clean (pre-commit hooks)
- Note: master is currently red on `Test API (changed)` from #1083's wave — unrelated to this PR.

## Out of scope / follow-up

Making `apps/server/api` directly tsc-checkable is a real project: with the TS5101 `baseUrl` deprecation silenced, direct `tsc --noEmit` reveals the inherited `baseUrl: "."` breaks `api/tsconfig.json`'s relative `paths` resolution, and `module: commonjs` breaks `import.meta` specs. Same blind spot exists for the other 11 server apps. Follow-up issue to be filed.


VincentShipsIt:
## 🔥 Thermo-Nuclear Review

Automated structural-quality review against the thermo-nuclear standard (behavior-preserving simplification, type contracts, duplication, layering). Method: 4 parallel finder passes (structure/abstraction, duplication, type contracts, layering/atomicity) over the diff and a full checkout of this PR's head, with adversarial verification gating anything reported.

### Findings

None. All four lenses came back empty, and nothing survived to the verification stage.

Notable in the context of this standard — the fixes here move in the right direction rather than papering over symptoms:

- The `SystemWorkflowProvenanceService` fix makes `createExecution`/`completeExecution` generic over `<T>` instead of widening to `unknown` or casting at call sites — the contravariance problem is solved at the signature, which *tightens* the type contract rather than loosening it (the inverse of the loose-type-contracts failure mode).
- `buildSeededSystemWorkflowMetadata` returning `Prisma.InputJsonObject` fixes all 9 call sites at the single producer instead of scattering `as never` casts — one change at the source rather than 9 ad-hoc suppressions.
- The ee-billing tsconfig alias (`@genfeedai/integrations/*` → `src/*`) mirrors the existing `apps/server/tsconfig.json` pattern rather than inventing a new mapping shape.
- The `turbo.json` inputs + `affectedUsingTaskInputs` change closes the CI blind spot at the configuration layer where it originated, which is the root-cause fix the PR description promises.

One non-blocking observation: the per-package `ee/packages/billing/turbo.json` inputs list (api, files, harness, packages) is a hand-maintained mirror of ee-billing's tsconfig path aliases. If those aliases grow, the two must be updated in lockstep — worth a one-line comment in the turbo.json noting it must track the tsconfig `paths`, so the next alias addition doesn't silently reopen the blind spot this PR closes.

---

### Verdict: **APPROVE**

Type-level/config-only change, root-caused at the producer, no structural regressions found.

<sub>Automated review · thermo-nuclear-code-quality standard · parallel finders + adversarial verification</sub>

VincentShipsIt:
## 🔥 Thermo-Nuclear Review

This is a tightly scoped, zero-runtime-behavior PR: it generalizes `createExecution`/`completeExecution` in `system-workflow-provenance.service.ts` over `<T>` to fix a `strictFunctionTypes` contravariance issue, retypes `buildSeededSystemWorkflowMetadata` as `Prisma.InputJsonObject` (eliminating `as never` casts at 9 call sites), adds a missing turbo `type-check` alias/inputs glob for `ee-billing`, and widens one caption variable to accept `null`. All 118 lines are type-level or CI-config hardening — no logic changes.

Adversarial review found no confirmed defects. The diff is well-isolated: the two touched god files (`workflows.service.ts` at 2671 lines, `system-workflow-provenance.service.ts` at 437 lines) are pre-existing and each receives only a handful of lines, not a size regression introduced by this PR. Removing 9 `as never` casts in favor of a correct `Prisma.InputJsonObject` type is a genuine type-safety improvement, and the added `ee/packages/billing/turbo.json` + CI step closes a real stale-cache blind spot rather than papering over it.

No blockers, warnings, or simplification opportunities survived verification.

### Verdict
**APPROVE**

<sub>🤖 Automated static read-only thermo-nuclear review of `44edc9850` · parallel finders × 3-lens adversarial verification · no tests/builds/typechecks were run.</sub>

Issue #1092, opened 2026-07-02
Migrate tenant-shaped platform crons to workflows


VincentShipsIt:
## Problem

The target architecture is "workers manage workflow events; zero predefined product cron — everything tenant-facing is a workflow." Today there are 20 active `@Cron` sites, all allowlisted as platform maintenance in `scripts/architecture/check-platform-cron-boundary.ts` — but several are tenant-product automation that the allowlist has absorbed:

| Cron | Schedule | File | Verdict |
|---|---|---|---|
| `publishScheduledPosts` | `*/15 * * * *` | `apps/server/workers/src/crons/posts/cron.posts.service.ts` | tenant product — workflow candidate |
| `checkPendingTiktokPosts` | every 5 min | `crons/tiktok/cron.tiktok-status.service.ts` | tenant product — workflow candidate |
| `checkScheduledYoutubeVideos` | daily 1am | `crons/youtube/cron.youtube-status.service.ts` | tenant product — workflow candidate |
| `processStreaks` | `30 0 * * *` | `crons/streaks/cron.streaks.service.ts` | tenant product — workflow candidate |
| `refreshExpiringTokens` | hourly | `crons/credentials/cron.credentials.service.ts` | borderline (platform-ish, per-tenant data) |
| `processMonthlyByokBilling` | monthly | `crons/byok-billing/cron.byok-billing.service.ts` | borderline (billing infra, per-tenant effect) |
| `checkScheduledWorkflows` (legacy step executor) | every minute | `crons/workflows/cron.workflows.service.ts` | delete after #1091 + legacy cron-jobs sunset |

Content-schedules conversion is already tracked in #1030 — this issue covers the rest. Genuine platform maintenance (temp-file cleanup, llm-idle shutdown, model watcher/deprecation, trends warmup) stays cron.

## Direction

Per cron: either convert to a workflow-backed trigger (system workflow per #1011) or explicitly re-justify its allowlist entry. Shrink `PLATFORM_CRON_ALLOWLIST` with each migration; the guard already fails on stale entries.

## Context

Repo mapping audit (`docs/audits/00-repo-map.md` §3.6, PR #1083). Related: #1030, #1011, #1034, #1091.

PR #1109, opened 2026-07-02
refactor(workflows): migrate schedules to BullMQ Job Schedulers

### LINKED ISSUE #1091: Migrate workflow schedules to BullMQ Job Schedulers
## Problem

Scheduled workflows fire from **in-process** `cron` CronJobs inside the API (`apps/server/api/src/collections/workflows/services/workflow-scheduler.service.ts`): every API replica builds its own CronJob per workflow, so schedules fire once per replica. PR #1083 added a Redis fire-window lock (`workflow-schedule-fire:{workflowId}:{minuteBucket}`) as the interim guard, but the architecture still puts workflow event firing inside the API.

## Direction

Replace the in-process CronJobs with **BullMQ Job Schedulers** (`queue.upsertJobScheduler`, BullMQ ≥5.16; repo is on 5.79.1) on the `workflow-execution` queue:

- API's role shrinks to `upsertJobScheduler` / `removeJobScheduler` when a workflow's schedule/timezone/enabled state changes (in `updateSchedule` and create/delete paths)
- BullMQ guarantees exactly-one fire per scheduler id across any number of replicas — the interim Redis lock and the hourly `syncScheduledWorkflows` reconciler cron both get deleted
- Workers own all workflow event firing, consistent with the target architecture (workers = workflow event engine, zero product cron)

**Scope discipline (see research comments below):** BullMQ is the timer/delivery mechanism only. Workflow records in Postgres remain the single source of truth for whether work should run; queue state must never be the only record of a schedule. Workers re-read durable workflow state before doing irreversible work.

## Acceptance

- No `CronJob`/`SchedulerRegistry` usage left in `workflow-scheduler.service.ts`
- `workflow-scheduler-reconciler` entry removed from `PLATFORM_CRON_ALLOWLIST` in `scripts/architecture/check-platform-cron-boundary.ts`
- Multi-replica double-fire test (two producers upserting the same scheduler id → one delayed job)
- Stale-fire no-op test: a scheduler job that fires after its workflow was disabled, rescheduled, or deleted re-reads workflow state and exits without executing (cf. gitroomhq/postiz-app#1213, where edits/deletes not propagating to the orchestrator made posts fire at the wrong time)
- Rebuild-from-DB test: after a Redis flush, a reconciliation path can recreate all BullMQ scheduler entries from workflow records with no lost or duplicated schedules (cf. gitroomhq/postiz-app#1550/#1551, where orchestrator startup races silently dropped scheduled posts)
- Timezone handling passes DST-transition cases using IANA zone identifiers, never fixed UTC offsets (cf. gitroomhq/postiz-app#1559/#1145)

## Context

Repo mapping audit (`docs/audits/00-repo-map.md` §6 risk 4, PR #1083).

VincentShipsIt:
Closes #1091

## What

Replaces the in-process `cron` CronJobs in `WorkflowSchedulerService` (one per workflow **per API replica**) with **BullMQ Job Schedulers** on the `workflow-execution` queue (`queue.upsertJobScheduler`, scheduler id `workflow-schedule:{workflowId}`).

- **API producer role only**: `WorkflowExecutionQueueService` gains `upsertWorkflowScheduler` / `removeWorkflowScheduler`; `updateSchedule`, `createWorkflow`, `remove()`, content-schedule ensure/disable, and the three seeding orchestrators (self-hosted seed, deployment backfill, org-settings provisioning via new `syncOrganizationWorkflowSchedulers`) keep schedulers in sync with DB state.
- **Workers fire schedules**: `WorkflowExecutionProcessor` handles the new `scheduled-fire` job type and delegates to `WorkflowSchedulerService.executeScheduledWorkflow` (now public; no locking — BullMQ already guarantees one delayed job per scheduler id per tick).
- **Deleted**: the interim Redis fire-window lock from PR #1083 (`workflow-schedule-fire:{workflowId}:{minuteBucket}`), the hourly `syncScheduledWorkflows` reconciler cron, and the `workflow-scheduler-reconciler` entry in `PLATFORM_CRON_ALLOWLIST`.
- **Migration/self-heal**: a one-time `onModuleInit` sync (idempotent upsert, gated by `GF_DEV_ENABLE_SCHEDULERS` as before) registers schedulers for pre-existing scheduled rows; fires against missing/inactive/systemic workflows remove their own scheduler.
- `scheduled-fire` jobs use `attempts: 1` — retrying a partially executed fire could double-run a workflow; the next tick covers transient failures.

## Acceptance criteria (issue #1091)

- ✅ No `CronJob` / `SchedulerRegistry` usage left in `workflow-scheduler.service.ts`
- ✅ `workflow-scheduler-reconciler` removed from `PLATFORM_CRON_ALLOWLIST`
- ✅ Multi-replica double-fire test: `workflow-scheduler.double-fire.integration.spec.ts` runs two real BullMQ producers against Redis upserting the same scheduler id → asserts exactly **one** job scheduler and **one** delayed job (auto-skips when Redis is unreachable, e.g. unit-test CI runners)

## Verification

- `apps/server/api` workflows + org-settings + content-schedules + seeds suites: 48 files / 259 tests green
- `apps/server/workers` execution processor spec: 11 green (incl. new `scheduled-fire` cases)
- Double-fire integration spec green against local Redis
- `bun run check:cron-boundary`: passed (20 platform crons; reconciler gone) + architecture script tests green
- `bunx turbo run type-check --filter=@genfeedai/api --filter=@genfeedai/workers`: green
- Biome: no new diagnostics in changed files

## Notes

- `apps/server/workers/vitest.config.ts` gains the `@genfeedai/pricing` src alias (mirrors api config) — the processor's new import chain reaches that package and it only ships a dist exports map.
- Not touched: `cron.workflows.service.ts` legacy step-workflow executor and other tenant-shaped crons (tracked in #1092).


VincentShipsIt:
Coordination note from #1111 (merged-order dependency):

**Port the system-workflow skip-guard into this PR's scheduler sync.** #1111 fixes a latent bug where seeded system workflows (single `systemWorkflowAction` node, no engine executor) carried `isScheduleEnabled: true` and were fired by `WorkflowSchedulerService`, logging failed executions every cycle. The fix is two-sided: seeds now write `isScheduleEnabled: false`, and `scheduleWorkflow` skips any workflow with `systemWorkflow` metadata (`getSystemWorkflowMetadata`) — covering pre-fix DB rows without a data migration.

This PR's one-time `onModuleInit` sync upserts a BullMQ Job Scheduler for every `isScheduleEnabled: true` row. Existing DBs still contain pre-fix system-workflow rows with `true`, so without the same metadata skip in the sync/upsert path they get schedulers registered and fire into the no-executor path again.

Also expect textual conflicts on rebase: both PRs touch `workflow-scheduler.service.ts`, its spec, and `scripts/architecture/check-platform-cron-boundary.ts` (#1111 removed 4 allowlist entries + added a `PENDING_TENANT_CRON_MIGRATIONS` entry for `legacy-step-workflow-executor` → #1091).

Suggested merge order: #1111 → this PR (rebase + port skip) → #1079 (trivial `workers/app.module.ts` overlap).

Issue #1090, opened 2026-07-02
Decouple workers from API source imports


VincentShipsIt:
**State refreshed 2026-08-24.** Naming corrected repo-wide: the interim `@genfeedai/server-domain` / `@domain/*` target named below was rejected before it shipped. The actual, already-built target is `apps/server/server` → package `@genfeedai/server`, alias `@server/*` in `apps/server/tsconfig.json`. See `.agents/memory/rules/server_not_core.md` for the permanent naming rule. This rewrite also reconciles child-issue status against what has actually landed.

## Executive Summary

Issue #84 moved BullMQ processors out of the API **process**, but not out of the API **codebase** — the workers app compiles API source directly via the `@api/*` alias. This epic removes that coupling: queue contracts extracted (done), pure infra moved to `packages/libs` (done), a shared server-tier package `@genfeedai/server` scaffolded and partially populated (done), with integration-service and workflow-family extraction still open, gated by a shrink-only CI ratchet that flips to a plain ban once the coupling is gone.

## Problem Statement

Workers deep-imports API source, so api+workers are one modular monolith compiled twice. Any API-internal refactor can break workers with no package-boundary warning, and workers can never version/deploy independently.

**Measured 2026-07-05 (AST scan, baseline):** 208 unique `@api/*` specifiers · 507 import statements · 136 of 168 `.ts` files in `apps/server/workers/src`.

**Current (2026-08-24):** `scripts/architecture/workers-api-imports.baseline.ts` holds **162** remaining specifiers across **114** of **218** files in `apps/server/workers/src` (393 raw `@api/*` occurrences). Down from 208 — real progress has landed, extraction is not finished.

## Decisions (2026-07-05, superseded 2026-07-25 on naming; both folded in here)

1. **Server-tier lib is `apps/server/server` / `@genfeedai/server`** (alias `@server/*`). Two earlier drafts are both cancelled: `apps/server/core` / `@genfeedai/core` (collided with `packages/core`; "core" is retired repo-wide) and the interim `apps/server/domain` / `@genfeedai/server-domain` (`@domain/*`) naming. Vincent flagged the stale `@domain` naming still on this issue on 2026-07-25; this rewrite is the correction.
[...]

Issue #84, opened 2026-04-06
refactor: move 32 BullMQ processors from API to Workers service


VincentShipsIt:
## Summary
Move all BullMQ `@Processor` registrations out of `apps/server/api` and into `apps/server/workers` so HTTP serving and background job processing can scale independently.

## Problem
The API process currently runs 32 BullMQ job consumers alongside request handling. The Workers process runs scheduled cron jobs but no processors. Heavy jobs such as workflow execution, content pipeline work, agent runs, clip analysis, analytics sync, ad jobs, and reply-bot polling compete with API CPU and memory.

### API processors
- workflow-execution, workflow-delay, batch-workflow
- agent-run, content-pipeline, batch-content, workspace-task
- clip-analyze, clip-factory, article-generation
- campaign-processing, orchestrator-run, trigger-evaluation, campaign-memory
- content-optimization, pattern-extraction
- analytics-sync, analytics-twitter, analytics-youtube, analytics-social
- ad-bulk-upload, ad-optimization, ad-sync-google, ad-sync-meta, ad-sync-tiktok, ad-insights-aggregation
- credit-deduction, email-digest, reply-bot-polling, telegram-distribute
- webhook-client

### Workers process
- Runs scheduled crons for content scheduling, analytics sync, ad sync, credential refresh, model deprecation, TikTok status, and related scheduled work.
- Has zero `@Processor` decorators today.

## Scope
- Move processor providers/modules from API runtime registration to Workers runtime registration.
- Keep queue producers available to API where request handlers enqueue jobs.
- Preserve shared queue names, job payload contracts, and serializer boundaries.
- Update module imports so workers own consumers and API owns HTTP plus producers.
- Update deployment/runtime docs so self-hosted and cloud operators know which process consumes jobs.

## Non-Goals
- Renaming all queues.
- Rewriting BullMQ job payload contracts.
- Splitting the monorepo into separate deployable repositories.

## Acceptance Criteria
- [ ] `apps/server/api` has no active BullMQ `@Processor` decorators after migration.
- [ ] `apps/server/workers` registers the migrated processors and can consume the same queues.
- [ ] API endpoints that enqueue jobs continue to work without importing consumer-only modules.
- [ ] Existing cron behavior in Workers still runs after processor migration.
- [ ] Tests or smoke checks prove at least one representative queue is produced by API and consumed by Workers.
- [ ] Deployment docs/config identify API as HTTP-only and Workers as cron plus queue consumer runtime.

[...]

PR #1325, opened 2026-07-05
refactor(server): extract ConfigService/ConfigModule into @genfeedai/libs (#1090)

VincentShipsIt:
## What

Third mechanical slice of the workers ↔ API decoupling in #1090 (successor to #84), after circuit-breaker (#1305) and encryption (#1308). This is the first **infra-foundation** slice under the locked two-tier architecture: **pure infra → `packages/libs`; entangled NestJS domain/integration services → `@genfeedai/server-domain`.** ConfigService is pure infra (a validated env reader), so it moves to `@genfeedai/libs`.

- `git mv` the NestJS config leaf out of the API source tree into `packages/libs/config/` (history preserved as renames):
  - `config.service.ts` + `.spec.ts`
  - `config.module.ts` + `.spec.ts`
  - `services/validation.config.ts` + `.spec.ts`
- Rewire **all consumers** from `@api/config/*` → `@libs/config/*` — **476 `.ts` sites** across `apps/server/api/src`, `apps/server/workers/src`, `apps/server/api/test`, and `ee/packages/billing/src`. Three targeted specifier swaps (`config.service` = 409, `config.module` = 58, `services/validation.config` = 9, plus test/ee/moved-file sites).
- `packages/libs/redis/redis.module.ts` JSDoc usage example repointed to `@libs/config/*` too (removes a stale `@api` reference from a libs doc comment).

## Why this slice / boundary

- `ConfigService` extends `BaseConfigService` from `@genfeedai/config` and depends on **only** `@genfeedai/config` + `@nestjs/common` + `joi` — a clean leaf with no DI graph into API domain services. Both are already resolvable inside `packages/libs` (heavily-used `@nestjs/common` isn't declared there either — consistent with the package's existing hoist convention), so **no new declared cross-package edges** and the lockfile is unchanged.
- This unblocks the **Prisma slice**: `PrismaService` depends on `ConfigService`, so config had to land in `@libs` first.

## Scope boundary

Only the ConfigService/ConfigModule/ValidationConfigService leaf. **Non-TS assets stay in the API** because they belong to the API, not the config service:
- `gpt-actions-openapi.json` (imported by `apps/server/api/src/main.ts`)
- `openapi-internal-routes.json` (API OpenAPI scripts)
- `test.prod.env` (API test env)
- `services/aws.config.ts` (+ spec) — API-only consumer; its `@api/config/config.service` import was repointed to `@libs`, but the file itself stays.

Remaining `@api/*` surface in `workers/src` (domain/integration services) + the final CI guard forbidding `@api/*` from `apps/server/workers/src` are follow-up slices per the issue direction — not bundled here.

## Verification

- ✅ `biome check --write` clean on changed files (import ordering auto-applied — `@api`→`@libs` shifts alphabetical order)
- ✅ Moved specs pass **13/13** under `@genfeedai/libs` vitest (proves `@libs/config/*`, `@genfeedai/config`, and `joi` all resolve + execute inside the package)
- ✅ `@genfeedai/api` **and** `@genfeedai/workers` per-app type-check green via `turbo run type-check` — **20/20 tasks, 0 TS errors** — validates all 476 rewired sites
- ✅ No stale `@api/config/{config.service,config.module,services/validation.config}` references remain in any `.ts`/`.js`/`.json`/webpack/vitest config
- ✅ File moves recorded as renames (history preserved); lockfile unchanged; `@libs` alias confirmed in api/workers/ee tsconfigs, `webpack.base.config.js`, and libs/workers vitest configs

Refs #1090, #84


VincentShipsIt:
## ⚠️ Blocking check root-cause + downstream sequencing (for the next review pass)

**Status:** `mergeable`, but `mergeState=UNSTABLE` — **1 failing check: `OpenAPI Spec Drift`** (22 ✓ / 1 ✗). This is on the critical path for #1090 (see bottom), so it needs to go green before merge.

### What `OpenAPI Spec Drift` actually failed on

It is **not** a spec-diff mismatch. That job **boots the API** (`node apps/server/dist/apps/api/main.js`) to emit the OpenAPI spec, and the **boot crashes**:

```
API bootstrap failed: UnknownDependenciesException
Nest can't resolve dependencies of the ValidationConfigService (?).
argument at index [0] is undefined at runtime
→ commonly caused by 'import type' instead of 'import' for an injectable
```
(run [28746195031](https://github.com/genfeedai/genfeed.ai/actions/runs/28746195031/job/85237422333))

### Root cause — a one-line `import type` introduced by this move

`packages/libs/config/services/validation.config.ts:1`
```ts
import type { ConfigService } from '@libs/config/config.service';   // ❌ type-only import
...
constructor(private configService: ConfigService) {}               // Nest DI needs the runtime ref
```

`import type` erases the class at runtime, so `emitDecoratorMetadata` records `undefined` for the constructor param and Nest can't resolve the token → the `(?)` at index [0]. On `master` this file used a **runtime** import (`import { ConfigService } from '@api/config/config.service'`), which is why master boots; the relocation into `@genfeedai/libs` flipped it to `import type`.

`packages/libs/config/config.module.ts` is otherwise correct (provides + exports both `ConfigService` and `ValidationConfigService`; factory-provides `ConfigService`). Token identity is fine — this is purely the type-only erasure.

### Fix (one line)

`packages/libs/config/services/validation.config.ts:1` — drop `type`:
```ts
import { ConfigService } from '@libs/config/config.service';
```
Then re-run the boot gate. Worth a grep for any other `import type { ... }` on injectables introduced in this diff (same failure mode), since the whole `config/services/*` tree moved.

### Downstream sequencing (per epic #1090 coordination comment, 2026-07-05)

This PR is **step 1** of the libs foundation and **blocks the Prisma/DB infra slice**: `PrismaService` imports `ConfigService`, so the Prisma → `packages/libs` move can only land cleanly once `@libs/config/config.service` is on `master`. Until then the Prisma slice would either re-introduce a `libs → api` edge or fail type-check. Siblings #1308 (encryption) and #1305 (circuit-breaker) are green/CLEAN and independent of this fix.

Refs #1090, #84.

PR #1364, opened 2026-07-06
refactor(api): dissolve onboarding RPC routes into resource PATCHes (REST audit)

VincentShipsIt:
## Summary

Dissolves the resource-shaped **write** routes on `OnboardingController` into their canonical resources, per the endpoint-consolidation report in #1354 (Tier 2, "Onboarding module dissolution"). Applies the north-star pattern from #1302 — side-effects live behind the standard verb, not a bespoke RPC path.

The onboarding module is **not** deleted: the genuine actions stay. Only the 6 resource-shaped writes move out, and the reusable orchestration (scrape, slug-sync, proactive cascade, cache invalidation) stays in `OnboardingService` and is delegated to from the resource controllers.

## Route mapping

| Dissolved RPC | New resource route |
|---|---|
| `PATCH /onboarding/brand-name` | `PATCH /brands/:id { label, syncOrganizationName }` |
| `POST /onboarding/account-type` | `PATCH /organizations/:id { accountType, category }` |
| `POST /onboarding/skip` | `PATCH /organizations/:id/settings { isFirstLogin: false }` |
| `POST /onboarding/complete-funnel` | `PATCH /users/me { isOnboardingCompleted: true }` |
| `PATCH /onboarding/brand/:brandId/confirm` | `PATCH /brands/:brandId` |
| `POST /onboarding/brand/:brandId/reference-images` | `POST /brands/:brandId/reference-images` |
| `POST /onboarding/brand-setup` | `POST /brands/:id/scrape` (renamed action — 9-step orchestration kept) |

**Kept on `OnboardingController`** (audit-verified genuine actions): `status`, `install-readiness`, `prefix/:prefix/available`, `POST prefix`, `proactive-workspace`, `proactive-claim`, `generate-preview`.

> Note: the audit spec listed `skip`'s target as `PATCH /organization-settings/:id`, but no such controller exists — org settings are managed via the existing `PATCH /organizations/:id/settings`, which the appendix itself offered as the same-effect target. `POST /onboarding/prefix` is intentionally left in place (handled by the separate mechanical-collapse chip).

## Cascades kept behind the service
- **Scrape** (`POST /brands/:id/scrape`): the full validate → scrape (site/LinkedIn/X) → AI voice → brand guidance → website-link → org+brand slug sync → journey-reward → first-login-complete pipeline stays in `OnboardingService.setupBrand`, now scoped to an explicit `:id`. `ensureOnboardingWorkspace` is retained as the provisioning safety-net.
- **complete-funnel** → delegates to the idempotent `OnboardingService.completeFunnel` (proactive `markPaymentMade` + 3 access-cache invalidations + timestamp/steps derivation).

## Security / correctness guards (from adversarial review)
- **`syncOrganizationName` org-rename is gated server-side** to the first-login window (`OrganizationSettings.isFirstLogin === true`). A generic brand PATCH can never rename an *established* organization even if a client sets the flag.
- **`markPaymentMade` is best-effort** — a proactive lead-status mismatch no longer blocks onboarding completion (the user has already paid via Stripe when this fires).
- New brand sub-routes verify brand ownership (`verifyBrandAccess`); `isFirstLogin`/`syncOrganizationName` are decorated on their DTOs so the whitelist ValidationPipe doesn't strip them; `syncOrganizationName` is stripped before any CRUD persist.

## Client SDK + callers
- Removed client `OnboardingFunnelService`; removed `setupBrand`/`updateBrandName`/`confirmBrandData`/`addReferenceImages`/`skip` from client `OnboardingService`.
- Added `BrandsService.scrape` / `renameWithOrganizationSync` / `addReferenceImages`, and `OrganizationsService.updateAccountType`.
- Rewired all callers to resolve resource ids client-side: `brand-content.tsx`, `success-content.tsx`, `AgentWorkspaceLayoutClient.tsx`, and **`packages/ui/.../ModalOnboarding.tsx`** — a second live surface the audit's caller grep missed (it only scanned `apps/`, so `confirm` was mis-flagged as 0-callers).

## Verification
- `biome check` — no errors on changed files.
- Type-check (`tsc`) clean on my changed code across `@genfeedai/api`, `services`, `interfaces`, `ui`, `serializers`. Remaining local errors are pre-existing worktree dep gaps (`@types/multer`, `ts-jsonapi`, stale `helpers/dist`) that CI does not have.
- Unit/behavior suites can't run in the worktree (missing `@genfeedai/prisma/testing`, `@testing-library/jest-dom`) — **relying on PR CI** for the full type-check + test run. Updated the affected specs (`onboarding.controller.spec`, `brand-content` / `success-content` behavior tests, `AgentWorkspaceLayoutClient.spec`).

## Reviewer focus
- **Module DI**: adds mutual `forwardRef` `BrandsModule ↔ OnboardingModule` and `UsersModule ↔ OnboardingModule` so the resource controllers can delegate to `OnboardingService`. Consistent with the module's existing pervasive `forwardRef` usage; boot-graph is the main thing to confirm green in CI.

Refs #1354


VincentShipsIt:
## Fixed the two real failures at the source (9fbfd8ba8)

These weren't spec-regen noise — they were an architectural regression in how the dissolution was wired.

### Root cause: the dissolution *added* onboarding coupling instead of removing it

The RPC routes were moved onto `BrandsController` / `UsersController`, but the logic stayed behind `OnboardingService` and was reached via `@Inject(forwardRef(() => OnboardingService))`. That added `BrandsModule → OnboardingModule` and `UsersModule → OnboardingModule` back-imports on top of the pre-existing `OnboardingModule → Brands/Users` edges, closing three new cycles and pushing the module-graph gate **37 → 47 (+10)**:

- `OnboardingModule <-> UsersModule`
- `BrandsModule <-> OnboardingModule`
- `BatchGenerationModule <-> OnboardingModule` (transitively, via `Onboarding → Batch → Brands → Onboarding`)

`forwardRef` is the wrong fix here — `module-graph.spec.ts` counts forwardRef edges too, so the cycles would still count.

### The fix: make OnboardingModule a sink

The true dissolution *removes* onboarding's dependency rather than adding a back-import. Only `AppModule` imports `OnboardingModule` now, so it can't be part of any cycle. **Gate is back to 37** (the +10 fully reverted).

- **`BrandSetupService`** (new, in `BrandsModule`) owns `setupBrand` / `updateBrandNameById` / `addReferenceImages`. Org scope is read from the resolved brand row (the controller already runs `verifyBrandAccess`), so it doesn't re-resolve user context and doesn't pull in Users/UserSetup. `BrandsController` injects it directly — no `OnboardingService`, no `forwardRef`.
- **`BrandPersistenceService` + `BrandDataMapper`** moved from the onboarding endpoint into `BrandsModule` (brand-domain). `BrandDataMapper` is exported for the onboarding preview pipeline, which already imports `BrandsModule`. Dead context-resolution helpers were deleted, dropping the `MembersService` dep.
- **`completeFunnel`** is inlined into `UsersController.updateMe`; the proactive-lead cascade is decoupled through an `ONBOARDING_FUNNEL_COMPLETED_EVENT` handled by `OnboardingCompletedListener` inside `OnboardingModule` — so `UsersModule` stops importing `OnboardingModule` without losing the cascade.

### Failure 2 (brands.controller.spec)
The spec lacked a provider for the controller's new dependency. Updated the DI setup to provide `BrandSetupService` (mock) instead of the old `OnboardingService`.

### Verification
- `module-graph.spec.ts`: **47 → 37**, gate green
- `brands.controller.spec` + all changed onboarding/brands/users specs: **78 tests pass**
- `bunx turbo run build --filter=@genfeedai/api`: clean; NestJS app **bootstraps with no DI/cycle errors**
- `tsc -p tsconfig.typecheck.json`: **0 errors**; biome clean
- OpenAPI spec regenerated (adds `POST /brands/{id}/scrape`, `POST /brands/{id}/reference-images`; onboarding RPC routes gone) — drift resolved

Same class of bug as the mechanical-collapse fix on `refactor/rest-collapse-mechanical`: a collapse that introduced a service/module import cycle.

Refs #1354

PR #1365, opened 2026-07-06
refactor(api): collapse media */latest endpoints into findAll (REST audit)

VincentShipsIt:
## What

Removes the four redundant `GET /{gifs,images,videos,musics}/latest` shortcut routes and migrates their only consumers to `GET /{collection}` (findAll).

## Why

`findAll` already returns newest-first — `handleQuerySort` defaults to `{ createdAt: -1 }` — with `limit` + pagination. The `/latest` variants were pure duplication whose only extra was a bolted-on 5-min cache. This continues the REST audit (#1354) collapse pattern.

## Changes

**API — routes removed**
- `gifs`, `images`, `videos`, `musics`: deleted `findLatest` handlers + their controller-spec blocks.
- `musics` inherits `GET /musics` from `BaseCRUDController`; its `buildFindAllQuery` override already defaults to newest-first, so nothing new was added — just pruned the now-dead imports.
- Dropped unused `@Cache` imports left behind on `gifs` and `images` controllers.

**Consumers migrated**
- **IDE extension** (`apps/extensions/ide/app/src/services/api.service.ts`): `/images/latest?limit=N` → `/images?limit=N`, `/videos/latest?limit=N` → `/videos?limit=N`. Response shape (`ApiResponse.data`) is unchanged.
- **Browser extension** (`apps/extensions/browser/app/src/background.ts`): `getLatestVideos` now calls `/videos?limit=10` and reads the JSON:API `data` array. It previously read a non-existent `videos` key off the collection envelope — **latent bug fixed in passing**.

**Spec**
- `openapi.json`: removed the three `/latest` path blocks (`/gifs/latest` had already been pruned earlier this branch). Surgical edit — the emitter needs a full workspace build this worktree can't run; validated the result is still well-formed JSON and no `*/latest` media paths remain.

## Verification

- No remaining `@Get('latest')` / `findLatest` on the four controllers.
- No consumer anywhere still references a media `*/latest` endpoint (only doc comments).
- Pre-commit biome check + secretlint passed on all 10 files.
- Type-check / unit tests / openapi-drift gate delegated to CI.

### DIFF
--- a/apps/extensions/browser/app/src/background.ts
+++ b/apps/extensions/browser/app/src/background.ts
@@ -505,11 +505,14 @@ async function createVideoFromPrompt(
 }
 
 async function getLatestVideos(sendResponse: SendResponse): Promise<void> {
-  await executeAuthenticatedRequest<{ videos: unknown[] }>(
-    '/videos/latest',
+  await executeAuthenticatedRequest<{ data: unknown[] }>(
+    '/videos?limit=10',
     { method: 'GET' },
     sendResponse,
-    (data) => sendResponse({ success: true, videos: data.videos }),
+    // GET /videos returns a JSON:API collection envelope — newest-first by
+    // default — so the array lives under `data` (the `/videos/latest` shortcut
+    // was collapsed into this endpoint in the REST audit).
+    (data) => sendResponse({ success: true, videos: data.data }),
     'fetch videos',
   );
 }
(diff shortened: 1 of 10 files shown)


VincentShipsIt:
Closing as superseded by #1370 (merged) — same four */latest media route collapses, but #1370 preserves each route's legacy default WHERE filters (training exclusion, isDefault OR-branch, user scoping) via latest=true and keeps the 5-min cache semantics; this PR's bare ?limit=N migration would have dropped them. The browser-extension data-envelope fix here is also covered by #1370's caller repointing.

PR #1568, opened 2026-07-10
fix(cli): paginate through all API keys in `gf keys list`

VincentShipsIt:
## Problem

`listApiKeys` in the CLI fetched a single `/api-keys?limit=100` page and returned only those results ([`api-keys.ts:22`](packages/cli/src/api/api-keys.ts)). Any account with more than 100 keys had the rest **silently dropped** — `gf keys list` under-reported with no truncation signal.

## Fix

Loop the collection's JSON:API `links.pagination` block until exhausted, accumulating every page:

- New reusable `extractPagination()` helper + `JsonApiPagination` type in [`json-api.ts`](packages/cli/src/api/json-api.ts), reading the top-level `links.pagination` (`page`/`limit`/`pages`/`total`) the server serializer emits (`setTopLinks` in `response.util.ts`).
- `listApiKeys` pages through `/api-keys?limit=100&page=N`; `100` is the backend per-page cap (`BaseQueryDto @Max(100)`).
- Stops on the last page, or immediately when the response carries **no** pagination block (safe single-page fallback for non-paginated responses / older APIs / empty result sets).
- Hard `1000`-page ceiling guards against a misbehaving pagination contract ever looping forever.

Chose the **loop-until-exhausted** approach (over a `--limit` flag + truncation warning) so `gf keys list` is correct by default with no new flags.

## Tests

- `tests/api/json-api.test.ts` — `extractPagination` reads the block, returns `undefined` for missing/malformed pagination.
- `tests/api/api-keys.test.ts` — single-page fetch, full multi-page traversal (3 pages), and stop-on-last-page.
- `bunx vitest run tests/api/api-keys.test.ts tests/api/json-api.test.ts` → **14 passed**. Biome + focused `tsc` clean on the changed files.

> Note: `tests/commands/keys.test.ts` fails to load in this worktree due to a missing `@inquirer/prompts` install (unrelated to this change — that suite mocks `api-keys` entirely). CI has the full dependency set.


VincentShipsIt:
Review found the CLI loop could never paginate in production: `setTopLinks` gated on `'total' in data` while `BaseService.findAll` returns `totalDocs`/`totalPages` (the mongo-era `customLabels` rename is swallowed by `PAGINATION_OPTION_KEYS` and never applied), so `links.pagination` was never emitted for any collection. Pushed the root-cause server fix: `setTopLinks` now reads the real `AggregatePaginateResult` shape and emits `pagination: { limit, page, pages: totalPages, total: totalDocs }`, spec updated to the real shape, plus `Number.isFinite` hardening in `extractPagination`.

Issue #334, opened 2026-05-09
Verify Cloud production E2E: signup → billing → credits → publish


VincentShipsIt:
# PRD: Verify Cloud production E2E — signup → billing → credits → publish

> **Rescoped 2026-07-06.** The original "deploy the stack" premise is stale: production is **live on ECS Fargate** (see `.agents/memory/reference_prod_aws_runtime.md`), deploys run from GitHub CI on `master`, and the Mongo-era runtime is gone. The Clerk org/member sync backfill in the original body is obsolete — Clerk is fully removed from the codebase; remaining legacy-ID cleanup is owned by **#739**. What remains here is the **production verification walkthrough** that proves the money path works.

## Executive Summary

Walk the full revenue path on live production as a real user would, capture evidence per step, and file a scoped bug for anything that breaks. This is the acceptance gate for v0.9 — Sell-Ready: nothing in this issue writes code; it produces evidence and bug issues.

## Problem Statement

Production is deployed and serving, but no end-to-end walkthrough of the paid funnel (signup → org → billing → credits → generate → publish) has been executed and documented against the current Better Auth + Postgres + Stripe stack. Individual pieces are verified in isolation (CI, unit/integration tests); the composed path on live infra is not.

## Goals

- Prove each step of the funnel on production with evidence (screenshots / response snippets, no secrets).
- Confirm launch pricing applies end-to-end (`EARLYGENFEED` → $39, #1285).
- Confirm credit grant, decrement on generation, and balance display are consistent.
- Confirm at least one real platform publish from a fresh account.
- File one scoped bug issue per failure found; link them here.

## Non-Goals

- Any code changes (bugs found are filed, not fixed inline here).
- Clerk-ID data cleanup or auth cutover ceremony (#739).
- Load/perf testing.
- Mobile client verification (deferred with the mobile track).

## Verification Checklist

1. **Signup (fresh email)** — email/password signup, email verification delivered (Redis → notifications → Resend) and link works.
2. **Google OAuth signup** — fresh Google identity; and sign-in for an existing email/password account (account-linking gate, #866 regression check).
3. **Org + brand creation** — onboarding completes; org switcher shows exactly the right orgs.
4. **Billing** — upgrade to Creator via Stripe checkout; `EARLYGENFEED` promo auto-applies → $39; webhook lands (no idempotency retry-window regression, #1398); subscription visible in app.
5. **Credits** — plan credit grant appears; generation decrements; balance matches transactions.
[...]

PR #1610, opened 2026-07-11
fix(billing): dedup subscription-invoice credit grants past the Redis TTL

VincentShipsIt:
## Summary

The only prior protection against a duplicate Stripe subscription credit grant was the 24h Redis `stripe:webhook:{event.id}` marker in `StripeWebhookController`. A replay of `invoice.paid` past that window (Stripe auto-retries for up to 3 days, or a manual Dashboard resend) — or a fresh event id for the same invoice — was an unguarded double-grant.

- The **MONTHLY** path already persisted a `referenceId`/`referenceType` on its credit transaction row via `addOrganizationCreditsWithExpiration`, so `StripeWebhookSupportService#hasSubscriptionInvoiceCreditGrant` could dedup it.
- The **YEARLY** path called `resetOrganizationCredits`, which had no mechanism to persist a reference at all — making the pre-check dead code for yearly organizations and leaving them exposed to a full-balance reset (and its side effects) on every replay.

## Changes

- Thread an optional `IAddCreditsOptions` through `ICreditsUtilsService#resetOrganizationCredits` (contract, OSS no-op, and the real `CreditsUtilsService` implementation) so the YEARLY path can persist the same `referenceId`/`referenceType` shape as MONTHLY.
- `StripeInvoiceWebhookHandler` now pre-checks `hasSubscriptionInvoiceCreditGrant` before granting on **both** the MONTHLY and YEARLY branches, and passes a `stripe-invoice:{invoice.id}` reference through to the transaction write on both.
- Add a partial unique index (`credit_transactions_stripe_invoice_reference_key`, `CREATE UNIQUE INDEX CONCURRENTLY`) on `credit_transactions(organizationId, referenceType, referenceId)` scoped to subscription-grant references, as a database-level backstop against a race between the pre-check and the write. `P2002` on that index is now treated as a duplicate no-op rather than an unhandled error.
- New real-backend integration spec, `apps/server/api/test/integration/stripe-webhook-credit-grant.integration.spec.ts` — unlike the existing `payment-processing.integration.spec.ts` (which mocks the entire Stripe SDK **and** the database), this spec:
  - Constructs a genuine Stripe webhook signature via the real Stripe SDK (`generateTestHeaderString` / `constructEventAsync`) and invokes `StripeWebhookController.handleStripe` directly with the raw signed payload — the same code path production traffic takes.
  - Writes to and reads from a real Postgres database via `PrismaService` (through `E2ETestModule.forRoot`), asserting on real `creditTransaction`/`creditBalance` rows.
  - Proves both idempotency layers: replaying the identical Stripe event id (Redis-level no-op) and Stripe redelivering the same invoice under a **new** event id for both MONTHLY (ADD) and YEARLY (RESET) subscriptions (Postgres-level dedup — this is the core #1398 proof, since YEARLY previously had no way to be protected).
  - Wired into `test:e2e:ci`.
- Unit coverage added in `credits.utils.service.spec.ts`, `stripe-invoice-webhook.handler.spec.ts`, and `stripe-webhook-support.service.spec.ts`.

Linked: #334

## Test plan

- [ ] CI: `test:e2e:ci` runs the new `stripe-webhook-credit-grant.integration.spec.ts` against a real Postgres instance
- [ ] CI: unit suites for `CreditsUtilsService`, `StripeInvoiceWebhookHandler`, `StripeWebhookSupportService` green
- [ ] CI: type-check + lint green
- [ ] Verify the new partial unique index migration applies cleanly (`CREATE UNIQUE INDEX CONCURRENTLY`, non-blocking)

### DIFF
--- a/apps/server/api/package.json
+++ b/apps/server/api/package.json
@@ -97,7 +97,7 @@
     "type-check": "tsc --noEmit -p tsconfig.typecheck.json",
     "test:cov": "vitest run --coverage --config vitest.config.ts",
     "test:e2e": "vitest run --config vitest.config.e2e.ts",
-    "test:e2e:ci": "vitest run --config vitest.config.e2e.ts test/e2e/integrations.e2e-spec.ts test/integration/payment-processing.integration.spec.ts test/integration/health.e2e-spec.spec.ts",
+    "test:e2e:ci": "vitest run --config vitest.config.e2e.ts test/e2e/integrations.e2e-spec.ts test/integration/payment-processing.integration.spec.ts test/integration/health.e2e-spec.spec.ts test/integration/stripe-webhook-credit-grant.integration.spec.ts",
     "test:e2e:cov": "vitest run --coverage --config vitest.config.e2e.ts",
     "test:file": "vitest run --config vitest.config.ts",
     "test:file:cov": "vitest run --coverage --config vitest.config.ts",
--- a/apps/server/api/src/collections/credits/services/credits.utils.service.spec.ts
+++ b/apps/server/api/src/collections/credits/services/credits.utils.service.spec.ts
@@ -252,6 +252,56 @@ describe('CreditsUtilsService', () => {
         500,
         txClient,
       );
+      expect(
+        creditTransactionsService.createTransactionEntry,
+      ).toHaveBeenCalledWith(
+        'org_1',
+        expect.anything(),
+        500,
+        100,
+        500,
+        'system',
+        'reset',
+        undefined,
+        txClient,
+      );
+    });
+
+    // #1398: the yearly subscription reset path relies on this reference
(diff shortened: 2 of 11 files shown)


VincentShipsIt:
**E2E validation:** dispatched `e2e.yml` on this branch — [run 29151319657](https://github.com/genfeedai/genfeed.ai/actions/runs/29151319657) — **API E2E Tests: success**. That job runs `test:e2e:ci`, which now includes the new `stripe-webhook-credit-grant.integration.spec.ts`, so the credit-grant + double-replay idempotency assertions passed against **real Postgres**. (The `Frontend Authed E2E` failure is the pre-existing `continue-on-error` job that needs the `E2E_BETTER_AUTH_SESSION_TOKEN` secret — unrelated to this backend change; the E2E Gate passed.)

PR #1617, opened 2026-07-11
test(credits): prove generate -> credit decrement with fake provider + real Postgres

VincentShipsIt:
## Summary

Integration proof for the "generate -> credit decrement" money path (#334), scoped entirely to `apps/server/api` — no cross-workspace `apps/server/workers` processor is driven.

New spec: `apps/server/api/test/integration/generation-credit-decrement.integration.spec.ts`

- **Test 1 — generation completes with a fake provider (zero network).** `ImageGenerationService.generateImage` is exercised through its real public entrypoint. **Narrowing note:** the service is constructed directly (mirroring the existing `image-generation.service.spec.ts` unit-spec pattern) rather than through the full NestJS `TestingModule` DI container — only the four AI provider clients (`ReplicateService`, `FalService`, `KlingAIService`, `ComfyUIService`) are swapped for canned-output mocks; every other dependency is a lightweight inline stub. This is real code exercised through its full public entrypoint (routing, dispatch, prompt building all run for real) — the narrowing is DI-container vs. direct construction, not depth of code coverage. Asserts the response resolves, the Fal mock was called exactly once, and the other three provider mocks were never called (proves correct routing with zero real network calls).

- **Test 2 — credit decrement is real and idempotent.** Uses `E2ETestModule.forRoot` with the real `CreditsUtilsService`/`CreditBalanceService`/`CreditTransactionsService` against real Postgres via `PrismaService`. Seeds an org and a starting balance via the real `addOrganizationCreditsWithExpiration` grant path, then calls the real `deductCreditsFromOrganization` — the exact method the worker's `CreditDeductionProcessor` calls after a generation completes.
  - Asserts `credit_balances.balance` decreases by exactly the deducted amount and exactly one `DEDUCT` `credit_transactions` row is written with the right amount + `referenceId`/`referenceType`.
  - Replays the identical call with the same `referenceId`/`referenceType` and asserts the balance is **unchanged** and still exactly one `DEDUCT` row — confirming the dedup pre-check inside `CreditsUtilsService.deductCore` (keyed on `{ category: DEDUCT, organizationId, referenceId, referenceType, isDeleted: false }`) prevents double-deduction on retry/redelivery.

Also adds `createMockFalService`, `createMockComfyUIService`, `createMockKlingAIImageService` to the shared `external-services.mocks.ts` factory file (the Replicate equivalent already existed), and registers the new spec in `test:e2e:ci`.

## Test plan

- [x] Biome format clean (`bunx biome check --write` on all touched files, zero warnings)
- [x] No local test/typecheck/build run per machine policy — relies on PR CI (`test:e2e:ci`, which now includes this spec)
- [ ] CI green

Linked: #334


VincentShipsIt:
**E2E validation:** dispatched `e2e.yml` on this branch — [run 29155674954](https://github.com/genfeedai/genfeed.ai/actions/runs/29155674954) — **API E2E Tests: success**. That job runs `test:e2e:ci`, which includes the new `generation-credit-decrement.integration.spec.ts`, so the provider-routing + real-Postgres credit-decrement + idempotency assertions passed. (The `Frontend Authed E2E` failure is the pre-existing `continue-on-error` job needing the `E2E_BETTER_AUTH_SESSION_TOKEN` secret — unrelated; E2E Gate passed.)

PR #1621, opened 2026-07-11
test(publish): real-backend proof for credential connect + Ghost publish

VincentShipsIt:
## Summary

Adds `apps/server/api/test/integration/publish-flow.integration.spec.ts` — a real-Postgres integration proof for the credential-connect -> publish path, per #334.

Follows the exact harness pattern from `test/integration/stripe-webhook-credit-grant.integration.spec.ts` (`SKIP_PRISMA_DB` guard, `createTestDatabaseHelper`, real `PrismaService` via `E2ETestModule.forRoot`).

Two tests, both against real Postgres and the real service layer, with **only the outbound Ghost Admin API client mocked**:

1. **Connect** — the real `CredentialsService.create()` (backed by the real `CredentialCryptoService`) persists a Ghost `Credential` row. Asserts `platform=GHOST`, `organizationId` scoped, `isDeleted=false`, the API key is encrypted at rest (stored ciphertext != plaintext), and round-trips correctly through the real `EncryptionUtil` cipher. Read back via real Prisma.
2. **Publish** — the real `GhostPublisherService.publish()` is invoked directly against a seeded draft `Post` row, with only `GhostService.createPost` DI-mocked to return success. Asserts the mock was called once with the expected `(ghostUrl, apiKey, title, html, 'published', undefined)` payload, then applies the exact `Post` state transition `CronPostsService`'s confirmed success path uses (`postsService.patch(..., { externalId, externalShortcode, publicationDate, status: PostStatus.PUBLIC })`) and reads the row back from real Prisma to confirm the transition persisted.

New spec is registered in `test:e2e:ci` (`apps/server/api/package.json`).

## Scope

- **Deliberately does NOT drive** BullMQ, `CronPostsService`, or the workers publish processor — this proves the in-process publish seam only, not the queue/cron orchestration layer around it.
- **Zero production code changes.** Only the new spec file and the `test:e2e:ci` script line were touched.
- No external network calls occur — Ghost's outbound Admin API client (`GhostService`) is the only DI-mocked dependency; everything else (Prisma, `CredentialsService`, `CredentialCryptoService`, `PostsService`, `GhostPublisherService`, `EncryptionUtil`) is real.

## Reviewer notes / assumptions to verify

- CI's Postgres test DB must be reachable and `SKIP_PRISMA_DB` unset for these tests to actually run (rather than skip).
- The default mocked `ConfigService` from `E2ETestModule.forRoot()` only exposes `.get()`/`.getNumber()`, which is incompatible with `CredentialCryptoService`/`GhostPublisherService`'s real class-getter needs (`.tokenEncryptionKey`, `.ingredientsEndpoint`). The spec's own `providers` array overrides `ConfigService` with a real instance (`{ provide: ConfigService, useValue: new ConfigService() }`), relying on last-provider-wins registration order, combined with `process.env.TOKEN_ENCRYPTION_KEY` set at module top-level (a literal test string, not sourced from any `.env*` file) before any module compilation occurs.
- FYI only, not addressed in this PR (out of scope per the "test only, no production changes" directive): a prior investigation of `apps/server/workers/src/crons/posts/cron.posts.service.ts` turned up two latent issues worth a follow-up — an enum-casing mismatch in `publishSinglePost`'s `getPublisher` call, and an undefined `post.credential`/`credentialId` legacy-alias read. Flagging for a future fix, not touched here.

## Test plan

- [ ] CI: `test:e2e:ci` runs `test/integration/publish-flow.integration.spec.ts` against the CI Postgres instance
- [ ] Both tests pass: credential connect + encryption round-trip, and publish + state transition
- [ ] No other `test:e2e:ci` specs regress

Linked: #334


VincentShipsIt:
Converting to **draft**. The publish **path** is already proven and fixed by #1623 (the #1622 P1 fix + green regression test) — this PR is a *supplementary* real-Postgres proof of credential encryption-at-rest + `GhostPublisherService.publish`.

It's blocked on finishing the Prisma-7 strict-create seed/DTO shapes, which surface one-at-a-time per CI round because local test runs are off (MacBook verification rule) — inefficient to iterate here. Remaining known shape fixes (all in `publish-flow.integration.spec.ts`):
- Test 1 `credentialsService.create(createDto)` — `CreateCredentialDto` passes `brand`/`organization`/`user` as strings; the service forwards them to Prisma which expects the scalar FK names (`brandId`/`organizationId`/`userId`) or `connect`. Align the DTO field names to what `credentialsService.create` maps.
- Test 2 `seedCollection('credentials', ...)` — `platform: CredentialPlatform.GHOST` is rejected as not a valid `CredentialPlatform`; reconcile with `seedCollection`'s enum-normalization (it uppercases enums) and the actual Prisma enum values.
- Likely next: the `seedCollection('posts', ...)` fixture uses `platform: 'ghost'` / `status: 'draft'` (lowercase) + relation-alias keys — verify against the Post model + `seedCollection` normalization.

Best finished by iterating against a local Postgres (`SKIP_PRISMA_DB` unset) to catch all shape errors in one pass. Tracked under #334.

Issue #1736, opened 2026-07-14
Decompose oversized service surfaces


VincentShipsIt:
# PRD: Decompose oversized service surfaces

> **State refreshed 2026-08-24 (board audit).** The previous census in this body was measured
> `2026-07-24` and is materially wrong today. Both "dominant remaining risk" god objects it names are
> **compliant and their issues closed**. Meanwhile the largest hand-written file in the repository is
> one the epic never tracked. Corrected below.

## Corrected census — measured on `master`, 2026-08-24

Scope matches the guard's own `RUNTIME_GLOB` (`{apps,ee,packages}/**/*.{cts,mts,ts,tsx}`) with its
exclusion rules applied — fixtures, generated/migration, spec/story/`.d.ts`, and declarative-only
sources removed. `scripts/` and `playwright/` are outside the guard and outside these counts.

| Metric | Stale figure (2026-07-24) | **Today** |
|---|---|---|
| Runtime files ≥ 1,000 lines | 14 | **42** |
| Runtime files ≥ 500 lines | 163 | **373** |
| Controllers > 500 lines | 18 | **11** |
| Agent tool executor | 9,142 lines | **543** |
| Agent orchestrator | 5,332 lines | **801** |

**The 9,142 / 5,332 figures were spec files, not services.** `agent-tool-executor.service.spec.ts`
is 6,229 lines and `agent-orchestrator.service.spec.ts` is 5,011; the services themselves are 543 and
801 and have been compliant long enough that **#519 and #520 are both already closed.** The epic body
has been describing them as the top risk while they were done.

The ≥1,000 and ≥500 counts rose because the earlier figures were narrower than the guard's real scope
and because new surfaces (integrations, extensions, desktop, remix) shipped since. Controllers
genuinely improved, 18 → 11.

## The untracked offender

`apps/server/api/src/collections/content-runs/services/brand-remix-runs.service.ts` — **3,320 lines**,
**3.3× the limit**, twice the next-largest hand-written service, tracked by **nothing** until today.
Now **#3445**.

## Root cause — why the epic drifted

`.github/workflows/ci.yml` runs the complexity gate **regression-only on changed files**;
`.github/workflows/codebase-health.yml` runs the full-repository census **report-only**.

Nothing fails when a *new* file lands already over the limit, and nothing fails when an *existing*
file grows past it incrementally. Both happened:

- `brand-remix-runs.service.ts` arrived at 3,320 lines and never failed a gate.
[...]

PR #1750, opened 2026-07-14
refactor(files): split files controller by route domain

VincentShipsIt:
## Summary

- split the 1,474-line files controller into queue/job, processing, metadata/temp-file, and storage controllers
- preserve all 19 HTTP routes and their existing method/path metadata
- keep existing behavior coverage through the controller suite and add an explicit route-boundary regression test

## Verification

- Biome check passed for the files controller directory
- git diff --check passed
- staged Gitleaks scan passed
- commit hooks passed Biome, control guard, and Secretlint
- local tests, typecheck, and build intentionally deferred to PR CI per workstation policy

Part of #1737
Epic: #1736


VincentShipsIt:
CI result: every scoped and broad job passed, including the @genfeedai/files typecheck, server tests, and full build. The sole failing check is the unrelated ee-billing alias failure already reproduced on master and fixed by #1749 (packages/api-types contract import resolution).

PR #1997, opened 2026-07-23
docs(terraform): warn against bare local apply on genfeed-prod

VincentShipsIt:
## What

Adds a prominent header warning to the `genfeed-prod` Terraform config plus inline flags on the two CI-injected variables. **Comment-only — no resource changes.**

## Why

The prod stack is **CI-first**. Two inputs are supplied at deploy time by GitHub Actions and are not hardcoded; their file defaults are placeholders that only exist so the config parses:

| Variable | File default | CI supplies |
|---|---|---|
| `image_tag` | `"latest"` | real commit SHA (`TF_VAR_image_tag`) |
| `enable_dns_cutover` | `false` | `true` (`TF_VAR_enable_dns_cutover`) |

A bare `terraform/tofu apply` from a laptop inherits those wrong defaults and plans **destructive drift against healthy production**:
- reverts all task definitions to `:latest` → full prod redeploy
- **destroys** the `api`/`mcp`/`notifications` Route53 A-records → prod DNS offline

Nothing is actually missing from AWS — the drift is purely an artifact of applying without the CI-injected vars. This surfaced during a targeted, out-of-band cost apply (Container Insights + log retention, PR #1994), where a full plan showed 13 add / 22 change / 16 destroy. The warning prevents the next person/agent from being startled — or worse, running it.

## Changes

- `providers.tf` — header warning block documenting the two required `TF_VAR_*` overrides and directing changes through CI or a `-target`'d apply.
- `variables.tf` — inline ⚠ flags on `image_tag` and `enable_dns_cutover` cross-referencing the header.

## Verification

- `tofu fmt -check` clean
- `tofu validate` → Success (config valid)

### DIFF
--- a/infra/terraform/genfeed-prod/providers.tf
+++ b/infra/terraform/genfeed-prod/providers.tf
@@ -6,9 +6,9 @@
 # only exist so the config parses:
 #
 #   • image_tag          default "latest"  → CI passes the real commit SHA
-#                                             (TF_VAR_image_tag)
+#                                             (-var="image_tag=<sha>")
 #   • enable_dns_cutover  default false     → CI passes true
-#                                             (TF_VAR_enable_dns_cutover)
+#                                             (-var="enable_dns_cutover=true")
 #
 # A plain local apply runs with those wrong defaults and will plan destructive
 # drift against healthy production: it reverts all task definitions to :latest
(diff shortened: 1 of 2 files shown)


VincentShipsIt:
Fixed in fc489cc. The workflow injects both values via explicit `tofu -var="image_tag=…"` / `-var="enable_dns_cutover=true"` flags, so the header and both variable comments now describe the `-var` mechanism instead of `TF_VAR_*`. The local-apply workaround notes both `-var` and the equivalent `TF_VAR_*` env vars (tofu honors both locally).

PR #2040, opened 2026-07-25
fix(articles): require a signed token for unpublished article reads

VincentShipsIt:
## What

The public article-by-slug endpoint decided whether to return unpublished content from an unauthenticated query flag. This replaces that flag with a signed, slug-bound, expiring preview token.

- **New** `article-preview-token.util` — HMAC-SHA256 over `article-preview:v1:<slug>:<expiry>`, keyed by `TOKEN_ENCRYPTION_KEY`, 7-day TTL, constant-time compare, fails closed on every ambiguous input. Colocated spec covers the fail-closed matrix.
- **Public controller** verifies the token instead of trusting a flag.
- **Redis interceptor** no longer caches or serves preview responses (cache keys are built from public identifiers only).
- **New** org-scoped `GET /articles/:articleId/preview-links` mints the shareable link — the authenticated entry point for the preview capability.
- **Twitter-thread conversion** no longer emits a link for a non-published article.
- **Website** passes `?previewToken=` through and derives the preview banner from `!article.publishedAt` rather than the query string.

## Notes

- No backward-compatible shim: the old query param is gone, per the repo's green-field policy.
- Reuses `TOKEN_ENCRYPTION_KEY`, so no new required env var and no deployment break. Where it is unset, minting returns 400 and verification fails closed.
- The minted URL is a bearer credential — anyone with the link can read that draft until it expires.

## Verification

CI. No local test/typecheck/build run.


VincentShipsIt:
Pushed fixes for the three failing checks. One of them required a structural change that reviewers won't expect from an auth PR, so flagging it up front.

**1. `Test API (changed)` — 19 failures, 2 spec files**

Both failed on the same root cause: this PR added a `ConfigService` constructor dependency, but neither testing module provided it, so `Test.createTestingModule` couldn't resolve the controller.

- `endpoints/public/controllers/articles/public.articles.controller.spec.ts` (9) — now provides a config mock returning no signing key, so preview tokens fail closed and every existing published-only assertion holds. Added a regression test proving a bogus `previewToken` still resolves to a published-only read.
- `collections/articles/controllers/articles.controller.spec.ts` (10) — same provider, plus coverage for the new preview-link route (happy-path signed URL, and `BadRequestException` when no signing key is configured).

The `ConfigService` import in both controllers was already a **value import**, so there was no `emitDecoratorMetadata` erasure bug lurking — `check:di-value-imports` passes clean (2620 files, 0 violations).

**2. `Guards` / `check:runtime-complexity` — the controller split**

`articles.controller.ts` was 890 physical lines against the 500-line ceiling. The bloat is pre-existing; the preview-link route just pulled the file into the changed-file scan. The guard is a regression ratchet against a git baseline with no baseline file and no suppression mechanism, so this had to be a real split rather than an exemption.

Split along cohesive route groups, following how `posts`/`images`/`videos` decompose:

| Controller | Routes |
|---|---|
| `operations/ArticlesOperationsController` | `POST /articles/generations`, `POST /articles/:articleId/reviews` — plus the credit-overdraft preflight both share |
| `transformations/ArticlesTransformationsController` | the eight article-to-artifact routes: thread-conversions, virality-analyses, enhancements, seo-scores, remixes, prompts, images, videos |
| `ArticlesController` | CRUD, versions, and the new preview-links route — **890 → 246 lines** |

Route paths, guards, interceptors, DTOs, and error payloads are unchanged; only the owning class moves. `NOT_FOUND` payloads still report `Article` / `ArticlesController` verbatim via module constants so client-visible errors don't shift.

Both new controllers register **before** `ArticlesController` in the module — their static path segments would otherwise be shadowed by the inherited BaseCRUD `:id` routes. Same ordering constraint already documented in `posts.module.ts`. Specs moved with their routes.

**3. `OpenAPI Spec Drift`**

Delta is the ten operationId renames (operationIds are `<ControllerClass>.<method>`, so the split renames them) plus the new `/articles/{articleId}/preview-links` path.

⚠️ Worth a reviewer's eye: this worktree has no `node_modules` and no generated Prisma client, and `bun install` isn't available here, so `openapi:emit` could not run (both the full build and `--skip-build` fail on baked-in unresolved modules). The spec delta was derived by hand from the emitter's own contract and verified against what CI actually checks: canonical form under `sortKeysDeep` + `JSON.stringify(…, null, 2)` ✅, zero duplicate operationIds ✅, correct alphabetical insertion point ✅, tags unchanged (`AutoSwagger` derives the tag from the controller *path*, not the class name, so `articles` is stable) ✅, `openapi-internal-routes.json` unaffected (no articles entries) ✅, and no other file in the repo references any `ArticlesController.*` operationId ✅. If the drift check still fails, a regenerated spec from a healthy checkout is the fix — the routes themselves are correct.

Locally verified: `check:runtime-complexity --base master` → 0 violations · `check:di-value-imports` → 0 violations · biome clean. Test/type-check/build verification is CI's.

